PR #45, opened 2026-04-22
Fix symmetric Deeprun tram cart rotation

erkagoon:
## Summary
Fix the Deeprun Tram cart pivot correction symmetrically for both wagon sets.

## Testing
- Verified in HermesProxy with the visible Stormwind-side tram (tram 2)
- Verified in HermesProxy with the visible Ironforge-side tram (tram 1)
- Confirmed the one-tram-visible behavior matches the native 1.12 client, so this PR only keeps the cart orientation fix


jameopotato:
Already applied — our fork has the full tram fix at UpdateHandler.cs:2865-2903 via upstream commit 9d2a8ea. Thanks Erkagon!

PR #57, opened 2026-04-26
More ah fixes

Mongrul:
Fixed the rest of the AH. Tested in game. Also sent over the same config outline you use in the launcher, that way I can test new builds via launcer.


jameopotato:
Cherry-picked onto beta as 20a4087 for v5.0.1. HermesProxy.config preserved.

PR #59, opened 2026-04-26
Fix realm swap PTR↔Live cascading session teardown

Mongrul:
Claude Code was able to finish this off correctly - FInal fix 

## Summary
  - Swapping between Kronos Live and PTR via the modern client's realm list required a full relog: the swap appeared to start, but ~150ms into the new realm's WorldClient handshake the proxy tore down the entire BNet session and the modern client hung until the user manually logged out and back in.
  - Root cause: seven separate call sites unconditionally called `GlobalSessionData.OnDisconnect()` on per-realm-socket failures, conflating per-realm-socket lifecycle with BNet account session lifecycle. When the OLD realm's `RealmSocket` entered its send-on-closed path during a swap, it fired the global teardown, which nulled the brand-new `WorldClient` that had just been created for the destination realm.
  - Fix: those seven sites now clean up only their own resource (close the socket, null the appropriate slot) and emit a structured `session.ondisconnect.suppressed` event. The legitimate full-session teardown still runs from `BnetSessionTicketStorage` when the BNet TCP itself disconnects.
  - Investigation methodology: shipped diagnostic events first (`realm.swap.change_ticket`, `realm.swap.join_realm`, `realm.swap.worldclient_orphan`), reproduced the swap, read the bundle, and let the data point to the actual root cause — which was not the SRP-key rotation hypothesis I started with.

  ## Root cause detail
  The pre-fix bundle showed:
  t=275735  realm.swap.join_realm        PTR→Live, fresh keys, world_client_authenticated=true
  t=275770  world.mangos.connect         New WorldClient connecting to Kronos Live :8087
  t=275923  session.disconnect           ← 153ms later. WIPES EVERYTHING.
                had_auth_client=true, had_world_client=true, had_realm_socket=true
  ... 5.5s of nothing — modern client hangs ...
  t=281744  auth.realmd.connect          User gives up, manually relogs
  The `had_world_client=true` on the teardown event proved it killed the
  new realm's connection mid-handshake.

  ## Sites neutralized
  | File:Line | Path | Old | New |
  |---|---|---|---|
  | `WorldSocket.cs:443` | `SendPacket` send-on-closed (the actual swap-killer) | nuke session | null
  per-realm slot only |
  | `WorldSocket.cs:572` | BadVersion auth fail | nuke session | close socket only |
  | `WorldSocket.cs:597` | Unknown OS auth fail | nuke session | close socket only |
  | `WorldSocket.cs:610` | HMAC mismatch | nuke session | close socket only |
  | `WorldSocket.cs:660` | WorldClient connect fail | nuke session | null `WorldClient` slot only |
  | `WorldClient.cs:221` | Legacy-server disconnect | nuke session | null slot **only if still us** |
  | `WorldClient.cs:285` | ReceiveLoop exception | nuke session | null slot **only if still us** |

  `ReferenceEquals(session.WorldClient, this)` guards in `WorldClient`
  prevent an old (dying) WorldClient from clobbering a freshly-created
  one during a swap race.

  ## Diagnostics added (kept in tree)
  Every step on the realm-swap path now emits a structured event:
  - `realm.swap.change_ticket` — auth state + SRP key prefix before/after `Reconnect()`, with `auth_key_changed` flag
  - `realm.swap.join_realm` — old/new realm IDs, `is_swap`/`is_reconnect` flags, key prefixes, snapshot of stale `GameSessionData` carryover
  - `realm.swap.worldclient_orphan` — fires only when a `WorldClient` already exists at the new-realm `HandleAuthSession` site
  - `session.ondisconnect.suppressed` — emitted at each of the 7 neutralized sites with the `reason` field, so future bundles immediately confirm the cascade is still suppressed

  ## Verification

  Two real PTR↔Live swaps in a single end-to-end session, no relog:

  | Signal | Pre-fix bundle | Post-fix bundle |
  |---|---|---|
  | `auth.realmd.connect` count (per session) | 2 (forced relog) | **1** |
  | `session.disconnect` events during swaps | 2 per swap | **0** |
  | `session.ondisconnect.suppressed` events | n/a | 3 (each one would have killed the swap) |
  | Successful Live↔PTR swaps end-to-end | **0** | **2 back-to-back** |
  | `auth_key_changed` per swap | true | true (unchanged — SRP rotation always worked) |

  t=309910  auth.realmd.connect              single login
  t=310548  realm.swap.join_realm             initial Kronos Live, is_swap=false
  t=314368  realm.swap.change_ticket          auth_key 363B75FB → 480EC67F
  t=316442  realm.swap.join_realm             Live → PTR, is_swap=true
  t=316603  session.ondisconnect.suppressed   worldclient_legacy_disconnect
  ... gameplay on PTR ...
  t=332535  realm.swap.change_ticket          auth_key 480EC67F → 22225EC4
  t=334331  realm.swap.join_realm             PTR → Live, is_swap=true
  t=334365  world.mangos.connect              Kronos Live ✓
  ... gameplay continues without relog ...

  ## Diagnostic bug also fixed
  The first-pass `is_swap` check in `RealmManager.JoinRealm` compared
  `oldRealmId.Index != 0`, which false-negatived when swapping FROM Live
  (index 0). Now keyed on `hadPriorRealm` (= whether a `WorldClient`
  already exists), with a sibling `is_reconnect` flag for same-realm
  rejoins.

  ## Known follow-up (not blocking)
  `gamestate_cached_players` shows 5–8 stale entries from the previous
  realm at swap time (no `GameState` reset on swap today). The swap
  works fine despite it, but a follow-up should call
  `GameSessionData.CreateNewGameSessionData(globalSession)` in
  `JoinRealm` when `is_swap=true` to drop carryover. Tracked separately.

  ## Test plan
  - [x] Two real Live↔PTR swaps in one session, no manual relog
  - [x] `session.ondisconnect.suppressed` fires for each old-realm WorldClient close, **`session.disconnect` does not fire** during swaps
  - [x] `auth.realmd.connect` count = 1 across the whole session
  - [x] `is_swap` reads correctly on both swap directions
  - [ ] Confirm an actual full-session teardown (close the modern client, kill the BNet TCP) still emits the legitimate `session.disconnect` from `BnetSessionTicketStorage`

### DIFF
--- a/HermesProxy/Realm/RealmManager.cs
+++ b/HermesProxy/Realm/RealmManager.cs
@@ -264,12 +264,41 @@ public sealed class RealmManager
 
     public BattlenetRpcErrorCode JoinRealm(GlobalSessionData globalSession, uint realmAddress, uint build, IPAddress clientAddress, byte[] clientSecret, string accountName, Bgs.Protocol.GameUtilities.V1.ClientResponse response)
     {
+        // JimsProxy: realm-swap diagnostics. Snapshot the pre-swap state so the
+        // emitted event tells us (a) whether this is an actual swap or initial join,
+        // (b) whether the prior realm's WorldClient is still alive at swap time,
+        // and (c) what stale GameSessionData carries over (no reset happens here).
+        // hadPriorRealm is the right "is this a swap" signal — the old logic compared
+        // oldRealmId.Index against 0, but Index=0 is the Live realm, not "no realm".
+        // Using HadPlayerLogin OR a non-default RealmId would be cleaner; for now we
+        // gate on whether a WorldClient already exists, which only happens after the
+        // first realm has been picked.
+        var oldRealmId = globalSession.RealmId;
+        var oldGameState = globalSession.GameState;
+        bool hadWorldClient = globalSession.WorldClient != null;
+        bool hadPriorRealm = hadWorldClient;
+        bool worldClientWasConnected = globalSession.WorldClient?.IsConnected() ?? false;
+        bool worldClientWasAuthenticated = globalSession.WorldClient?.IsAuthenticated() ?? false;
+        bool authClientWasConnected = globalSession.AuthClient != null && globalSession.AuthClient.IsConnected();
+        string authKeyPrefix = SafeAuthKeyPrefixForRealm(globalSession.AuthClient);
+        string bnetKeyBefore = SafeKeyPrefix(globalSession.SessionKey);
+
         globalSession.RealmId = new RealmId(realmAddress);
         Realm? realm = GetRealm(globalSession.RealmId);
         if (realm != null)
         {
             if (realm.Flags.HasAnyFlag(RealmFlags.Offline) || realm.Build != build)
+            {
+                Framework.Logging.Log.Event("realm.swap.join_realm", new
+                {
+                    phase = "rejected_offline_or_build",
+                    new_realm_index = (uint)globalSession.RealmId.Index,
+                    realm_offline = realm.Flags.HasAnyFlag(RealmFlags.Offline),
+                    realm_build = realm.Build,
+                    client_build = build,
+                });
                 return BattlenetRpcErrorCode.UserServerNotPermittedOnRealm;
+            }
 
             RealmListServerIPAddresses serverAddresses = new RealmListServerIPAddresses();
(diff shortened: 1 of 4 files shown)


jameopotato:
Cherry-picked onto beta as 2d6ab36 for v5.0.1.

PR #49, opened 2026-04-23
Fork fixed, pushing over 10

Mongrul:
Lotso of bug fixes  here, the spell visual fix was the big one.

PR #61, opened 2026-04-26
Flying Mob Fix - FInal

Mongrul:
Claude code was able to finish it perfectly.

The bug

  The previous hover fix (PR #49 / commit 4954ade) gated everything on UNIT_FIELD_HOVERHEIGHT > 0. That works
  on cores that populate the field, but Twinstar's 1.12.1 server leaves it at 0 for whelps, bats, dragonhawks,
   and other hovering mobs. PTR JSONL diagnosis confirmed the silent failure: 366 SMSG_ON_MONSTER_MOVE packets
   in a single Redridge session, zero hover-override events fired.

  Whelps would fly along their splines correctly, then "basketball-bounce" up and down between flight legs —
  the modern client receiving heartbeats and Stop packets with stale Falling flags, alternating between
  gravity-snapping the unit to the navmesh and catching up to its true airborne position on the next update.

  What was leaking

  Three packet paths bypassed the original fix even when HOVERHEIGHT was set:
  1. SplineTypeLegacy.Stop early-return — sent SplineFlags = None, letting the client fall back to cached Falling state
  2. HandleMovementMessages (MSG_MOVE_HEARTBEAT and 18 other movement opcodes) — never applied any hover correction
  3. HandleMoveUpdateSpeed — same

  Plus a real bug in the original strip mask: moveInfo.Flags &= ~0x00000800u was clearing
  MovementFlagWotLK.Root (0x800), not Falling — the magic number matched MovementFlagModern.Falling but the
  flags at that point were still in WotLK space.

  The fix

  Server-agnostic detection. New GameSessionData.KnownHoveringMobs (HashSet<WowGuid128>) seeded from two
  reliable signals:
  - SplineFlagVanilla.Flying on incoming SMSG_ON_MONSTER_MOVE (every flying mob hits this)
  - MovementFlagVanilla.FixedZ on MovementInfo blocks (already captured into MovementInfo.Hover, previously unused)

  Player guid is excluded from detection so taxi flights don't mark the player.

  Single source of truth. New IsHoveringMob(guid) helper OR's the registry against UNIT_FIELD_HOVERHEIGHT > 0,
   used at every override site.

  Shared override helper. ApplyHoverOverrideIfNeeded strips Falling/FallingFar (now using named WotLK enums,
  not the buggy magic number), forces MovementFlagWotLK.DisableGravity, and zeros
  FallTime/JumpVerticalSpeed/JumpHorizontalSpeed. Called from:
  - ReadMovementUpdateBlock (UPDATE_OBJECT Living block) — was already there but with the wrong flag
  - HandleMovementMessages — heartbeats, root/swim toggles, fall_land, hover toggles (new)
  - HandleMoveUpdateSpeed — server-broadcast speed changes carrying full MovementInfo (new)

  Stop branch fix. MonsterMove Stop packets now carry Flying | AnimTierHover spline flags for hovering mobs,
  so the client settles in hover state instead of inheriting cached Falling.

  AnimTier persistence. The UNIT_FIELD_BYTES_1 block now sets AnimTier=2 based on either HOVERHEIGHT > 0 or
  KnownHoveringMobs membership, so registry-detected mobs still get the correct anim tier on subsequent
  updates.

  Diagnostics

  Four new structured events for JSONL bundles:
  - hover.detect_flying_spline — first time we see a guid's flying spline
  - hover.detect_fixedz_flag — first time we see a guid's FixedZ movement flag
  - hover.spline_override — per spline that gets the override applied
  - hover.spline_stop_override — per Stop packet that gets hover flags

  Test plan

  - Builds clean (0 warnings, 0 errors)
  - Verified on PTR (Twinstar 1.12.1) — Redridge whelps no longer bounce between flight legs, no slowfall dip at end of flight paths
  - Verify no regression on Kronos (where HOVERHEIGHT is populated) — both detection paths should fire
  - Verify no regression on player taxi flights (player guid excluded from detection)

  Files changed

  - HermesProxy/GlobalSessionData.cs — KnownHoveringMobs registry
  - HermesProxy/World/Client/PacketHandlers/MovementHandler.cs — helpers, Stop branch, heartbeat/speed-update paths, spline detection seed
  - HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs — flag-mask bug fix, FixedZ detection seed, AnimTier registry check

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -139,6 +139,12 @@ public sealed class GameSessionData
     public HashSet<uint> RequestedItemHotfixes = [];
     public HashSet<uint> RequestedItemSparseHotfixes = [];
 
+    // Mobs we've seen send Flying spline or FixedZ movement flags. Vanilla servers
+    // don't populate UNIT_FIELD_HOVERHEIGHT consistently (Twinstar e.g. leaves it at 0),
+    // so we need a server-agnostic hover signal. Once a guid lands here, all subsequent
+    // packets for it get the hover override regardless of HOVERHEIGHT.
+    public HashSet<WowGuid128> KnownHoveringMobs = [];
+
     private GameSessionData()
     {
         
--- a/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
@@ -10,6 +10,34 @@ namespace HermesProxy.World.Client;
 
 public partial class WorldClient
 {
+    // Returns true if the mob should be treated as hovering. Detection is OR of:
+    //   - UNIT_FIELD_HOVERHEIGHT > 0 (works on cores that populate it)
+    //   - KnownHoveringMobs membership (seeded when we see Flying spline / FixedZ flag)
+    private bool IsHoveringMob(WowGuid128 guid)
+    {
+        if (GetSession().GameState.KnownHoveringMobs.Contains(guid))
+            return true;
+        return GetSession().GameState.GetLegacyFieldValueFloat(guid, UnitField.UNIT_FIELD_HOVERHEIGHT) > 0.0f;
+    }
+
+    // Strips Falling/FallingFar and forces DisableGravity on a hovering mob's movement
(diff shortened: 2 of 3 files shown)


jameopotato:
Cherry-picked onto beta as 0e28ae8 for v5.0.1.

PR #62, opened 2026-04-26
Fix gold chat delay: synthesize loot notify before Kronos forward

Mongrul:
The synthesized SMSG_LOOT_MONEY_NOTIFY in HandleLootMoney was sent
  AFTER forwarding CMSG_LOOT_MONEY to the legacy server. SendPacketToServer
  performs a synchronous, locked, encrypted write to the Kronos socket,
  which gated the chat line behind the full server-write latency.

  Reorder so the synthesized notify ships to the modern client first,
  then forward CMSG_LOOT_MONEY to Kronos. This matches native 1.12 client
  behavior, which printed the chat line locally on CMSG send (zero latency).

  JSONL diagnostics across 6 loot cycles confirmed:
  - synth fires same millisecond as CMSG arrival (0-1ms)
  - ~140ms ahead of Kronos's SMSG_LOOT_CLEAR_MONEY ack
  - Kronos never sends its own SMSG_LOOT_MONEY_NOTIFY (no double-print)

  Reported by testers on beta build after the loot-failure fix shipped.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/LootHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/LootHandler.cs
@@ -1,5 +1,4 @@
 ﻿using Framework;
-using Framework.Logging;
 using HermesProxy.Enums;
 using HermesProxy.World.Enums;
 using HermesProxy.World.Objects;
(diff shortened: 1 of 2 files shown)


jameopotato:
Cherry-picked onto beta as 255d31b for v5.0.1.

PR #63, opened 2026-04-26
 Fix over-head quest progress toast: track item count proxy-side

Mongrul:
The over-head "Item: X/N" banner was stuck at "1/N" on every pickup
  because Kronos never refreshes PLAYER_QUEST_LOG_x_3 on partial item
  pickups. Two consumers were reading from that stale cache:

    - HandleItemPushResult computed QuantityInInventory = item.Quantity
      + cachedField, which always evaluated to (1 + 0) = 1.
    - HandleQuestUpdateAddItem emitted the SIMPLE credit variant, which asks the modern client to re-read from the same dead field.

  Switch to the full SMSG_QUEST_UPDATE_ADD_CREDIT with explicit
  Count/Required, and source QuantityInInventory from the proxy's own
  QuestItemObjectiveProgress running total (kept in sync per pickup).

  Lifecycle fixes so the running total stays accurate across sessions:
    - Seed from the legacy quest log cache on first credit per objective, so logging in mid-quest at 3/10 progresses to 4/10 (was 1/10).
    - Clear running totals on CMSG_QUEST_LOG_REMOVE_QUEST (abandon) so re-accepting the same quest doesn't inherit stale counts (was showing 7/10 on the first pickup of a fresh accept).
    - Clear running totals on SMSG_QUEST_UPDATE_COMPLETE / _FAILED / _FAILED_TIMER so turn-in / failure paths reset the same way.

  Verified across fresh-accept, mid-quest-login, abandon-and-reaccept,
  and turn-in scenarios via JSONL diagnostic capture (since stripped).

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/ItemHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/ItemHandler.cs
@@ -58,27 +58,40 @@ public partial class WorldClient
             item.QuantityInInventory = packet.ReadUInt32();
         else
         {
-            uint currentCount = 0;
+            //MIRASU - Vanilla legacy SMSG_ITEM_PUSH_RESULT doesn't carry running QuantityInInventory.
+            //MIRASU   Original code read the count from the player's PLAYER_QUEST_LOG_x_3 cache, but
+            //MIRASU   Kronos never refreshes that field on item pickups -- so QuantityInInventory was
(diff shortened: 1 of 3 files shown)


jameopotato:
Cherry-picked onto beta as 6995e72 for v5.0.1.

PR #64, opened 2026-04-27
Fix mob cast-bar dismiss + animation drift on interrupt

Mongrul:
When a player Kicked/Counterspelled a casting mob on Kronos 1.12, the
  target-frame cast bar continued progressing despite the spell being
  interrupted server-side. Animation chunks were missing, casting sounds
  clipped, and the bar ran to its programmed CastTime regardless of the
  SMSG_SPELL_FAILED_OTHER the proxy was forwarding.

  Three root causes all present together:

  1. Deterministic CastID reuse: every mob cast of the same spell shared the same CastID (spellId+casterCounter). Modern client treats consecutive SPELL_STARTs with identical CastIDs as one continuous in-flight cast, which bled visual state across cycles -- chunks dropped, sounds clipped, and no dismiss path could correlate to a specific cast.

  2. Missing SMSG_SPELL_INTERRUPT_LOG: Vanilla 1.12 has no equivalent opcode, so the proxy never synthesized one. The 1.14 Classic client uses this as the actual target-frame cast-bar dismiss trigger -- not SpellFailedOther, which only drives combat log + sound suppression. Crucially, the client gates the dismiss on BackfireSpellID != 0; with BackfireSpellID=0 the interrupt is treated as informational and the bar persists. This was the silent gotcha after iterating through every other approach.

  3. Missing SMSG_CANCEL_SPELL_VISUAL: lingering caster animation kept playing even when other dismiss signals were sent. Required the underlying SpellVisualID (e.g. 13 for Frostbolt), not the SpellXSpellVisual wrapper ID emitted in SpellStart.

  Fix synthesizes the four-packet chain the 1.14 client expects:

  - Unique CastIDs per mob cast via per-(caster, spell) sequence counter stored in GameState.OtherCasterActiveCastIds. SPELL_GO and SPELL_FAILED_OTHER look up the same CastID minted at SPELL_START so the entire cast lifecycle references one consistent ID.

  - SpellInterruptLog synthesized on FAILED_OTHER with reason=Interrupted for non-player/non-pet casters. Caster=local player, Victim=mob, BackfireSpellID=interrupted spell ID (loops back since 1.12 doesn't send the school-lockout spell on this opcode).

  - CancelSpellVisual emitted with the proper SpellVisualID via a new GameData.SpellXSpellVisualToSpellVisual mapping, sourced from HermesProxy/CSV/SpellVisualResolved1.csv (14,390 rows derived from the 1.14.2.42214 client DBC dump via wago.tools). Loader is missing-file tolerant so TBC/WotLK can ship CSVs later.

  - Existing SpellFailure + SpellFailedOther emissions retained for combat log + sound suppression.

  Also adds dequeue safety in HandleSpellFailure (player's own cast
  failures) -- the prior peek-only FirstOrDefault path could leak
  PendingNormalCasts entries when a cast failed before its SpellGo
  arrived. Switched to TryDequeuePendingNormalCast/TryDequeuePendingPetCast
  with a synthetic SpellPrepare emission for the !HasStarted case so the
  client correlates the failure to the right cast.

  Diagnostics: spell.cast event extended with casterCounter and
  castIdCounter; new spell.failure.routed and spell.failed_other.routed
  events to correlate cast lifecycle in the JSONL stream.

  File checklist before git add:
  - HermesProxy/GlobalSessionData.cs
  - HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
  - HermesProxy/World/GameData.cs
  - HermesProxy/World/Server/Packets/SpellPackets.cs
  - HermesProxy/CSV/SpellVisualResolved1.csv (new file — must be git add'd explicitly)


jameopotato:
Cherry-picked onto beta as a3769dd for v5.0.1.

PR #66, opened 2026-04-27
Channeled spells  WIP

Mongrul:
What's committed

  58ebe28 — SpanWriter caps + chat-noise demote (passed bug-hunter, smoke-tested clean)
  - MovementPackets.cs:69 — MaxSplinePoints 16 → 48 (kills populated-zone fallback storm)
  - SpellPackets.cs:311 — MaxEntries 64 → 128 (kills per-login SendSpellHistory fallback)
  - ChatHandler.cs:374 — Console.WriteLine → Log.Print(LogType.Debug, ...) (quiets addon-comm spam in
   bundles; in-game chat unaffected)
  - MonsterMoveTests.cs — bumped fallback test from 20 → 64 points to stay above the new cap

  Merge from beta — brings Redemption Headpiece char-select fix and other beta-only fixes (realm
  swap, AH armor slot, gold chat delay, flying mob, over-head quest progress) onto this branch.

  What's uncommitted (in-progress channel cast bar work)

  Goal: render the target-frame channel cast bar when looking at another player/NPC channeling.
  Self-channel bar already worked via existing MSG_CHANNEL_START path.

  Three files touched, additive only:

  GlobalSessionData.cs — added ConcurrentDictionary<WowGuid128, uint> OtherCasterActiveChannels to
  track last-observed UNIT_CHANNEL_SPELL per non-self unit and detect 0↔spellId transitions across
  consecutive SMSG_UPDATE_OBJECT arrivals.

  UpdateHandler.cs:1882 — wraps the existing UNIT_CHANNEL_SPELL UpdateField read. On observed 0 →
  spellId or spellId → spellId' transitions for non-self units, fabricates SMSG_SPELL_CHANNEL_START
  with caster GUID, spell ID, GameData.GetSpellVisual(spellId), and duration from
  GameData.GetAuraSpellDuration(spellId) (5000ms fallback). Emits channel.update_field.observed and
  channel.synthesized.start events with raw Low/High of guid and CurrentPlayerGuid for diagnostic
  confirmation.

  Important: isSelf uses explicit guid.Low == cpg.Low && guid.High == cpg.High instead of the
  auto-generated record-struct ==. A previous bundle showed the operator returning false for what
  looked like identical values; the explicit form is defensive against whatever was going on there.

  SpellHandler.cs:1211 — added channel.msg_start.received diagnostic to the existing self-channel
  handler so bundles can confirm vanilla never sends MSG_CHANNEL_START for non-self casters (expected
   behavior).

  What the latest two-box test bundle showed

  Account A (412HAYWARD, player GUID Low=1728) channeled Arcane Missiles. Account B (413HAYWARD,
  player GUID Low=843) observed, both proxies running with logs co-mingling into
  jimsproxy-20260427-124816.jsonl (same Logs directory — that's the "mixing" — not a bug, just two
  processes).

  What worked in the bundle:
  - B's proxy correctly observed A's UNIT_CHANNEL_SPELL transition with is_self: false (cpg_low=843,
  target_low=1728, transition=start).
  - B's proxy emitted channel.synthesized.start with spell_visual_id: 239917, duration_ms: 5000
  (fallback).
  - A's proxy correctly identified its own channel as is_self: true and skipped synthesis.
  - The explicit Low/High check correctly distinguished self from non-self.

  What still didn't work:
  - Cast bar on B's target frame still does not render despite the synthesized
  SMSG_SPELL_CHANNEL_START reaching B's modern client.

  Where the next coder should look

  This is scenario B from the diagnostic plan: the synthesis is firing correctly but the modern 1.14
  client isn't rendering the bar from SMSG_SPELL_CHANNEL_START for non-self units. Hypotheses, in
  order of likelihood:

  1. Packet ordering. The synthesized SMSG_SPELL_CHANNEL_START is sent during UNIT_CHANNEL_SPELL
  UpdateField processing, which means it likely arrives after the SMSG_UPDATE_OBJECT carrying
  UnitData.ChannelData. Modern client may need the start packet to arrive before the UnitData update.
   Try queuing it to send before the next UpdateObject flush, or via _delayedPacketsToClient.
  2. Spell ID translation. Vanilla spell 5143 is being forwarded raw. Modern Classic Era client
  should recognize it (same DBC), but worth confirming via GameData.LegacyToModernSpellId. If modern
  expects a translated ID, the current code is wrong.
  3. SpellXSpellVisualID validity. GameData.GetSpellVisual(5143) returned 239917. Worth verifying
  that's a valid SpellXSpellVisual.db2 row in the 1.14 client. If it's a stale or fabricated value,
  the client may reject the whole packet.
  4. Missing companion packet. The recent dismiss fix (a3769dd) showed the client gates UI on packets
   vanilla doesn't have an equivalent for (SMSG_SPELL_INTERRUPT_LOG). Channel rendering may similarly
   need SMSG_SPELL_GO or SMSG_SPELL_START first, with the channel as a follow-up. Look at how the
  client renders cast bars when SPELL_GO is received vs SPELL_CHANNEL_START — there may be a
  precondition we're missing.
  5. Duration semantics. 5000ms fallback may be wrong for Arcane Missiles (vanilla rank 1 is 3000ms).
   The MSG_CHANNEL_START path uses the wire duration; the synthesis can't (we never get one for
  non-self). Could try populating AuraDurations CSV with channel durations for known channeled
  spells, or look up duration from a different source.

  Suggested next-step diagnostic

  Add a log dump of the actual bytes sent in the synthesized SMSG_SPELL_CHANNEL_START so it can be
  diffed against what the client receives in a real (self-cast) SMSG_SPELL_CHANNEL_START from the
  same session. If the wire format differs, that's the bug. Otherwise it's an ordering or
  precondition issue.

  Test recipe

  Two-box on Twinstar/Kronos PTR with X as caster and Y as observer. Have observer
  target caster, caster channels Arcane Missiles. Observer's bundle should show
  channel.synthesized.start with is_self: false — and either show the bar (success) or not (continue
  debugging).

  Build clean, all 312 unit tests passing, no warnings.


jameopotato:
Channel cast bars are now handled via a different approach in v5.0.3-beta.1: the proxy sends channel start/stop data to the JimsPlus addon via CHAT_MSG_SYSTEM sideband (`JP_CH:S/X` messages from UNIT_CHANNEL_SPELL UpdateField transitions). This works for all channeled spells (AM, Mind Flay, Drain Life, Blizzard, etc.) and gives instant start + exact per-rank duration.

The UpdateField-triggered SMSG_SPELL_CHANNEL_START synthesis in this PR was tested and doesn't work — the 1.14 client silently ignores SPELL_CHANNEL_START for non-self casters.

Please rebase any remaining non-channel changes onto current master (`be8c824` / v5.0.3-beta.2), or this PR can be closed if the channel work was the main goal.

jameopotato:
Channel cast bars shipped via CHAT_MSG_SYSTEM sideband in v5.0.3-beta.1. The SMSG_SPELL_CHANNEL_START synthesis approach was tested and the 1.14 client rejects it for non-self. Closing as superseded.

PR #65, opened 2026-04-27
 Fix SpanWriter caps for populated zones and demote chat log spam

Mongrul:
Three small changes that reduce GC pressure and console noise on busy
  sessions, with no behavior change visible to the player.

  MonsterMove MaxSplinePoints 16 -> 48 (MovementPackets.cs:69)
    City-guard patrols in Stormwind/Org generate splines longer than 16
    points, which tripped the SpanWriter fallback path on every login
    into a populated zone. New cap puts max packet at 878 bytes, well
    inside ArrayPool's 1024-byte bucket.

  SendSpellHistory MaxEntries 64 -> 128 (SpellPackets.cs:311)
    Level-60 chars with full cooldown rotations exceeded 64 entries on
    every login, falling back to ByteBuffer. New cap (max 3204 bytes)
    is comfortably under the 65535-byte legacy packet cap.

  ChatHandler.cs:374 Console.WriteLine -> Log.Print(LogType.Debug, ...)
    Every outbound chat message - including silent addon-comm
    heartbeats from Attune/TMW/AtlasLoot/etc - was unconditionally
    printed to the proxy console, drowning out warnings in user
    diagnostic bundles. Demoting to Debug keeps the trace available
    when needed without polluting normal sessions. The packet send
    path on the next line is untouched, so addon comms continue to
    work byte-for-byte the same.

  Updated MonsterMoveTests fallback test to use 64 points (above the
  new 48 cap) so the -1 contract stays covered.

  Smoke-tested with a 60s login/move/cast session: 1032 MonsterMove
  packets handled with zero SpanMiss events. Full test suite 312/312
  passing.

 

  PR description

  ## Summary

  Three small changes that quiet noise on busy sessions without changing
  player-visible behavior.

  - **MonsterMove** `MaxSplinePoints` 16 → 48: stops the SpanWriter fallback storm on every populated-zone login (city-guard patrol splines exceed 16 points)
  - **SendSpellHistory** `MaxEntries` 64 → 128: stops the per-login fallback for level-60 cooldown rotations
  - **ChatHandler** `Console.WriteLine` → `Log.Print(LogType.Debug, ...)`: stops every outbound chat message (including silent addon-comm heartbeats) from polluting user diagnostic bundles. Packet send path is unchanged — addon comms still work identically.

  Cap bumps stay well within ArrayPool's bucket sizes (878 / 3204 bytes).

  ## Test plan

  - [x] Full unit-test suite passes (312/312)
[...]

PR #69, opened 2026-04-27
More interrupt fixes

Mongrul:
## Summary

  Follow-up to #65 (mob cast-bar dismiss). The same fix doesn't cover PvP:
  when you Counterspell/Kick an enemy player, their cast bar keeps filling
  on the target frame until the enemy moves, at which point an unrelated
  movement-driven dismiss path catches it.

  ## Root cause

  Last night's fix synthesized `SMSG_SPELL_INTERRUPT_LOG` +
  `SMSG_CANCEL_SPELL_VISUAL` only inside `HandleSpellFailedOther`. PvP
  interrupts arrive on a different opcode.

  In MaNGOS-style 1.12 servers (Kronos), `Spell::SendInterrupted` broadcasts
  `SMSG_SPELL_FAILURE` to all nearby observers via `SendMessageToSet(true)`,
  not just the caster. So when an enemy player gets interrupted, the proxy
  receives `SMSG_SPELL_FAILURE` for THEIR cast, runs `HandleSpellFailure`,
  and falls into the `else` branch (caster isn't our player or our pet).

  Two bugs in that branch:

  1. It minted a deterministic CastID (`spellId + casterCounter`) instead
     of recalling the unique CastID stored at SPELL_START in
     `OtherCasterActiveCastIds`. The `SpellFailure` packet's CastID
     therefore didn't match the cast bar the modern client was tracking,
     so the bar didn't reset.

  2. It never synthesized the modern interrupt packets
     (`SpellInterruptLog`, `CancelSpellVisual`) — those were only emitted
     from the FAILED_OTHER handler. So neither dismiss path fired.

  Result: cast bar fills visually until the next movement update on the
  enemy unit hits the modern client's movement-cancel path.

  ## Fix

  In `HandleSpellFailure`'s non-player/non-pet branch:

  - Pull the unique CastID from `OtherCasterActiveCastIds.TryRemove` —
    same dict populated by SPELL_START in `HandleSpellStartOrGo`. Falls
    back to the deterministic seed only if no entry was tracked.
  - When `reason == 61 (Interrupted)` and we successfully consumed the
    active-cast entry, synthesize `SpellInterruptLog` and
    `CancelSpellVisual` exactly like the FAILED_OTHER handler does.

  The `foundActiveCastId` gate is the dedup mechanism: if Kronos sends
  BOTH `SMSG_SPELL_FAILURE` and `SMSG_SPELL_FAILED_OTHER` for the same
  interrupt, whichever handler runs first removes the dict entry; the
  second handler sees no entry and skips synthesis. Single fire either
  way.

  `HandleSpellFailedOther` is intentionally untouched — the mob fix from
  #65 keeps working unchanged.

  ## Test plan

  - [ ] Counterspell an enemy mage casting on Kronos PvP — target-frame
        cast bar dismisses immediately (no fill-to-completion delay)
  - [ ] Pummel/Kick an enemy player casting — same
  - [ ] Counterspell a mob — still works (existing fix unaffected)
  - [ ] Get Counterspelled yourself — your own cast bar dismisses (local-
        player branch unchanged)
  - [ ] `spell.failure.routed` JSONL events show `sentInterruptLog: true`
        and `foundActiveCastId: true` for PvP interrupts

  Resolves the lingering cast bar follow-up reported after #65.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -439,6 +439,7 @@ public partial class WorldClient
         uint spellVisual;
         bool dequeued = false;
         bool wasStarted = false;
+        bool foundActiveCastId = false;
         if (GetSession().GameState.CurrentPlayerGuid == casterUnit &&
             GetSession().GameState.TryDequeuePendingNormalCast(spellId, out var pendingNormal))
         {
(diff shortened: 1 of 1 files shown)


jameopotato:
Merged in v5.0.3-beta.2 via cherry-pick of `7674a25` onto master (`be8c824`). The base commit `9bf37d7` (SpellCastLogData + SPELL_EXECUTE_LOG) was already in the tree from a prior merge. PvP interrupt dismiss smoke-tested. Nice fix on the OtherCasterActiveCastIds lookup for the broadcast SPELL_FAILURE path.

jameopotato:
Merged in v5.0.3-beta.2 via cherry-pick of 7674a25.

PR #70, opened 2026-04-28
Opposite faction communication

Mongrul:
## Summary

  Fixes HermesProxy issues #100 and #213. As a Horde player, you could read
  Alliance speech (and vice versa) regardless of language. As any race, you
  could read every faction-specific and racial language as plain English.

  Resolves #100 — "able to read enemy languages"
  Resolves #213 — "Unknown Languages appear as plain text"

  ## Root cause

  Vanilla 1.12 emulators (Kronos, vmangos, cmangos) ship plain text + a
  language ID over `SMSG_CHAT` and rely on the legacy 1.12 client to
  scramble unknown languages locally — that's how vanilla worked. The
  modern 1.14 Classic client doesn't honor that contract for legacy
  language IDs, so the proxy was forwarding plain English to receivers
  who shouldn't be able to read it.

  The proxy's chat handlers passed the `Language` field through unchanged
  to `ChatPkt` with no scrambling step anywhere in the pipeline.

  ## Fix

  Reproduce the legacy client's scrambling proxy-side using the **exact
  syllable tables baked into the original WoW client** (extracted via
  Hoizame's `WoW_ClassicUIResources/RawData/LanguageWords.lua` dump —
  1,482 syllables across the 13 vanilla language IDs).

  New file `HermesProxy/World/LanguageScrambler.cs`:

  - `CanUnderstand(race, class, language)` — vanilla racial language
    matrix. Alliance races: Common + their racial. Horde races: Orcish +
    their racial. Universal/addon channels always understood. Warlocks
    always understand Demonic.
  - `Scramble(text, language)` — replaces alphabetic runs in the chat body
    with deterministically-chosen syllables from the language's authentic
    table. Same input always produces the same output (FNV-1a 32-bit hash
    of the lowercased word seeds the syllable picks), matching vanilla
    client behavior where players learn to recognize repeated phrases.
  - Pipe-escape sequences (`|c…`, `|H…|h`, `|T…|t`, `|h`, `|r`, `|n`)
    copy through verbatim so item link wire data and color codes stay
    intact. Display names inside `[…]` brackets are still scrambled
    (which is what vanilla does — clickable link, gibberish name).
  - Output length matches input length exactly.

  Loaded via `GameData.LoadEverything()`'s parallel-invoke at startup.

  `HandleServerChatMessageVanilla` and `HandleServerChatMessageWotLK`
  both call `MaybeScrambleForeignLanguage(text, language)` after the
  existing addon-prefix check and before constructing the outgoing
  `ChatPkt`. That helper:

  - Reads the local player's race/class from `GameState.CurrentPlayerInfo`
  - If the receiver understands the language, passthrough.
  - If no syllabary is loaded for the language, passthrough + log
    `chat.scramble.skipped` with the language ID.
  - Otherwise scramble + log `chat.scramble.applied` with original
    length, scrambled length, receiver race, and receiver class for
    diagnostic bundles.

  ## What this does NOT change

  - **No CMSG-side filtering.** Outgoing chat from the modern client
    forwards to Kronos with its language ID intact; Kronos already
    enforces faction restrictions on the server side.
  - **No changes to mob/system chat.** `MonsterSay/Yell/Whisper`,
    `RaidBossEmote`, achievement broadcasts, etc. typically use
    `Language.Universal` and pass through unchanged.
  - **No level-50 gating on Demonic.** Vanilla actually gates Demonic
    comprehension on Warlock level 50 (Ritual of Doom quest); we grant
    it to any Warlock. Over-permissive is the safer default — no
    warlock will see scrambled summon-channel chat and panic.
  - **No new languages handled.** Draenei (35) doesn't appear in the
    1.12 syllabary dump; if Kronos somehow sends it, we fall through to
    passthrough and log `chat.scramble.skipped`. Vanilla shouldn't ever
    emit it.

  ## Files

  - `HermesProxy/CSV/LanguageWords.csv` (new) — 1,482 syllables, the
    vanilla client's actual data
  - `HermesProxy/World/LanguageScrambler.cs` (new) — loader + matrix +
    scrambler
  - `HermesProxy/World/GameData.cs` — one line added to
    `LoadEverything()` parallel-invoke
  - `HermesProxy/World/Client/PacketHandlers/ChatHandler.cs` — both
    vanilla and WotLK chat handlers call the new helper


jameopotato:
Not yet reviewed. Please rebase onto current master (`be8c824` / v5.0.3-beta.2) if there are conflicts. Phase 2 review coming soon.

PR #71, opened 2026-04-28
Fix mount/item re-click leaving cast active server-side after UI dismiss

Mongrul:
## Summary

  Tester report: clicking a mount item, then re-clicking while casting,
  visually dismissed the cast bar and stopped the local animation — but
  the sound queue kept playing and the mount applied a few seconds later.
  Diagnosed from JSONL packet trace + reproduced fix verified in-game.

  ## Root cause

  Two related issues compounded.

  **1. Wrong opcode handler.** The mount under test is an item (summoning
  tome), so the modern client sends `CMSG_USE_ITEM`, not `CMSG_CAST_SPELL`.
  `HandleUseItem` had no in-flight cast gate at all — every USE_ITEM was
  enqueued and forwarded to Kronos unconditionally.

  **2. FIFO dequeue mismatch.** Because the second USE_ITEM also enqueued
  into `PendingNormalCasts`, the queue contained two entries with the same
  SpellId — the in-flight cast (HasStarted=true) and the duplicate
  (HasStarted=false). When Kronos rejected the duplicate with
  `SMSG_SPELL_FAILURE`, `HandleSpellFailure`'s
  `TryDequeuePendingNormalCast(spellId)` dequeued the **oldest** match,
  which was the in-flight cast. The modern client then received
  `SpellFailure` bound to the first cast's IDs → cast bar dismissed and
  animation stopped. But Kronos's original mount cast was never actually
  cancelled, so it completed normally and the mount aura applied — out of
  sync with what the user just saw on screen.

  JSONL packet trace from the bug bundle (timestamps in ms):

  186723  CMSG_USE_ITEM      (1st click)
  186898  SMSG_SPELL_START   (cast bar visible, animation starts)
  186933  CMSG_USE_ITEM      (2nd click — user wants to cancel)
  187100  SMSG_SPELL_FAILURE (Kronos rejected the duplicate)
  187104  SMSG_CAST_FAILED   (...)
  187143  CMSG_USE_ITEM      (3rd click — spam continues)
  ... rejection pattern repeats; original cast still active server-side

  ## Fix

  Two parallel branches, same pattern:

  - **`HandleCastSpell`**: when an in-flight normal cast exists and the new request's SpellID matches it, send `CMSG_CANCEL_CAST` to Kronos before falling through to the existing `SendCastRequestFailed`. The resulting `SMSG_SPELL_FAILURE` from Kronos cancels the cast properly and dismisses the bar via the legitimate failure path. Different-spell duplicates (e.g. Heal while casting Frostbolt) keep the original reject behavior — that's still "Another action is in progress" and matches vanilla.

  - **`HandleUseItem`**: same pattern, plus matching by `LegacySpellId` for SoM-renumbered items (e.g. Diamond Flask 17626 → 363880). On match: send `CMSG_CANCEL_CAST`, fail the duplicate locally via `SendCastRequestFailed`, **don't enqueue and don't forward the second USE_ITEM** — the second forward + Kronos rejection is exactly what caused the FIFO mismatch in the first place.

  New helper `GameSessionData.GetStartedNormalCast()` — returns the
  in-flight (HasStarted=true) cast or null. Same iteration cost as the
  existing `HasStartedNormalCast()`; consolidates the lookup so both
  handlers can compare the new request's SpellId against the running
  cast.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -698,6 +698,21 @@ public sealed class GameSessionData
     }
 
     /// <summary>
+    /// Returns the in-flight (started) normal cast if any. Used by HandleCastSpell to
+    /// detect "user re-clicked the spell they're already casting" and translate that
+    /// into a server-side cancel rather than silently rejecting the duplicate press.
+    /// </summary>
+    public ClientCastRequest? GetStartedNormalCast()
+    {
+        foreach (var item in PendingNormalCasts)
+        {
+            if (item.HasStarted)
+                return item;
+        }
+        return null;
+    }
+
+    /// <summary>
     /// Clear only pending normal casts that haven't started yet.
     /// Keeps started casts so SPELL_GO can dequeue them later.
     /// Returns the cleared casts so they can be failed.
--- a/HermesProxy/HermesProxy.config
+++ b/HermesProxy/HermesProxy.config
@@ -1,122 +1,23 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <appSettings>
-    <!--
-            Option:      ClientBuild
-            Description: The version of the client you will play with.
-            Default:     "40618" (1.14.0)
-                         "41794" (1.14.1)
-                         "42597" (1.14.2)
-                         "40892" (2.5.2)
-                         "42328" (2.5.3)
-    -->
-    <add key="ClientBuild" value="40618" />
-    <!--
-            Option:      ClientSeed
-            Description: The auth seed of the client you will play with.
-            Default:     "179D3DC3235629D07113A9B3867F97A7" (static)
-    -->
-    <add key="ClientSeed" value="179D3DC3235629D07113A9B3867F97A7" />
-    <!--
-            Option:      ServerBuild
-            Description: The version of the server you want to connect to.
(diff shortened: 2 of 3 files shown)


jameopotato:
Not yet reviewed. Please rebase onto current master (`be8c824` / v5.0.3-beta.2) if there are conflicts. Phase 2 review coming soon.

jameopotato:
Took a different approach to this bug. Instead of cancelling the in-flight cast on re-click, we silently drop the re-click entirely — the cast bar stays up and the original cast completes normally (vanilla behavior).

The 1.14 client dismisses the cast bar when it receives `CastFailed(SpellInProgress)`, even though the CastIDs differ. Sending no response at all keeps the bar intact.

Changes in `8998c39`:
- `HandleCastSpell`: changed `SendCastRequestFailed` to silent `return` when `HasStartedNormalCast()` is true
- `HandleUseItem`: added the same guard (was completely unprotected — the mount re-click bug)

This preserves vanilla semantics where the original cast always completes. Your cancel approach would work too, but changes behavior from what 1.12 clients experienced.

Closing in favor of `8998c39` on master. Thanks for the detailed bug report — the UseItem path having no guard at all was a good catch.

jameopotato:
Fixed via silent-drop approach in 8998c39 — see comment above.

PR #72, opened 2026-04-28
Fix GCD locking for full duration on LoS / mid-cast spell failures

Mongrul:
## Summary

  Tester report: casting a spell on a target behind a wall — the cast doesn't
  go off (correct), but the action button greys out for the full 1.5s GCD as
  if the spell had completed. Reproduces on both instant casts and cast-time
  spells like Frostbolt. Verified fixed in-game.

  ## Root cause

  Two separate mechanisms compounded, both rooted in how vmangos / Twinstar
  emits failure packets and how the modern 1.14 Classic client interprets
  them.

  **1. SMSG_SPELL_START leaks "cast committed" for instants.**

  For instant spells, vmangos still emits `SMSG_SPELL_START` *before* running
  its LoS / range / target validation. The modern client processes the START,
  commits the action-bar GCD anticipation as locked, and then refuses to roll
  it back when the subsequent `SMSG_CAST_FAILED` arrives ~200ms later. Once
  SPELL_START is processed, GCD is latched.

  **2. SMSG_SPELL_FAILURE has "in-flight interrupt" semantics on the modern
  client.**

  vmangos `Spell::SendInterrupted()` hardcodes the SMSG_SPELL_FAILURE reason
  byte to `0` (a generic "interrupted" signal) and follows it ~0-5ms later
  with `SMSG_CAST_FAILED` carrying the real reason. On the modern 1.14
  client:

  - SMSG_SPELL_FAILURE for the local caster → "in-flight cast was interrupted" → action-bar GCD stays committed (vanilla rule: cast went into flight, GCD applies)
  - SMSG_CAST_FAILED for the local caster → "pre-cast failure" → dismisses cast bar AND cancels GCD anticipation

  The SpellFailure handler runs first and *dequeues* the pending cast entry,
  so when CastFailed arrives 5ms later `TryDequeuePendingNormalCast` returns
  false — no CastFailed packet is ever forwarded to the client. The modern
  client only sees the SpellFailure with its GCD-lock semantics, and the
  proper cancel signal is silently dropped.

  JSONL packet trace from the bug bundle (Frostbolt, spell 116):

  165477  CMSG_CAST_SPELL
  165611  SMSG_SPELL_START   (cast bar appears)
  166410  SMSG_SPELL_FAILURE (~800ms mid-cast — wasStarted=true)
  166410  SMSG_CAST_FAILED   (5ms later — proxy can't dequeue, drops it)
          → modern client: GCD locked for full duration

  ## Fix

  Two parallel changes in `HermesProxy/World/Client/PacketHandlers/SpellHandler.cs`:

  **`HandleSpellStart`**: when the caster is the local player or local pet
  *and* `CastTime == 0`, do all internal bookkeeping (mark queue entry
  started, send `SpellPrepare` to bind IDs, clear non-started casts) but
  skip the `SendPacketToClient(spell)` that forwards the SPELL_START. The
  client never sees a server-side "cast committed" signal for instants, so
  GCD anticipation stays in soft-anticipation state rather than committing.
  Successful instants still apply real GCD via SPELL_GO. Cast-time spells
  (`CastTime > 0`) and NPC casts forward as before — for them the cast bar
  is needed and GCD has genuinely been committed server-side.

  **`HandleSpellFailure`**: when the caster is the local player or local
  pet, early-return without dequeuing or forwarding. `HandleCastFailed`
  then becomes the canonical caster-side signal — it dequeues the queue,
  sends `SpellPrepare` + `CastFailed` with the real reason, dismisses the
  cast bar, and cancels GCD anticipation cleanly. Other casters (mobs,
  remote players) keep the existing path so the mob cast-bar dismiss fix
  from PR #65 still works for Counterspell / Kick on enemy mobs.

  ## Diagnostics

  - `spell.start.instant_suppressed` — `{ spell_id, cast_time_ms, caster_is_player, caster_is_pet }`
  - `spell.failure.suppressed_for_caster` — `{ spellId, raw_reason_byte, is_pet }`

  Both fire on the new branches so the next bug bundle confirms the fix is
  engaging.

  ## Files

  - `HermesProxy/World/Client/PacketHandlers/SpellHandler.cs` — both `HandleSpellStart` and `HandleSpellFailure` now early-return for caster-side cases that should defer to SPELL_GO / CAST_FAILED

### DIFF
--- a/HermesProxy/HermesProxy.config
+++ b/HermesProxy/HermesProxy.config
@@ -1,122 +1,23 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <appSettings>
-    <!--
-            Option:      ClientBuild
-            Description: The version of the client you will play with.
-            Default:     "40618" (1.14.0)
-                         "41794" (1.14.1)
-                         "42597" (1.14.2)
-                         "40892" (2.5.2)
-                         "42328" (2.5.3)
-    -->
-    <add key="ClientBuild" value="40618" />
-    <!--
-            Option:      ClientSeed
-            Description: The auth seed of the client you will play with.
-            Default:     "179D3DC3235629D07113A9B3867F97A7" (static)
-    -->
-    <add key="ClientSeed" value="179D3DC3235629D07113A9B3867F97A7" />
-    <!--
-            Option:      ServerBuild
-            Description: The version of the server you want to connect to.
-            Default:     "auto" will choose best version based on ClientBuild
-                         "5875" (1.12.1)
-                         "8606" (2.4.3)
-    -->
-    <add key="ServerBuild" value="auto" />
-    <!--
-            Option:      ServerAddress
-            Description: The address of the server you want to connect to.
-                         aka. what you use in "SET REALMLIST <ip>"
-            Default:     "127.0.0.1" (localhost)
-    -->
-    <add key="ServerAddress" value="127.0.0.1" />
-    <!--
-            Option:      ServerPort
-            Description: The port to reach the server you want to connect to.
-            Default:     "3724"
-    -->
-    <add key="ServerPort" value="3724" />
-    <!--
-            Option:      ReportedOS
-            Description: The OS identifier that will be sent to the remote server.
-            Default:     "OSX"
-    -->
-    <add key="ReportedOS" value="OSX" />
-    <!--
-            Option:      ReportedPlatform
-            Description: The platform identifier that will be sent to the remote server.
-            Default:     "x86"
-    -->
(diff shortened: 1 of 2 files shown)


jameopotato:
Merged in v5.0.3-beta.2 via cherry-pick onto master (`4606d9f`). HermesProxy.config excluded (personal settings — same issue as the revert in `2781058`). Please keep config changes out of future PRs. SpellHandler.cs changes smoke-tested — GCD no longer locks on LoS failures.

jameopotato:
Merged in v5.0.3-beta.2 via cherry-pick of dace1b9 (config excluded).

PR #75, opened 2026-04-28
Fix quest item toast: persist progress across logout, relog, and proxy restart

Mongrul:
This wasnt working after relog so we took a better approach.  

## Summary
  The over-head modern quest toast was resetting to "1/N" instead of continuing from the actual
  progress in three scenarios on Kronos 1.12. This PR (two commits) fixes all three and adds disk
  persistence so the running total survives even a launcher-induced proxy kill on game close.

  ## Bugs fixed
  1. **Logout to charselect → relog (in-proxy)**: the legacy server only sends per-pickup deltas
  (not totals), and `SMSG_LOGOUT_COMPLETE` recreates `GameState`, wiping the proxy-internal
  running totals. Toast read 1/N on next pickup instead of resuming from prior count.
  2. **Off-by-one for unfamiliar quest items**: when `SMSG_QUEST_UPDATE_ADD_ITEM` arrived before
  the QuestTemplate was cached, the credit was lost (legacy server only sends each credit once).
  First pickup of an unfamiliar quest item permanently undercounted by 1.
  3. **Stale running total after re-accept**: `QuestItemObjectiveProgress` wasn't cleared on
  COMPLETE/FAILED/FAILED_TIMER/abandon, so re-accepting the same quest inherited the prior
  session's count.
  4. **Progress lost when launcher kills proxy on game close**: in-memory snapshot dies with the
  proxy. Most user installs run with kill-proxy-on-logout enabled, so the original fix didn't help
   in real-world flows.

  ## Root cause notes
  - Cache-seed fallback (read progress from `PLAYER_QUEST_LOG_x_2`) is permanently dead on Kronos:
   the legacy update-field cache populates the slot's `QuestID` but never the per-objective
  progress byte. `quest.item.seed.attempt` events confirm `matched_slot != -1` with
  `matched_slot_progress: null` every time. The proxy must own this state.
  - Switched item-credit packet from `QuestUpdateAddCreditSimple` to full `QuestUpdateAddCredit`
  with explicit `Count` / `Required`. SIMPLE makes the modern client manage its own counter, which
   (a) starts at 0 on every relog and (b) doesn't get re-seeded from `PLAYER_QUEST_LOG_x_3` on
  login.

  ## Fix architecture
  **In-memory layer** (commit 838ca8a):
  - `GlobalSessionData.SavedQuestItemProgressByCharacter` keyed by `WowGuid128` so Char A→B→A
  doesn't cross-contaminate
  - `SnapshotQuestItemProgressForRestore()` runs in `HandleLogoutComplete` AND `OnDisconnect`
  (Wi-Fi-drop path)
  - `EnsureQuestItemProgressRestored()` runs at the top of `ProcessQuestItemCredit`, idempotent
  per `GameSessionData` instance via `ReferenceEquals` (relog produces a fresh `GameState` even
  with same playerGuid — a guid-keyed guard would skip the restore)
  - COMPLETE/FAILED/FAILED_TIMER/abandon clear from both live and saved dicts

  **Buffered-credit layer** (commit 838ca8a):
  - `PendingQuestItemCredits` buffers credits whose QuestTemplate isn't cached yet
  - `ReplayPendingQuestItemCredits` runs after `StoreQuestTemplate` in
  `HandleQueryQuestInfoResponse`
  - `DropPendingQuestItemCreditsForQuest(questId)` clears buffered entries whose itemId belongs to
   a quest leaving the log, so a deferred replay can't misattribute count to a re-accept

  **Disk layer** (commit 373ac0c):
  - Per-character JSON at `AccountData/<account>/<realm>/<character>/quest-item-progress.json`,
  schema-versioned, atomic write (temp + rename)
  - Eager write: `ProcessQuestItemCredit` persists after every pickup, not batched-on-logout, so a
   launcher-induced kill loses at most one in-flight credit
  - Lazy load on first `EnsureQuestItemProgressRestored` per session
  - Static lock around the file write (server thread on abandon vs WorldClient thread on credit)
  - Cleared on COMPLETE/FAILED/abandon in both client and server handlers

  **Concurrency hardening** (commit 373ac0c):
  - `QuestItemObjectiveProgress` and `SavedQuestItemProgressByCharacter` promoted to
  `ConcurrentDictionary` (cross-thread mutation: server thread on abandon vs WorldClient thread on
   credit/COMPLETE)
  - `PendingQuestItemCredits` accesses guarded by a dedicated lock object
  - `_lastRestoredForGameState` marked `volatile`

  ## Diagnostics added
  - `quest.progress.snapshot` — fired on logout/disconnect, includes `entries` count and
  `persisted` flag
  - `quest.progress.disk.loaded` — fired when disk file is read on first session-startup credit
  - `quest.progress.restored` — fired when in-memory or disk-loaded state is copied to the new
  GameState
  - `quest.progress.restore.empty` — fired when no saved entries exist for the current playerGuid
  - `quest.item.seed.attempt` — captures `update_fields_present`, `matched_slot`,
  `matched_slot_progress` for the dead Kronos cache-seed fallback
  - `quest.item.add` — full per-credit context: `stored_before`, `running_total`,
  `seeded_from_cache`, `clamped`, `replayed`
  - `quest.item.add.buffered` — credit deferred because QuestTemplate not yet cached
  - `quest.item.pending.dropped` — buffered credit dropped because its quest is leaving the log

  ## Test plan
  - [x] In-game test: pick up 2/8 → logout to charselect → log back in → pick up 1 → toast reads
  3/8
  - [x] In-game test: pick up 2/N → close game (launcher kills proxy) → relaunch → log in → pick
  up 1 → toast continues from prior total. JSON file at
  `AccountData/<acc>/<realm>/<char>/quest-item-progress.json` shows the saved count after kill
  - [x] In-game test: abandon-then-logout-then-relog-then-re-accept doesn't restore stale entries
  - [x] Build clean (0 warnings, 0 errors)
  - [x] Test suite green (312/312)
  - [x] Bug-hunter review pass complete; concurrency, lifecycle, edge cases addressed

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -11,6 +11,7 @@ using System.Runtime.InteropServices;
 using System.Linq;
 using System.Text;
 using System.Threading;
+using Framework.Logging;
 using Framework.Realm;
 using HermesProxy.World.Server.Packets;
 using ArenaTeamInspectData = HermesProxy.World.Server.Packets.ArenaTeamInspectData;
@@ -127,8 +128,17 @@ public sealed class GameSessionData
     // MSG_CHANNEL_START to the caster, not to nearby players).
     public ConcurrentDictionary<WowGuid128, int> UnitChannelSpells = new();
     public WowGuid64 LastLootTargetGuid;
-    public Dictionary<(uint QuestID, sbyte StorageIndex), uint> QuestItemObjectiveProgress = new(); //MIRASU - proxy-local running totals for quest item pickups; legacy update-field cache isn't refreshed on partial updates, so we track ourselves
-    public List<(uint ItemId, uint Count)> PendingQuestItemCredits = new(); //MIRASU - SMSG_QUEST_UPDATE_ADD_ITEM credits received before the QuestTemplate was cached; replayed on QueryQuestInfoResponse so the toast doesn't drop the first pickup of an unfamiliar quest item
+    //MIRASU - ConcurrentDictionary because abandon-clear runs on the modern-server thread
+    //MIRASU   (CMSG_QUEST_LOG_REMOVE_QUEST handler in Server/QuestHandler.cs) while item-credit
+    //MIRASU   reads/writes and COMPLETE/FAILED clears run on the WorldClient thread. Plain
+    //MIRASU   Dictionary risks torn-state corruption on cross-thread enumeration.
+    public ConcurrentDictionary<(uint QuestID, sbyte StorageIndex), uint> QuestItemObjectiveProgress = new();
+    //MIRASU - PendingQuestItemCredits is a List (no concurrent equivalent supports predicate-remove).
+    //MIRASU   Guard with PendingQuestItemCreditsLock for cross-thread safety. Cleared on
+    //MIRASU   COMPLETE/FAILED/abandon for the affected quest's item objectives so a stale buffered
+    //MIRASU   credit can't be replayed against a re-accept (or a different quest sharing the item).
+    public List<(uint ItemId, uint Count)> PendingQuestItemCredits = new();
+    public readonly object PendingQuestItemCreditsLock = new();
     public uint CurrentLootCoins; //MIRASU - remembers coin amount from SMSG_LOOT_RESPONSE so proxy can synthesize SMSG_LOOT_MONEY_NOTIFY when client picks up gold (Kronos/TC-1.12 doesn't emit it)
     public List<WowGuid128>? MasterLootCandidates;
     public WowGuid64 LastMasterLootSentTarget;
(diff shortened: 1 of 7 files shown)


TakaNoDan:
**## Test 6 — Quest item toast persistence (PR #75)**

- [x] Close the game entirely (launcher kills proxy) → relaunch → collect 1 more → toast continues from where you left off

 ---------------------

_The first quest item you loot after doing that, will display 1/X of progress in the middle of the screen when picked up, regardless of how many are present in your inventory. Starting from the following quest item looted after that, the correct number will then be shown.
The quest tracker to the right side of the screen will show progress correctly the whole time.

Tested all other ## Test 6 — Quest item toast persistence (PR #75) options, everything else seems to work._

Zone: Redridge Mountains
Quest: Blackrock Menace
Objective: Collect 10 Battleworn Axes

PR #76, opened 2026-04-28
Distract dc fix

Mongrul:


jameopotato:
Closed — superseded by #80, which is a strict superset (adds effects 3/50/77, breadcrumb diagnostics, structured Log.Event on parse failure) on top of this branch's effect-69-only fix. Merged via #80.

PR #84, opened 2026-04-29
Zombie session dc fix

Mongrul:
- 8 chat-message handlers (`AFK`, `DND`, `Channel`, `Whisper`, `Emote`, the multiplexed
  `Say/Yell/Guild/Officer/Party/Raid/RaidWarning/InstanceChat`, `AddonMessage`, `AddonMessageTargeted`) now early-return
   when `GetSession().WorldClient == null` instead of dereferencing it.
  - New helper `LogChatDroppedNoWorldClient(opcode)` emits a structured `packet.dropped` event with `reason:
  "no_worldclient"` so the zombie window is visible in JSONL bundles.

  ## Why

  User bundle (`jimsproxy-20260429-153625.jsonl`) showed `System.NullReferenceException at
  WorldSocket.HandleChatMessageAFK` ~37 minutes after a `worldclient_legacy_disconnect` (legacy TCP closed cleanly after
   `SMSG_UPDATE_OBJECT` 169 from Kronos PTR). The realm-swap fix nulls `session.WorldClient` and suppresses propagation
  to the modern client; subsequent chat opcodes then NRE on `GetSession().WorldClient!.SendMessageChat*`. The `!` only
  silences the compiler warning — it does not prevent the runtime NRE.

  The dispatch try/catch already swallows the throw (`WorldSocket.cs:374`), so this isn't a crash — it's silent packet
  loss with one `packet.error` event per chat message during the zombie window. Adding the guard makes the loss explicit
   and matches the existing pattern in `HandleChatJoinChannel`/`HandleChatLeaveChannel`, which already null-check before
   sending.

  ## Scope notes

  - Doesn't address the underlying zombie session (legacy DC suppressed while in-world, modern client never told). That
  requires gating the realm-swap suppression on session state and propagating in-world legacy DCs as a clean
  modern-client kick — tracked separately.
  - `JoinChannel`/`LeaveChannel` already had the null check (without diagnostics) — left alone to keep the diff focused.

  ## Test plan

  - [x] `dotnet build HermesProxy` — clean, 0 warnings, 0 errors
  - [x] `dotnet test HermesProxy.Tests` — 312/312 passing
  - [ ] Live session: induce legacy-side DC mid-session, then send `/afk` from modern client → confirm no NRE in logs,
  `packet.dropped` with `reason: "no_worldclient"` appears in JSONL bundle


jameopotato (approved):
**Approve** — clean, minimal, correctly-scoped fix.

Null guards on 8 chat handlers prevent NREs during the zombie session window (legacy DC while modern client is still connected). The `LogChatDroppedNoWorldClient` helper emits structured `packet.dropped` events making the zombie window visible in JSONL bundles.

Pattern matches existing null-check guards in `HandleChatJoinChannel`/`HandleChatLeaveChannel`. Scope is explicitly limited — underlying zombie session root cause acknowledged as a separate tracked issue.

Note: PR #85 is a superset of this PR (same branch base + one more commit). Merging #85 makes this PR redundant. If merging separately, merge #84 first.

jameopotato:
Closing — PR #85 is a superset of this PR (same branch base + quest toast deferred replay commit). All changes from #84 are included in #85. The unique chat null-guard commit will merge via #85.

PR #85, opened 2026-04-30
Quest toast first loot after relog

Mongrul:
After a full proxy close/relaunch, the first pickup of an in-progress quest item rendered the modern over-head quest
  toast as "1/N" briefly, then corrected to the right count ~150ms later. Bundle-driven investigation across 4 sessions
  narrowed it down through three failure layers and produced a deferred-replay fix that survives the buffered-credit
  scenario.

  ## Root cause

  In Kronos's pickup burst order, `SMSG_QUEST_UPDATE_ADD_ITEM` arrives first and gets buffered (template not cached
  yet). `SMSG_ITEM_PUSH_RESULT` arrives ~3ms later — and the proxy's `QuantityInInventory` computation falls through its
   early `if (objective != null)` guard, because `GameData.GetQuestObjectiveForItem(itemID)` returns null until the
  template arrives via `SMSG_QUERY_QUEST_INFO_RESPONSE`. Net effect: `QuantityInInventory` defaulted to `item.Quantity`
  (delta=1). The modern client used that for the over-head toast, rendering "1/N" until the buffered credit replayed.

  ## Fix

  **`Client/PacketHandlers/ItemHandler.cs` + `Client/PacketHandlers/QuestHandler.cs` + `GlobalSessionData.cs`**

  Defer `ITEM_PUSH_RESULT` forwarding when its objective lookup fails AND a buffered credit exists for the same
  `itemId`. The fully-built packet is held in a new `PendingItemPushResults` list. `ReplayPendingQuestItemCredits`
  drains the list after each credit replays — by that point the template is cached, the dict has the running total, the
  objective lookup now works, and `QuantityInInventory` is recomputed correctly before send. Items whose objective still
   doesn't resolve are re-buffered for the next `QUERY_QUEST_INFO_RESPONSE`.

  ## Two correctness fixes uncovered along the way

  **`GlobalSessionData.cs` — `EnsureQuestItemProgressRestored` gate-latch.** The method set `_lastRestoredForGameState =
   GameState` *before* attempting `TryLoadQuestItemProgressFromDisk`, so any silent precondition failure
  (AccountMetaDataMgr null, OwnCharacters/Realm not yet populated) locked out all future restore attempts on that
  GameState. Preconditions now checked before latching, with a `quest.progress.restore.deferred` event surfacing the
  missing precondition for future bundles.

  **`Server/PacketHandlers/CharacterHandler.cs` — eager disk-load on login.** `HandlePlayerLogin` now calls
  `EnsureQuestItemProgressRestored()` right after `CurrentPlayerInfo` is set, so `QuestItemObjectiveProgress` is hot
  before any pickup burst — instead of loading lazily on first credit, which races `SMSG_ITEM_PUSH_RESULT` in Kronos's
  ordering.

  **`ItemHandler.cs` — pending-delta math.** When the dict has the pre-pickup count (eager disk-load worked) but the
  credit is still buffered (template not cached), `QuantityInInventory` now sums `dict[key] + matching pending deltas` =
   post-pickup count.

  ## Diagnostics added

  - `quest.progress.restore.deferred` — fires when `EnsureQuestItemProgressRestored` skips because preconditions aren't
  met (cause field: `account_meta_data_mgr_null`, `char_not_in_own_characters`, `char_info_incomplete`)
  - `item.push.deferred` — fires when an `ITEM_PUSH_RESULT` is held for replay
  - `item.push.replayed` — fires when a held packet is sent after credit replay


jameopotato (dismissed):
**Approve with suggestions** — three improvements worth making before or shortly after merge.

## 1. Consolidate double `GetQuestObjectiveForItem` in `HandleItemPushResult`

`GameData.GetQuestObjectiveForItem(item.Item.ItemID)` is called twice — once at line 72 (existing, for `QuantityInInventory`) and again at line 117 (new, for deferral check). Pass the result from the first call to the deferral block to avoid the redundant lookup. Minor efficiency, but more importantly clarifies the data flow — the reader can see both decisions stem from the same objective resolution.

## 2. Defensive upper bound on `PendingItemPushResults` and `PendingQuestItemCredits`

If a quest template query never returns (pathological server behavior), these lists grow without bound. A cap of ~50 entries with oldest-first eviction would prevent unbounded memory growth. Unlikely to matter in practice, but defensive coding for a component that already handles crash recovery.

## 3. Replace per-pickup snapshot with dirty flag only

`SnapshotQuestItemProgressForRestore()` runs on every `ProcessQuestItemCredit`, allocating a `ConcurrentDictionary` copy each time. Since `FlushQuestItemProgressToDisk` runs on logout and disconnect (the two paths where persistence matters), the per-pickup call could be replaced with just marking `_questProgressDiskDirty = true` and letting the flush paths do the snapshot. Reduces per-pickup allocations from one dict copy to zero. The tradeoff (crash during combat loses last few credits) is already accepted by the 5s debounce window.

---

Implementation is thorough and correct. The gate-latch fix, eager-restore on login, and deferred ITEM_PUSH_RESULT replay together eliminate the post-restart "1/N" flicker completely. Thread safety is well-reasoned. Diagnostic coverage is excellent.

jameopotato:
## Code Review — Approve

**Summary:** First quest item pickup after proxy restart showed "1/N" briefly before correcting. Root cause: SMSG_QUEST_UPDATE_ADD_ITEM arrives before template is cached, so ITEM_PUSH_RESULT can't look up the objective and defaults to delta=1. Three-layer fix: defer ITEM_PUSH_RESULT until credits replay, fix gate-latch precondition ordering, eager disk-load on login.

**Correctness:**
- Deferred ITEM_PUSH_RESULT correctly avoids the template-not-cached race
- Gate-latch fix is a critical correctness improvement — the restore method was permanently locking out retries on precondition failure
- Eager disk-load on HandlePlayerLogin ensures dict is warm before pickup burst
- AccountDataManager: clean per-character disk persistence with atomic write (temp + rename)

**Efficiency:** Deferred packets list is tiny (0-2 during login burst). Disk writes on pickup are infrequent. No hot-path concern.

**Future compatibility:** AccountDataManager is reusable infrastructure. QuestHandler changes are our fork's addition — low Xian55 rebase risk.

**Note:** PR includes duplicate code from already-merged PRs (#70, #81, latency-fixes, HighGuid). Git handles the overlap cleanly on merge.

LGTM — merging to master.

PR #86, opened 2026-04-30
Fix mount/spell action button stuck "lit" after rapid double-click

Mongrul:
The 8998c39 silent-drop guard prevented the FIFO mismatch from
  commit 7b9e8aa, but left the modern client's per-click action-button
  tracking unresolved — no SpellPrepare, SpellGo, or CastFailed was ever
  sent for the dropped duplicate, so the action button sat in "queued"
  state indefinitely.

  Two changes in HandleCastSpell and HandleUseItem:

  1. Add HasInFlightNormalCastForSpell() — catches same-spell duplicates while the first cast is still in flight (forwarded but no SPELL_START yet), which the existing HasStartedNormalCast() gate misses.

  2. On either gate firing, call SendCastRequestFailed(reason: DontReport) bound to the DUPLICATE's unique ClientGUID/ServerGUID instead of silently returning. SpellPrepare binds the per-click tracking and CastFailed releases the queued state. DontReport suppresses the "Another action is in progress" toast — same silent-from-user-perspective semantics as before, but with the IDs the client needs to clear the action-button highlight.

  Cast bar regression risk handled by binding both responses to the
  duplicate's distinct ServerCastID; the running cast's ServerCastID is
  different, so its tracking should be unaffected.

  Reason parameter on SendCastRequestFailed defaults to SpellInProgress
  so existing call sites (NextMelee/AutoRepeat conflict, GCD displacement,
  SPELL_FAILURE drain) keep their current behavior.

  Verified in JSONL bundle: cast.dropped.duplicate fires on click #2
  (reason=in_flight_same_spell_use_item, queue_depth=1), button clears
  correctly, running mount cast bar/aura unaffected.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -702,6 +702,28 @@ public sealed class GameSessionData
     }
 
     /// <summary>
+    /// JimsProxy (Mount-Button-Stuck-Lit): returns true if any pending normal cast — started OR
+    /// merely in flight to the legacy server — matches the given SpellId (or its LegacySpellId
+    /// for SoM-renumbered USE_ITEMs). HasStartedNormalCast only catches duplicates AFTER
+    /// SMSG_SPELL_START matches the queue entry; rapid mashing within the c→s→c round-trip slips
+    /// past it and the duplicate USE_ITEM/CAST_SPELL gets forwarded. The legacy server then rejects
+    /// the duplicate with SpellInProgress, and the FIFO dequeue (TryDequeuePendingNormalCast picks
+    /// the oldest match) binds the failure to the in-flight cast's IDs — the bug commit 7b9e8aa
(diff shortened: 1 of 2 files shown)


jameopotato (approved):
**Approve with suggestions**

Correctly solves the stuck action button problem. `HasInFlightNormalCastForSpell()` catches same-spell duplicates in the pre-SPELL_START window. `SendCastRequestFailed(DontReport)` bound to the duplicate's own IDs releases client tracking without showing an error toast or disrupting the running cast bar. Good parity between HandleCastSpell and HandleUseItem.

**Suggestion:** Add a comment noting the guard ordering dependency with the GCD hold system (beta branch, PR #87). When the GCD hold guards merge, `IsGcdHoldActive()` and `HasForwardedPendingCast()` must run BEFORE `HasInFlightNormalCastForSpell()` so instant-cast spam is held rather than dropped. A one-liner like `// Must run AFTER GCD hold guards when those merge from beta (#87)` above the `gateInFlight` check would prevent accidental misordering during rebase.

PR #87, opened 2026-04-30
Latency fixes: Nagle ignore + GCD double-cast window + RTT-adaptive offset

jameopotato:
## Summary

- **Ignore CMSG_ENABLE_NAGLE** — the 1.14 client sends this when "Optimize Network for Speed" is unchecked, re-enabling Nagle's ~200ms write coalescing on both sockets. The proxy now ignores it and keeps TCP_NODELAY=true permanently. Verified with JSONL: opcode fires when the setting is disabled.

- **Close GCD double-cast window** — `OnGcdTimerElapsed` was clearing `_gcdExpireTimestampMs = 0` when the timer fired, removing GCD protection during the ~RTT window before the server responds. Spam-presses during this window bypassed all guards and sent duplicate casts. Three sub-fixes: keep GCD active after timer fire, add `HasForwardedPendingCast()` guard for the post-expiry window, and fire held casts immediately on server failure.

- **RTT-adaptive GCD fire offset** — measures proxy→server round-trip via CMSG_PING/SMSG_PONG timing (EMA-smoothed, 3-sample warm-up). Fires held casts RTT/2 ms early so they arrive at the server right as the server-side GCD expires. Capped at 100ms. Falls back to static `SpellCastEarlyFireOffsetMs` config during warm-up. Arcane Explosion spam data showed ~1700ms GCD intervals (on 1500ms GCD) with offset=0 — this should reduce to ~1600ms.

## Test plan

- [x] `dotnet build` clean (0 warnings)
- [x] `dotnet test` — 312/312 passing
- [ ] Mash Arcane Explosion on Kronos — verify no double casts in JSONL, one `spell.held_fire` per GCD
- [ ] After 2+ min, check `gcd.begin` events for `fire_offset_ms > 0` and `smoothed_rtt_ms`
- [ ] Check `rtt.sample` events for EMA convergence
- [ ] Verify zero NOT_READY failures during spam
- [ ] LoS fail during GCD spam — held cast fires on failure, player not stuck
- [ ] Uncheck "Optimize Network for Speed" — confirm CMSG_ENABLE_NAGLE is ignored

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -67,3 +67,49 @@ Three observation-layer improvements applied on top of the rebase commits above.
 3. **Benign realmd close re-tag:** every successful login produced a red `Error | AuthClient | Socket Closed By Server` after auth completion because Kronos realmd intentionally closes the socket after serving the realmlist. `ReceiveCallback` now checks `_response.Task.IsCompleted` — if true, logs `Network` with "Realmd disconnected after successful auth (expected)"; if false, keeps the `Error` + `SetAuthResponse(FAIL_INTERNAL_ERROR)` path for real auth failures.
 
 No behaviour changes to dispatch or translation; all three are observation-layer improvements.
+
+## 2026-04-29 — Latency fixes: Nagle ignore + GCD double-cast window + RTT-adaptive offset
+
+Three related latency improvements, shipped as PR #87 targeting beta.
+
+### CMSG_ENABLE_NAGLE ignore
+
+**Issue:** The 1.14 client sends `CMSG_ENABLE_NAGLE` when the user unchecks "Optimize Network for Speed," re-enabling Nagle's algorithm (~200ms write coalescing) on both the client-facing and server-facing sockets. For a game proxy that needs low-latency bidirectional forwarding, this adds pure delay.
+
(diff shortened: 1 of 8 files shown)


jameopotato (commented):
**Review: Approve with suggestions**

Implementation is clean, correct, and well-decomposed across commits. All code references verified against the codebase. Three suggestions:

## 1. Guard ordering comment for PR #86

Above the `HasForwardedPendingCast()` block in HandleCastSpell, add a comment like `// PR #86's HasInFlightNormalCastForSpell guard slots here (position 4)` to prevent accidental misordering during rebase.

## 2. Consider measuring keepalive pings too

Currently skips keepalive pings (bit 31 set) in `RecordPingSent`. Since keepalives traverse the same proxy→server path and arrive every 30s, including them would double the sample rate and halve warm-up from ~90s to ~45s. Only change needed: remove the early return in `RecordPingSent`. `RecordPongReceived` already runs before the keepalive filter in HandlePingResponse. Optional — 90s warm-up is acceptable — but it's free accuracy.

## 3. Keep manual Settings clamp at 50ms

The manual fallback clamp was raised from 50 to 100ms to match the adaptive cap. A user manually setting `SpellCastEarlyFireOffsetMs=100` without RTT data fires 100ms early — on a <100ms RTT connection, this guarantees NOT_READY failures. Consider keeping the manual clamp at 50ms (conservative for blind manual tuning) and letting only the adaptive path reach 100ms. The adaptive path has EMA smoothing as its safety net; the manual path doesn't.

PR #88, opened 2026-04-30
Fix Rupture/Kidney Shot duration on vanilla: snapshot CP at cast time

Mongrul:
- Vanilla 1.12 servers compute CP-scaled finisher durations server-side (`base + perCp × CP`) but never echo them for enemy debuffs — only the player's own auras get `SMSG_UPDATE_AURA_DURATION`. The proxy was falling back to `AuraDurations1.csv`, which only carries the raw `SpellDuration.dbc` base (flat 6000 ms for all 6 ranks of Rupture). Result: a 5 CP Rupture's debuff icon expired at 6 s on the modern client while the server kept ticking for 16 s, leaving a "lingering" mismatch.
  - Fix: snapshot the player's combo points at `CMSG_CAST_SPELL` (before the server consumes them) and use the snapshot in the aura-apply paths to synthesize the correct duration locally. Snapshot wins over the slot cache so successive recasts with varying CP no longer inherit a stale duration.
  - Covers Rupture (6 ranks) and Kidney Shot (2 ranks). Slice and Dice is intentionally not in the table — it's a self-buff and already gets the talented duration via `SMSG_UPDATE_AURA_DURATION`. Eviscerate has no aura. Expose Armor is flat 30 s on vanilla regardless of CP.

  ## Vanilla-specific gotcha worth noting
  Vanilla 1.12 has no `SMSG_UPDATE_COMBO_POINTS` opcode — combo points and combo target both arrive as regular
  unit-field updates (`PLAYER_FIELD_BYTES >> 8` and `PLAYER_FIELD_COMBO_TARGET`). The CP cache is now populated from
  `UpdateHandler.cs` for vanilla and from the existing `CharacterHandler.cs` opcode handler for TBC/WotLK.

  ## Files touched
  - `HermesProxy/World/GameData.cs` — `ComboPointFinisherDurations` table + `TryGetComboPointDuration` / `IsComboPointFinisher` helpers
  - `HermesProxy/GlobalSessionData.cs` — `CurrentComboPoints` / `CurrentComboTarget` cache + finisher-cast snapshot with 3 s TTL
  - `HermesProxy/World/Client/PacketHandlers/CharacterHandler.cs` — store CP/target on `SMSG_UPDATE_COMBO_POINTS` (TBC/WotLK)
  - `HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs` — store CP/target from vanilla `PLAYER_FIELD_BYTES` / `PLAYER_FIELD_COMBO_TARGET`; aura-discovery now prefers snapshot over cache for finishers
  - `HermesProxy/World/Server/PacketHandlers/SpellHandler.cs` — snapshot CP at `CMSG_CAST_SPELL` for known finishers
  - `HermesProxy/World/Client/PacketHandlers/SpellHandler.cs` — `SendAuraRefreshUpdate` prefers snapshot over cache for finishers

  ## Diagnostics added
  - `finisher.cmsg_cast` — fires on every finisher CMSG with the cached CP/target state (kept for future debugging)
  - `finisher.snapshot` — fires when the snapshot is stored

  ## Test plan
  - [x] `dotnet build` clean (0 warnings, 0 errors)
  - [x] `dotnet test` — 312 passed, 0 failed
  - [x] Live-test on Kronos, mixed CP rupture casts on the same target across slot reuse:
    - 1 CP → 8 s, 2 CP → 10 s, 3 CP → 12 s, 4 CP → 14 s, 5 CP → 16 s — all verified in `aura.slot.set` events
    - Recasts with decreasing CP correctly shorten the displayed duration (no more inheriting prior 16 s)
  - [ ] Spot-check Kidney Shot on a humanoid (1–5 CP → 2–6 s)
  - [ ] Confirm Slice and Dice unchanged (player-self path untouched, still uses `SMSG_UPDATE_AURA_DURATION`)
  - [ ] Confirm Expose Armor unchanged (flat 30 s from CSV)

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -162,6 +162,17 @@ public sealed class GameSessionData
     public Dictionary<WowGuid128, Dictionary<byte, int>> UnitAuraDurationLeft = [];
     public Dictionary<WowGuid128, Dictionary<byte, int>> UnitAuraDurationFull = [];
     public Dictionary<WowGuid128, Dictionary<byte, WowGuid128>> UnitAuraCaster = [];
+
+    // JimsProxy (Rupture-DoT-Lingering-Icon): combo-point cache + finisher-cast snapshot.
+    // Vanilla servers don't send aura duration for enemy debuffs, and CP-scaling finishers
+    // (Rupture, Kidney Shot) compute their duration server-side as (base + perCp × CP).
+    // We cache CP from SMSG_UPDATE_COMBO_POINTS and snapshot it on the outgoing
+    // CMSG_CAST_SPELL — at that moment the server hasn't consumed CP yet, so the cached
+    // value is the real CP that will be applied. Aura-apply paths consult the snapshot to
+    // synthesize the correct duration locally before the legacy server clears the CP.
+    public byte CurrentComboPoints;
+    public WowGuid128 CurrentComboTarget;
+    private (uint SpellId, WowGuid128 Target, byte ComboPoints, int Tick)? _pendingFinisherCast;
     public Dictionary<WowGuid128, PlayerCache> CachedPlayers = [];
     public HashSet<WowGuid128> IgnoredPlayers = [];
     public Dictionary<WowGuid128, uint> PlayerGuildIds = [];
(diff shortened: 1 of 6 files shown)


jameopotato:
## Code Review — Approve with suggestions

**Summary:** Rogue Rupture/Kidney Shot debuffs showed wrong duration on enemies (6s instead of 16s for 5 CP Rupture). The server computes `base + perCp × CP` but never echoes it for enemy debuffs. Fix: snapshot CP at CMSG_CAST_SPELL time, synthesize correct duration on aura apply.

**Correctness:**
- CP source for vanilla correctly identified — no SMSG_UPDATE_COMBO_POINTS in 1.12, uses PLAYER_FIELD_BYTES update instead. Both vanilla and TBC+ paths covered.
- Snapshot timing is correct — before server consumes CP, before GCD hold decision.
- Formulas match vanilla DBC: Rupture `6000 + 2000×CP` ms, Kidney Shot `1000 + 1000×CP` ms. All ranks covered.
- Slice and Dice correctly excluded (self-buff gets authoritative duration from server).
- 3s TTL on snapshot prevents stale cross-fight bleed. Refresh override prevents CP-count inheritance.

**Efficiency:**
- Core logic is O(1) — FrozenDictionary lookup, infrequent snapshot storage. No hot-path concern.
- **Suggestion:** `aura.slot.set`, `aura.slot.cleared`, `aura.slot.skipped`, and `aura.tick` diagnostic events fire per-aura-per-update on every unit. These are tagged "remove once root cause confirmed." They should be gated behind `DebugOutput` — will be addressed in a post-merge cleanup pass across all hot-path diagnostic events.

**Future compatibility:** Low rebase risk on GameData/GlobalSessionData. Medium risk on UpdateHandler aura loop (touched by multiple PRs). Vanilla-only spell IDs — no TBC+ impact.

**Verdict:** The snapshot-at-cast-time design is the only viable approach — the server doesn't send CP-scaled durations for enemy debuffs. Core logic is correct and well-documented.

LGTM — merging to beta. Diagnostic event gating backlogged for cleanup pass.

PR #94, opened 2026-04-30
Macro gcd fix

Mongrul:
Fixes a cluster of action-button quirks observed when using macro-batched
  `/use 13 + /use 14 + /cast <Spell>` patterns (and rapid double-clicks of
  mount/trinket items) against vanilla 1.12 emulators (Kronos):

  - Mount/spell button getting stuck **lit/queued forever** after a rapid
    double-click.
  - Sinister Strike (and other instant on-GCD spells) showing **no GCD
    swipe** for ~300ms after pressing a trinket-bracketed macro — the click
    was acked, but the swipe didn't appear until the trinket completed and
    SS's own SPELL_GO arrived.
  - Held casts occasionally **stranded for several seconds** (one bundle
    showed 7.5s) when the in-flight cast that was holding them happened to
    be off-GCD (e.g. a trinket).

  ## Changes

  **Mount-Button-Stuck-Lit fix (cherry-picked)** — adds a
  `HasInFlightNormalCastForSpell()` guard so duplicate `CMSG_CAST_SPELL` /
  `CMSG_USE_ITEM` presses are rejected with a `CastFailed` bound to the
  duplicate's own `ClientGUID/ServerGUID`, releasing the modern client's
  per-click queued state without disturbing the running cast bar.

  **`SpellInProgress` instead of `DontReport` on duplicate drops** — the
  mount fix originally used `DontReport` to suppress the "Another action is
  in progress" toast, but the 1.14 client kept the click queued/lit anyway.
  `SpellInProgress` matches what the displaced-hold path already uses and
  the client treats it as "overridden by newer cast" — releases the lit
  state cleanly. Cosmetic: rapid double-clicks may now show a brief red
  error toast.

  **Drain `_heldGcdCast` in `HandleSpellGo`** — when a forwarded cast
  completes with no GCD active and no other forwarded cast pending, fire
  the held cast immediately. Off-GCD trinket completions don't trigger
  `BeginGcd`, so without this the held cast sat until something else
  (usually an unrelated CastFailed) happened to drain it. New diagnostic:
  `spell.held_fire_on_go`.

  **Synthesize `SMSG_COOLDOWN_EVENT` at `held_pending` time** — when SS is
  held because a trinket is in flight ahead of it, the SpellPrepare ack
  fires at click time but the modern client's GCD swipe doesn't start
  until SS's real SPELL_GO arrives ~300ms later. Synthesizing
  `SMSG_COOLDOWN_EVENT` at hold time lets the client compute the GCD
  locally from the spell's DBC `StartRecoveryTime` and starts the swipe
  at click-ack. `SMSG_SPELL_COOLDOWN` with `ForcedCooldown` was tried
  first but got "anticipated-then-canceled" by the client's macro
  evaluator. Vanilla-only (TBC+ haste-adjusted GCDs would be wrong). New
  diagnostic: `gcd.cooldown.synth_held`.

  ## Diagnostics added

  | Event | Fires when |
  |---|---|
  | `cast.dropped.duplicate` | Mount-fix gate rejects an in-flight duplicate cast |
  | `spell.held_displaced` | A held cast was replaced by a newer press |
  | `spell.held_fire_on_go` | A held cast was released because the cast it was waiting on completed |
  | `gcd.cooldown.synth_held` | Synthesized `SMSG_COOLDOWN_EVENT` at hold time to start the macro swipe early |

  ## Test plan

  - [x] `dotnet build HermesProxy.sln` — clean
  - [x] `dotnet test` — 312/312 pass
  - [x] Live: rapid double-click mount item — button clears (no lit-stuck)
  - [x] Live: SS + `/use 13 + /use 14` macro spam — GCD swipe appears
        immediately on first press, no stuck buttons, holds resolve in
        ~140ms
  - [x] Verify against bundles 144056, 150846, 151514, 153444 (each
        iteration's diagnostic events confirmed expected behavior)
  - [ ] Sanity-check unrelated flows: solo SS spam, instant casts, plain
        `/use` (no spell), normal mounting — no regressions vs `beta`

### DIFF
--- a/Addons/JimsPlus/JimsPlus.toc
+++ b/Addons/JimsPlus/JimsPlus.toc
@@ -1,4 +1,4 @@
-## Interface: 11400
+## Interface: 11402
 ## Title: JimsPlus
 ## Notes: Bundled client-side fixes for JimsProxy / Classic WoW 1.14 Launcher
 ## SavedVariables: JimsPlusDB, JimsPlusCastbarsDB
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -67,3 +67,49 @@ Three observation-layer improvements applied on top of the rebase commits above.
 3. **Benign realmd close re-tag:** every successful login produced a red `Error | AuthClient | Socket Closed By Server` after auth completion because Kronos realmd intentionally closes the socket after serving the realmlist. `ReceiveCallback` now checks `_response.Task.IsCompleted` — if true, logs `Network` with "Realmd disconnected after successful auth (expected)"; if false, keeps the `Error` + `SetAuthResponse(FAIL_INTERNAL_ERROR)` path for real auth failures.
 
 No behaviour changes to dispatch or translation; all three are observation-layer improvements.
+
+## 2026-04-29 — Latency fixes: Nagle ignore + GCD double-cast window + RTT-adaptive offset
+
+Three related latency improvements, shipped as PR #87 targeting beta.
+
+### CMSG_ENABLE_NAGLE ignore
+
+**Issue:** The 1.14 client sends `CMSG_ENABLE_NAGLE` when the user unchecks "Optimize Network for Speed," re-enabling Nagle's algorithm (~200ms write coalescing) on both the client-facing and server-facing sockets. For a game proxy that needs low-latency bidirectional forwarding, this adds pure delay.
+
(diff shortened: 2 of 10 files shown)


Mongrul:
Superceeded by https://github.com/jameopotato/jimsproxy/pull/113

PR #96, opened 2026-05-01
Tallstrider fix

Mongrul:
Fixes tallstriders (and other vanilla ground creatures) visibly running sideways for a few hundred ms when aggroing,
  while preserving smooth body rotation for patrolling NPCs at waypoint corners.

  Root cause

  Vanilla creatures running on a default-Runmode SplineFlagVanilla spline were unconditionally given
  SplineFlagModern.Steering | Unknown10 by the proxy whenever UnitFlagsVanilla.InCombat wasn't set. The legacy server
  pushes the InCombat flag a packet or two after the MONSTER_MOVE that initiates the chase, so the very first aggro
  spline always slipped past the guard.

  Steering tells the modern client to smoothly interpolate body rotation along the path — fine for patrol corners, wrong
   for aggro turns where vanilla snaps the creature to face the player.

  Fix

  Replaces the unit-flag check with a two-gate angle-and-distance heuristic that runs after path parsing:

  - LastKnownOrientation cache (GlobalSessionData) — populated from every Living-unit ObjectUpdate movement block and
  from MSG_MOVE_HEARTBEAT-style packets.
  - Gate 1 (angle): apply Steering only when the first spline segment direction is within 90° of the creature's cached
  facing. Patrol corners stay under this; large state-transition turns (creature was facing N, now charges S) blow past
  it.
  - Gate 2 (distance): skip Steering when the total spline distance exceeds 20 yd. Catches aggro paths whose first
  segment happens to align with the creature's current heading (would otherwise pass Gate 1) — those are typically 15–40
   yd chases to the player, while patrol segments are 5–10 yd between waypoints.

  Fallback when no orientation is cached: skip Steering. Snap is the vanilla default and the safer mis-classification.

  Tradeoff

  - Tallstrider aggro now snaps to face the player (vanilla-faithful — the bug fix).
  - Patrolling NPCs at corners still get smooth body rotation through their waypoint paths (modern-look, preserves the
  pre-fix behavior that direct 1.12 clients showed for those motions).

  Taxi-flight catmull-rom paths in MovementHandler/UpdateHandler are untouched — those legitimately need Steering for
  bezier interpolation and aren't gated on creature state.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -209,6 +209,15 @@ public sealed class GameSessionData
     // packets for it get the hover override regardless of HOVERHEIGHT.
     public HashSet<WowGuid128> KnownHoveringMobs = [];
 
+    // JimsProxy (Tallstrider-Fix): per-GUID last-known facing orientation, populated from
+    // any MovementInfo we observe (spawn, heartbeat, ObjectUpdate movement block). Used by
+    // MovementHandler.HandleMonsterMove to compare the creature's current facing against
+    // the spline's first-segment direction — if the angle change is large, we treat the
+    // move as a state-transition (aggro/turn-to-target) and skip SplineFlagModern.Steering
+    // so the modern client snaps to the new heading instead of slowly rotating the body
+    // through the path. Small angle changes get Steering for smooth patrol corners.
+    public Dictionary<WowGuid128, float> LastKnownOrientation = new();
+
     private GameSessionData()
     {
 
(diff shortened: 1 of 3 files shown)


jameopotato:
## Code Review — Approve

**Summary:** Vanilla ground creatures (tallstriders, wolves) ran sideways for ~300ms on aggro because Steering was applied unconditionally to non-combat splines. The InCombat flag arrives a packet or two after the aggro MONSTER_MOVE, so the first chase spline always slipped past the guard.

**Correctness:** Two-gate heuristic correctly separates patrol corners (Steering = smooth rotation) from aggro chases (snap to heading):
- Gate 1 (angle < 90°): patrol turns stay under, aggro turns blow past
- Gate 2 (distance < 20yd): patrol segments are 5-10yd, aggro chases 15-40yd
- Fallback with no cached orientation: skip Steering (snap = vanilla default)
- Edge cases verified: flee (large angle → snap ✓), leash/evade (small angle + short distance → smooth ✓), mounted players (unaffected ✓)

**Efficiency:** O(1) dictionary lookup + MathF.Atan2/Sqrt per MONSTER_MOVE — hardware-accelerated, ~10ns. Orientation cache write per movement update is negligible. Minor note: `LastKnownOrientation` dictionary grows unbounded as units enter view (no cleanup on despawn). Not a real problem for typical sessions.

**Future compatibility:** Isolated to vanilla branch (gated on `RemovedInVersion V2_0_1_6180`). Deferred `steeringCandidate` evaluation is a clean refactor. Medium rebase risk on MovementHandler but insertion point is stable.

LGTM — merging to beta.

PR #97, opened 2026-05-01
Inspect random suffix

Mongrul:
HandleInspectResult (vanilla/TBC branch) read each visible item's ID
   and perm enchant from PLAYER_VISIBLE_ITEM_X_0 / +1 but never read the
   adjacent PLAYER_VISIBLE_ITEM_X_PROPERTIES dword that carries the random
   property/suffix ID. The modern client therefore received every
   inspected item with RandomPropertiesID = 0 and had no way to compute
   the suffix string, so green "of the Bear / Owl / ..." gear inspected
   without its suffix name.

   Within-slot offset to PROPERTIES is slot_stride - 4 in both eras
   (verified against V1_12_1_5875 UpdateFields: 0x50 - 0x48 = 8 with
   stride 12, and V2_4_3_8606: 0x7C - 0x70 = 12 with stride 16). The
   field stores a signed int32 in a uint32 dword (negative ->
   ItemRandomSuffix.dbc, positive -> ItemRandomProperties.dbc); copying
   the raw uint32 into ItemInstance.RandomPropertiesID preserves the
   sign-extension and matches the convention already used by the auction,
   loot, mail, and trade handlers.

   Seed is intentionally left at 0 -- the visible-item field block
   doesn't carry ITEM_FIELD_RANDOM_PROPERTIES_SEED, and the suffix name
   depends only on the property ID; the seed only affects rolled stat
   values, which inspect doesn't display anyway.

   Adds inspect.item Log.Event per slot so future user bug bundles
   capture target/slot/item/property/enchant.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/CharacterHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/CharacterHandler.cs
@@ -1,4 +1,5 @@
-﻿using HermesProxy.Enums;
+﻿using Framework.Logging;
+using HermesProxy.Enums;
 using HermesProxy.World.Enums;
 using HermesProxy.World.Objects;
 using HermesProxy.World.Server.Packets;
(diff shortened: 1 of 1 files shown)


jameopotato:
## Code Review — Approve

**Summary:** Green "of the Bear / Owl / ..." gear inspected on other players showed without its suffix. The inspect handler read item ID and enchant but never read `PLAYER_VISIBLE_ITEM_X_PROPERTIES` — the dword carrying the random property/suffix ID.

**Correctness:** The offset calculation reads the PROPERTIES field at the correct position within each visible-item block. Signed/unsigned handling is correct — negative = ItemRandomSuffix.dbc, positive = ItemRandomProperties.dbc. Verified locally on Kronos: suffix names now display correctly on inspected green gear.

**Efficiency:** One extra `ContainsKey` per equipment slot per inspect. Dictionary lookup, O(1). Inspects are rare — zero hot-path impact.

**Future compatibility:** Isolated to `HandleInspectResult` in CharacterHandler.cs. Low rebase risk. Addresses SugarProxy competitive claim #2 ("correctly view target enchantments").

**Diagnostic event:** `inspect.item` fires per slot per inspect (up to 19). Acceptable for a rare operation; will be gated behind DebugOutput in a future cleanup pass.

Tested locally — LGTM, ready to merge.

PR #101, opened 2026-05-01
 Add diagnostics for MC Quintessence-on-Rune cast bug (preempt #374)

Mongrul:
Plant scoped Log.Event diagnostics on the item-on-GameObject cast path so
  when Molten Core opens we can root-cause the Aqual/Eternal Quintessence
  "out of range" bug from one bundle, rather than burning raid attempts on
  trial-and-error. Bug originally reported in WowLegacyCore/HermesProxy#374
  on a CMaNGOS TBC server; needs to be verified on Kronos when MC unlocks.

  - spell.use_item.target_dump (gated on Target.Flags & GameObject) — logs spell ID, item ID, target GUID 128/64, target entry, target flags, and src/dst locations for any item-on-GameObject cast. Absence of this event during a failed Quintessence attempt is itself a signal (= client-side pre-validation refused to send).
  - spell.cast_failed.diag — logs raw + named SpellCastResult reason for any cast failure, paired with spell ID. Already low-volume.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -158,6 +158,18 @@ public partial class WorldClient
         if (packet.CanRead())
             arg2 = packet.ReadInt32();
 
+        // JimsProxy: MC-Quintessence-Investigation. Pair with spell.use_item.target_dump in the
+        // bundle to see exactly why a server-side cast was rejected (range/LOS/wrong-target/etc).
+        // Reason is the raw 1.12 SpellCastResult enum value; resolve via SpellCastResultClassic.
+        Log.Event("spell.cast_failed.diag", new
+        {
+            spell_id = spellId,
+            reason_raw = reason,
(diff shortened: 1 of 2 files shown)


jameopotato:
## Triage Note

This PR added diagnostics for the MC Quintessence bug (HermesProxy#374). **PR #121 now contains the actual fix** — GameObjectFlags vanilla→modern translation with entry-gated override for the MC runes, plus tests.

#121's `gameobject.flags.translated` Log.Event provides equivalent diagnostic coverage to the `spell.use_item.target_dump` event here (and broader — it fires on all GO flag translations, not just Quintessence casts).

The `spell.cast_failed.diag` event from this PR is still independently useful (raw SpellCastResult logging for any failed cast), but it's also included in #124's larger casting changes.

**Recommend: close without merging** — superseded by #121 for the Quintessence investigation, and by #124 for the cast_failed diagnostic.

jameopotato:
Closing — superseded by #121 which contains the actual MC Quintessence fix (GameObjectFlags translation). The spell.cast_failed.diag diagnostic is also covered by #124.

PR #113, opened 2026-05-02
Miras mega cast fix

Mongrul:
Jim - this combines #94 (Macro-GCD-Fix) and #103 (Pickpocket-Macros) and
  adds instant-cast feel tuning on top, all empirically validated.

  ## What's in here

  ### 1. Both prior PRs merged
  - **#94 (Macro-GCD-Fix)**: macro-tooltip GCD swipe + stuck-button fixes for
    `/use 13 + /cast` patterns
  - **#103 (Pickpocket-Macros)**: release the held cast on `SMSG_SPELL_GO` success
    for off-GCD instants — fixes Pickpocket → Cheap Shot needing two presses

  ### 2. Adaptive GCD fire-offset tuning
  A player reported instants felt sluggish — "the GCD sweep finishes a slight tick
  before you can fire another off." Traced it: with `RTT*0.5` the held cast
  arrived at the legacy server right as its GCD ended, but Twinstar adds ~30ms of
  its own processing jitter, so the modern client's swipe ends ~50–60ms before
  the next `SMSG_SPELL_GO` arrives.

  Switched to `RTT − 10ms` (cap 100ms). The 10ms is a constant safety margin at
  every RTT — important because the previous percentage-based formula gave only
  `0.05 × RTT` margin (~2ms at 35ms RTT), which would have flooded NOT_READY
  rejections for low-latency euro players. Cycle dropped from ~1663ms → ~1550ms
  with zero NOT_READY events across multiple sessions.

  ### 3. RTT robustness
  The original `> 5000ms` outlier reject was way too lenient. A realm swap
  inserted 672ms and 641ms startup pings (TCP+auth+world-entry overhead, not
  real RTT) which pinned the EMA at 588ms for the entire session and saturated
  the fire offset at the cap, causing stuck buttons. Lowered to `> 300ms` and
  added a `ResetRttSmoothing()` call on every `WorldClient.ConnectToWorldServer`
  so realm swaps warm up cleanly.

  ### 4. Removed the synthesized SMSG_COOLDOWN_EVENT from #94
  That block sent the cooldown event at `SpellPrepare` ack time (mid-GCD), which
  locked the action bar early and made instants feel _more_ sluggish. The
  adaptive fire-offset addresses the gap properly without locking the bar.

  ## What I caught while preparing this PR

  The auto-merge of #94 silently dropped the **vanilla CP-finisher snapshot
  block** in `Server/SpellHandler.cs` — the `IsComboPointFinisher` →
  `StorePendingFinisherCast` snapshot that lets the proxy compute correct
  Rupture / Kidney Shot durations locally (the legacy server applies
  CP-scaled durations server-side and never echoes them for enemy debuffs).
  Without that block, vanilla rogue finisher debuff icons would regress to the
  default unscaled duration on the modern client. Restored as part of the
  tuning commit.

  Also removed a duplicate `HasInFlightNormalCastForSpell` method that
  auto-merge produced when #94 and #103 each added their own copy.

  ## Validation

  - Build clean
  - Bug-hunter agent review caught the CP-finisher regression before push
  - Live play across multiple AE-spam sessions: zero NOT_READY, no stuck
    buttons, no DCs, instant-cast cycle averaging 1548ms
  - Predictive `SMSG_SPELL_GO` synthesis was attempted to push past this floor
    and DC'd via `worldclient_legacy_disconnect / header` (modern client emits
    follow-up CMSGs keyed on synthesized state). Reverted; floor is RTT+jitter-bound.

  Closes #94, closes #103.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -751,6 +751,28 @@ public sealed class GameSessionData
     }
 
     /// <summary>
+    /// JimsProxy (Mount-Button-Stuck-Lit): returns true if any pending normal cast — started OR
+    /// merely in flight to the legacy server — matches the given SpellId (or its LegacySpellId
+    /// for SoM-renumbered USE_ITEMs). HasStartedNormalCast only catches duplicates AFTER
+    /// SMSG_SPELL_START matches the queue entry; rapid mashing within the c→s→c round-trip slips
+    /// past it and the duplicate USE_ITEM/CAST_SPELL gets forwarded. The legacy server then rejects
+    /// the duplicate with SpellInProgress, and the FIFO dequeue (TryDequeuePendingNormalCast picks
+    /// the oldest match) binds the failure to the in-flight cast's IDs — the bug commit 7b9e8aa
+    /// originally fixed and that resurfaced when 8998c39 switched HandleUseItem to silent-drop
+    /// gated on HasStartedNormalCast only. Used by HandleCastSpell and HandleUseItem to silent-drop
+    /// the in-flight same-spell duplicate before it ever reaches the queue.
+    /// </summary>
+    public bool HasInFlightNormalCastForSpell(uint spellId)
+    {
+        foreach (var item in PendingNormalCasts)
+        {
+            if (CastMatchesSpellId(item, spellId))
+                return true;
+        }
+        return false;
+    }
+
+    /// <summary>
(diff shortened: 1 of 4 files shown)


jameopotato:
Closing — subsumed by #124 which includes the adaptive RTT fire offset, Pickpocket held-cast release, RTT outlier rejection, ResetRttSmoothing, and CP-finisher snapshot restore from this PR.

PR #114, opened 2026-05-03
Warrior charge revisit

Mongrul:
Twinstar/Kronos broadcast a second SMSG_SPELL_GO ~1ms after the parent
  ▎ Charge/Intercept GO for the triggered stun (7922 / 20615) with caster=
  ▎ local player. The modern 1.14 client kicked the stun kit's caster events
  ▎ on the warrior on top of the lunge kit, producing a visible animation
  ▎ twitch ("warrior on crack"). Rewrite CasterGUID/CasterUnit to the first
  ▎ hit target so the stun kit plays on the mob instead. Stun aura/icon and
  ▎ damage attribution are unaffected.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -751,6 +751,28 @@ public sealed class GameSessionData
     }
 
     /// <summary>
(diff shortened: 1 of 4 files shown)


Mongrul:
Superseded by #117

PR #116, opened 2026-05-03
Pet scale vanilla parity

Mongrul:
- Modern 1.14 Classic renders the local player's pet (warlock + hunter) visibly smaller than the vanilla 1.12
  reference, even though `CreatureDisplayInfo.CreatureModelScale` and `CreatureModelData.ModelScale` are byte-identical
  between the two builds (verified by extracting both sets from a 1.12.1 client and diffing against 1.14.2.42597 — only
  39 of ~10,500 DisplayIDs differ in CMS, none of them pets).
  - Root cause unverified — likely an undocumented client-side scale modifier on local-player pet units in modern
  Classic. Sugar Proxy ships an equivalent compensation knob.
  - A flat global multiplier doesn't work because the modern client compounds it with the per-display CMS: a Felhunter
  (CMS 0.5) at 1.5× still renders too small, while a Dire Wolf (CMS 1.5) at 1.5× ends up at ~2.2 effective (verified:
  tester screenshot showed wolf-on-steroids).

  ## Fix — inverse-CMS scaling
  For the local pet, emit `(wire_OBJECT_FIELD_SCALE_X / CreatureModelScale) × K` instead of the raw wire scale. The
  modern client's `wire × ModelScale × CMS` multiply collapses to `wire × ModelScale × K`, so every pet gets a uniform K
   bump regardless of its CMS quirks.

  `K = 1.5` is the empirical "matches 1.12" tuning value validated against same-pet side-by-side screenshots.

  | Pet | CMS | wire | emit | renders | matches 1.12 |
  |---|---|---|---|---|---|
  | Dire Wolf 9370 | 1.5 | 0.975 | 0.975 | 1.46 | yes |
  | Felhunter 850 | 0.5 | 0.7 | 2.1 | 1.05 | yes |
  | Voidwalker 1132 | 0.8 | wire | wire×1.875 | wire×1.5 | yes |
  | Imp 4449 | 0.5 | 0.5 | 1.5 | 0.75 | yes |

  ## Detection
  - Prefer `CurrentPetGuid` match (set when `SMSG_PET_SPELLS_MESSAGE` arrives, typically before the pet's first
  CREATE_OBJECT).
  - Fallback to `SUMMONEDBY == CurrentPlayerGuid` so the very first CREATE_OBJECT for a pet (which carries SUMMONEDBY)
  gets normalized even if `CurrentPetGuid` hasn't latched yet.
  - Only fires when `OBJECT_FIELD_SCALE_X` is present in this update — delta updates leave the already-normalized scale
  sticky on the client and avoid compounding multiplications.
  - Bounds-safe (`UNIT_FIELD_DISPLAYID_idx < updateMaskArray.Length`) so the path doesn't trip on Item/GameObject
  updates whose mask doesn't cover unit fields.

  ## Scope
  - Affects only the visual scale forwarded to the modern client. Server-side `UNIT_FIELD_BOUNDINGRADIUS` and
  `UNIT_FIELD_COMBATREACH` are untouched, so range checks, melee hit detection, and ability targeting are unchanged.
  - Click hitbox grows with the visual scale (the modern client derives the click area from the rendered model).
  - Vanilla-source-only (gated on `LegacyVersion.ExpansionVersion == 1`); no behavior change for TBC+ legacy servers.

  ## Diagnostics
  New `unit.pet_scale.applied` JSONL event per pet SCALE_X update with `display_id`, `creature_model_scale`, `k`,
  `raw_scale`, `emitted_scale`, `matched_via`, and `cms_fallback`. Lets a tester report on a single pet be triaged from
  the bundle without needing local repro.

  ## Test plan
  - [x] Hunter pet (Dire Wolf 9370): same-pet side-by-side vs 1.12 client at K=1.5 — matches.
  - [ ] Warlock pet (Felhunter): tester verification pending — math says it should be roughly correct at K=1.5; if
  reported still small, bump K or split per-family.
  - [x] Build clean (zero warnings, zero errors).
  - [x] No crash on Item/GameObject updates (bounds check verified after earlier `BitArray` index 131 regression).

### DIFF
--- a/HermesProxy/Configuration/ConfigurationParser.cs
+++ b/HermesProxy/Configuration/ConfigurationParser.cs
@@ -93,6 +93,20 @@ public class ConfigurationParser
         return defValue;
     }
 
+    public float GetFloat(string key, float defValue)
+    {
+        KeyValueConfigurationElement s = _settingsCollection[key];
+        if (string.IsNullOrEmpty(s?.Value))
+            return defValue;
+
+        float aux;
+        if (float.TryParse(s.Value, NumberStyles.Float, CultureInfo.InvariantCulture, out aux))
+            return aux;
+
+        Console.WriteLine("Warning: \"{0}\" is not a valid float value for key \"{1}\"", s.Value, key);
+        return defValue;
+    }
+
     public TEnum GetEnum<TEnum>(string key, TEnum defValue) where TEnum : struct
     {
         KeyValueConfigurationElement s = _settingsCollection[key];
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -751,6 +751,28 @@ public sealed class GameSessionData
     }
 
     /// <summary>
+    /// JimsProxy (Mount-Button-Stuck-Lit): returns true if any pending normal cast — started OR
+    /// merely in flight to the legacy server — matches the given SpellId (or its LegacySpellId
+    /// for SoM-renumbered USE_ITEMs). HasStartedNormalCast only catches duplicates AFTER
+    /// SMSG_SPELL_START matches the queue entry; rapid mashing within the c→s→c round-trip slips
+    /// past it and the duplicate USE_ITEM/CAST_SPELL gets forwarded. The legacy server then rejects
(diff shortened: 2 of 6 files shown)


Mongrul:
Superseded by #117

PR #119, opened 2026-05-03
Unplanned-DC auto-reconnect: try cached-key resume, fall back to clea…

Mongrul:
…n DC

 Fixes the 37-second-frozen-world UX when the legacy server forcibly closes
  the proxy's TCP socket mid-session. Bundle 20260503-195159 from a tester on
  Kronos PTR captured the failure mode: SocketException (TCP RST) at
  19:58:55, last opcode SMSG_ON_MONSTER_MOVE 219ms earlier, then the proxy's
  prior suppress-only handling left the modern client connected to a ghost
  world for 37 seconds — no SMSG updates, no error feedback — until the
  player gave up and logged out via CMSG_LOG_DISCONNECT.

  Behavior change:

  When a legacy-side socket exception or unexpected close is detected AND
  this WorldClient is still the active session.WorldClient (i.e. it's NOT a
  realm swap, where a replacement WorldClient has already been wired up),
  attempt one cached-session-key reconnect:
    1. New WorldClient + ConnectToWorldServer (re-uses the realmd session key the original WorldClient captured at connect time).
    2. On auth success, re-issue CMSG_PLAYER_LOGIN with the cached GameState.CurrentPlayerGuid so the legacy server places the character back in the world.
    3. Bounded by Settings.UnplannedReconnectTimeoutMs (default 5000ms, clamped 1000–30000) so a dead listener doesn't stall the modern client past the OS-level TCP timeout.

  If reconnect fails, times out, or is disabled, close session.InstanceSocket
  so the modern client sees "Disconnected" within ~1 second instead of being
  stuck. Matches "proxy mimics direct 1.12 client" parity for the
  non-recoverable case.

  Realm-swap path is unchanged — the new WorldClient already in flight
  during a swap was never the active client when the old one tore down,
  so the unplanned-reconnect branch is skipped (no behavior change).

  Config:
    EnableUnplannedReconnect       (default true)  — master toggle
    UnplannedReconnectTimeoutMs    (default 5000)  — reconnect budget

  Both default in Settings.cs, so the shipped HermesProxy.config template
  doesn't need new keys to work.

  Heavy structured Log.Event coverage along the reconnect path so a JSONL
  bundle from any tester encountering this failure mode shows exactly where
  it succeeded or failed without needing a repro:
    - session.unplanned_dc.detected
    - session.unplanned_reconnect.skipped_disabled / skipped_state
    - session.unplanned_reconnect.start
    - session.unplanned_reconnect.connect_completed
    - session.unplanned_reconnect.timeout
    - session.unplanned_reconnect.player_login_sent
    - session.unplanned_reconnect.success / exception
    - session.unplanned_dc.propagated
    - session.unplanned_dc.close_error Each event carries an attempt_id so multiple attempts within a session correlate cleanly.

  Known caveats (worth follow-up bundles to validate):
    - Kronos and similar private-vanilla servers may invalidate the realmd session key server-side on disconnect; in that case the auth re-handshake will fail and we'll propagate the DC. The diagnostic events will show this clearly.
    - On reconnect success the modern client will receive a fresh SMSG_LOGIN_VERIFY_WORLD + spawn burst from the legacy server. This may visibly flash a loading screen. Acceptable vs the 37-second freeze; can be mitigated in a follow-up by suppressing select packets on the modern path during reconnect.

### DIFF
--- a/HermesProxy/Configuration/Settings.cs
+++ b/HermesProxy/Configuration/Settings.cs
@@ -35,6 +35,16 @@ public static class Settings
     // releases a held cast. 0 = fire exactly at expiry (cast lands ~RTT late at server).
     // Clamped to 0..50 in LoadAndVerifyFrom.
     public static int SpellCastEarlyFireOffsetMs;
+    // JimsProxy (unplanned-dc-auto-reconnect): when the legacy world server forcibly
+    // closes the proxy's TCP socket mid-session (anticheat, server crash, transient
+    // network reset), attempt one cached-session-key reconnect before giving up. If
+    // false, fall straight through to clean DC propagation (close modern InstanceSocket
+    // so the user sees "Disconnected" within a second instead of being stuck in a
+    // ghost world for tens of seconds).
+    public static bool EnableUnplannedReconnect;
+    // Hard timeout on the reconnect attempt — beyond this, abandon and propagate DC.
+    // Clamped to 1000..30000 in LoadAndVerifyFrom.
+    public static int UnplannedReconnectTimeoutMs;
 
     public static bool LoadAndVerifyFrom(ConfigurationParser config)
     {
@@ -61,6 +71,8 @@ public static class Settings
         StructuredLog = config.GetBoolean("StructuredLog", true);
         VerboseLog = config.GetBoolean("VerboseLog", false);
         SpellCastEarlyFireOffsetMs = Math.Clamp(config.GetInt("SpellCastEarlyFireOffsetMs", 0), 0, 50);
+        EnableUnplannedReconnect = config.GetBoolean("EnableUnplannedReconnect", true);
+        UnplannedReconnectTimeoutMs = Math.Clamp(config.GetInt("UnplannedReconnectTimeoutMs", 5000), 1000, 30000);
         Log.StructuredLogEnabled = StructuredLog;
         Log.VerboseLogEnabled = VerboseLog;
         // Open the JSONL file now so session.start's payload can include the full path.
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -1825,124 +1825,198 @@ public class GlobalSessionData
     // to investigate). Bundle the reconnect attempt as a self-contained sequence
     // tagged with a unique attempt_id so multiple attempts in a session can be
     // correlated.
-    public void TryUnplannedReconnectAndPropagate(World.Client.WorldClient deadClient)
+    // Per-session guard: ensures only one reconnect attempt runs at a time even if
+    // HandleDisconnect and the ReceiveLoop catch fire simultaneously from the same TCP RST.
+    // 0 = idle, 1 = attempt in flight. Compare-and-swapped at entry; reset in finally.
+    private int _reconnectInProgress;
+
+    public void TryUnplannedReconnectAndPropagate(
+        World.Client.WorldClient deadClient,
+        string? originalExceptionType = null,
+        string? originalExceptionMessage = null,
+        int? originalSocketErrorCode = null)
     {
         var attemptId = Guid.NewGuid().ToString("N").Substring(0, 8);
-        var realm = RealmManager.GetRealm(RealmId);
-        var playerGuid = GameState?.CurrentPlayerGuid ?? default;
 
-        Framework.Logging.Log.Event("session.unplanned_dc.detected", new
+        // Race guard FIRST — HandleDisconnect and ReceiveLoop's catch can both fire from
+        // the same TCP RST on different threads. Only one wins the CAS; the loser exits
+        // immediately without re-emitting `detected`, propagating, or queueing a Task.
+        if (Interlocked.CompareExchange(ref _reconnectInProgress, 1, 0) != 0)
         {
-            attempt_id = attemptId,
-            realm_name = realm?.Name,
-            player_guid = playerGuid.ToString(),
-            has_authclient = AuthClient != null,
-            has_instance_socket = InstanceSocket != null,
-            reconnect_enabled = Framework.Settings.EnableUnplannedReconnect,
-            reconnect_timeout_ms = Framework.Settings.UnplannedReconnectTimeoutMs,
-        });
-
-        if (!Framework.Settings.EnableUnplannedReconnect)
-        {
-            Framework.Logging.Log.Event("session.unplanned_reconnect.skipped_disabled", new { attempt_id = attemptId });
-            PropagateUnplannedDcToModern(attemptId, "reconnect_disabled");
+            Framework.Logging.Log.Event("session.unplanned_reconnect.skipped_in_progress", new { attempt_id = attemptId });
             return;
         }
 
-        if (realm == null || playerGuid == default || AuthClient == null)
+        bool taskQueued = false;
+        try
         {
-            Framework.Logging.Log.Event("session.unplanned_reconnect.skipped_state", new
+            var realm = RealmManager.GetRealm(RealmId);
+            var playerGuid = GameState?.CurrentPlayerGuid ?? default;
+
+            Framework.Logging.Log.Event("session.unplanned_dc.detected", new
             {
                 attempt_id = attemptId,
-                has_realm = realm != null,
-                has_player_guid = playerGuid != default,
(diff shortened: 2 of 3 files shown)


jameopotato:
## Deep Review: PR #119 — Unplanned-DC auto-reconnect

### 1. Summary

When the legacy server forcibly closes the proxy's TCP socket mid-session (anticheat, server crash, transient network reset), the modern client was stuck in a ghost world for up to 37 seconds — no SMSG updates, no error feedback — until the user manually logged out. This PR detects the unplanned DC, attempts one cached-session-key reconnect (bounded by a configurable timeout, default 5s), and falls back to clean DC propagation (close InstanceSocket → user sees "Disconnected" within ~1 second). Heavy JSONL instrumentation at every step.

### 2. Correctness

**The reconnect fundamentals work**: detect unplanned DC → create new WorldClient → ConnectToWorldServer → auth handshake with cached session key → re-issue CMSG_PLAYER_LOGIN → legacy server places character back in world.

**However, there is a critical missing step:**

After successful reconnect and CMSG_PLAYER_LOGIN, the new WorldClient is never registered with the session:
```csharp
// Success path (GlobalSessionData.cs, around line 1930):
newClient.SendPacketToServer(loginPacket);
Log.Event("session.unplanned_reconnect.success", ...);
// ← session.WorldClient = newClient; IS MISSING
```

The normal connection flow has the caller (WorldSocket) explicitly assign `session.WorldClient = newClient` after ConnectToWorldServer returns. In the reconnect path, this never happens. Consequences:

1. **Modern→legacy packet forwarding breaks**: WorldSocket handlers route CMSGs via `GetSession().WorldClient`. With WorldClient=null, the modern client can send packets (move, cast, chat) but they'll never reach the legacy server. The player is in a read-only ghost world — they can see their character respawn but can't control it.

2. **Second DC isn't recoverable**: HandleDisconnect's `wasActiveWorldClient` check will be false for the new WorldClient (since `session.WorldClient != this`), so any subsequent unplanned DC won't trigger a reconnect attempt — it'll be silently suppressed.

**Double-reconnect race condition:**

The `wasActiveWorldClient` check + null assignment in HandleDisconnect isn't atomic:
```csharp
bool wasActiveWorldClient = session != null && ReferenceEquals(session.WorldClient, this);
if (wasActiveWorldClient)
    session!.WorldClient = null;
// ...
if (wasActiveWorldClient)
    session!.TryUnplannedReconnectAndPropagate(this);
```

If HandleDisconnect and ReceiveLoop's catch fire simultaneously (same TCP RST), both threads could pass the ReferenceEquals check before either nulls WorldClient, leading to two parallel reconnect attempts racing to create new WorldClients.

**Failed reconnect cleanup**: If ConnectToWorldServer fails or times out, `newClient` is abandoned without explicit socket close. The socket will eventually be GC'd, but it may linger in CLOSE_WAIT.

**Other observations:**
- The realm-swap guard is correct — `wasActiveWorldClient` is false during realm swaps (new WorldClient already in flight), so reconnect is correctly skipped.
- The 5000ms default timeout is reasonable — enough for a TCP connect + auth handshake at 170ms RTT, short enough to not leave the user hanging.
- Settings clamping (1000-30000ms) and config defaults are clean.
- JSONL instrumentation is thorough — every decision point emits a diagnostic event with an attempt_id for correlation.

### 3. Efficiency

The reconnect runs on a Task.Run off the receive-loop thread — good, doesn't block exception handling. The nested Task.Run for ConnectToWorldServer with `.Wait(timeout)` is appropriate for bounding a potentially-blocking connect.

No hot-path overhead — the reconnect logic only fires on disconnect, not per-packet.

Minor allocation concern: `Guid.NewGuid().ToString("N").Substring(0, 8)` on every disconnect for the attempt_id. Negligible.

### 4. Future compatibility

- **Upstream rebase**: Modifies WorldClient.cs (disconnect handlers) and GlobalSessionData.cs. The disconnect handler changes are additive (new boolean + new call at the end). Moderate conflict risk with #124 which also touches WorldClient.cs.
- **Conventions**: Follows the existing JSONL event pattern and realm-swap protection pattern.
- **Kronos coupling**: The PR description correctly notes Kronos may invalidate session keys on disconnect. The diagnostic events will capture this clearly (auth_failed reason). This is a feature that degrades gracefully.

### 5. Three-way comparison

| Criteria | PR approach (cached-key reconnect) | Alt A: Immediate clean DC + user-facing "Reconnecting..." | Alt B: Full re-auth via AuthClient |
|----------|-----------------------------------|-----------------------------------------------------------|-----------------------------------|
| Complexity | Medium — new WorldClient + bounded reconnect | Low — just close InstanceSocket immediately | High — full SRP handshake, realm list, character select |
| Runtime efficiency | One TCP connect + auth (~500ms best case) | Instant | Full login flow (~3-5 seconds) |
| Maintainability | Self-contained in GlobalSessionData | Trivial | Touches auth, realm, world layers |
| Rebase safety | Moderate — touches disconnect handlers | Good — minimal changes | Poor — touches many systems |
| Edge-case coverage | Covers transient RSTs; degrades to clean DC on key rejection | No recovery — user must re-launch | Full recovery but complex and slow |

**Recommendation**: PR approach is the right one. The fallback-to-clean-DC design means the worst case is what we had before (except now it's ~1 second instead of 37). Alt A is the safe subset (and is the fallback path). Alt B is over-engineered for a first pass.

### 6. Verdict

**Request changes** — the concept is sound and the fallback behavior is a huge improvement over status quo, but two issues need fixing before merge:

1. **Critical: Add `this.WorldClient = newClient`** (or equivalent assignment) in the reconnect success path. Without this, the modern→legacy direction is broken post-reconnect. The player can see the world respawn but can't interact with it.

2. **Important: Guard against double-reconnect** — either use `Interlocked.CompareExchange` on the WorldClient slot, or add a per-session reconnect-in-progress flag with `Interlocked.Exchange`, so only one reconnect attempt can run at a time.

3. **Minor: Explicitly close `newClient`'s socket** on failure/timeout paths to avoid socket lingering in CLOSE_WAIT.

The JSONL instrumentation and fallback-to-clean-DC are excellent. With the WorldClient assignment fix, this is ready for v5.0.9.

jameopotato (changes requested):
Request changes — concept is sound, fallback-to-clean-DC alone is a huge win (1s vs 37s frozen world). Three issues before merge:

### Critical
**Add `this.WorldClient = newClient`** in the reconnect success path (after CMSG_PLAYER_LOGIN). Without this, modern→legacy packet forwarding is broken — the player sees the world respawn but can't control their character (movement, casting, chat all fail silently because WorldSocket routes CMSGs through `session.WorldClient`, which is still null).

### Important
1. **Guard against double-reconnect**: HandleDisconnect and ReceiveLoop's catch can fire simultaneously from the same TCP RST. The `wasActiveWorldClient` check + null isn't atomic — both threads could pass ReferenceEquals before either nulls. Use `Interlocked.CompareExchange` on the WorldClient slot or a per-session `_reconnectInProgress` flag.

2. **Log the TCP disconnect reason, don't fully suppress it**: The current `session.ondisconnect.suppressed` event captures `last_opcode` and `ms_since_last_opcode`, which is good. But we should also capture the **exception type and message** (SocketException error code, etc.) from the original disconnect — not just in the reconnect events, but in the initial detection event. If we start seeing a pattern of TCP RSTs at specific intervals or after specific opcodes, it could indicate we're triggering Warden mid-session (our tolerance patch only covers the handshake, not mid-session checks). We need visibility into *why* the server is kicking us, not just *that* it kicked us. Consider adding `exception_type`, `exception_message`, and `socket_error_code` to the `session.unplanned_dc.detected` event payload.

### Minor
Explicitly close `newClient`'s socket on failure/timeout paths to avoid lingering CLOSE_WAIT.

jameopotato (approved):
All requested changes addressed:

1. ✅ `WorldClient = newClient` before CMSG_PLAYER_LOGIN — modern→legacy routing restored
2. ✅ Double-reconnect guard — Interlocked.CompareExchange on WorldClient slot + _reconnectInProgress CAS with proper finally cleanup
3. ✅ Socket cleanup — newClient.Disconnect() on failure paths
4. ✅ TCP reason logging — exception type, message, and SocketErrorCode forwarded to session.unplanned_dc.detected for Warden pattern analysis

Approved for v5.0.9/v5.0.10.

Issue #5, opened 2026-04-22
Cannot use Aqual/Eternal Quintessence in Molten Core


TakaNoDan:
**Describe the bug**
Cannot use Aqual/Eternal Quintessence in Molten Core after having killed bosses. Needs to be tested.

https://github.com/WowLegacyCore/HermesProxy/issues/374

**Game Server**
Everlook

**Hermes Proxy Version**
HermesProxy and Xian

Mongrul:
I have logging setup for this so we can catch it. We will need to leave this open until MC opens up.

 Plant scoped Log.Event diagnostics on the item-on-GameObject cast path so
  when Molten Core opens we can root-cause the Aqual/Eternal Quintessence
  "out of range" bug from one bundle, rather than burning raid attempts on
  trial-and-error. Bug originally reported in WowLegacyCore/HermesProxy#374
  on a CMaNGOS TBC server; needs to be verified on Kronos when MC unlocks.

  - spell.use_item.target_dump (gated on Target.Flags & GameObject) — logs
    spell ID, item ID, target GUID 128/64, target entry, target flags, and
    src/dst locations for any item-on-GameObject cast. Absence of this event
    during a failed Quintessence attempt is itself a signal (= client-side
    pre-validation refused to send).
  - spell.cast_failed.diag — logs raw + named SpellCastResult reason for any
    cast failure, paired with spell ID. Already low-volume.

Mongrul:
Fix is in for this, it is sitting as a PR rigth now. If it fails robust logging will be in place.  Will leave open until we confirm it works. 


 Aqual (17333) and Eternal (22754) Quintessence cast spell 21358 (Dousing,
  SPELL_EFFECT_OPEN_LOCK, 5y interact range) against the seven Runes of
  Warding in Molten Core (176951–176957). On the modern 1.14 client through
  HermesProxy, the cast fails with "Out of range" before any packet leaves
  the client — Majordomo never spawns, raid can't progress to Ragnaros.
  Tracked upstream as WowLegacyCore/HermesProxy#374, classified in our
  own RESEARCH.md as Tier 3 / "Spell target validation" but never root-caused.

  Root cause: silent bit-value drift in GameObjectFlags between vanilla 1.12
  and the modern client.

    Vanilla 1.12                                Modern1.14
    ──────────────────────              ──────────────────────────────────
    0x10  GO_FLAG_NO_INTERACT           0x10  NOT_SELECTABLE
[...]

Issue #20, opened 2026-04-22
Nefarian showing 'out of range', only melee swings work


TakaNoDan:
**Describe the bug**

During the Nefarian boss fight, all abilities show "out of range" and only melee auto-attack swings work.

https://github.com/Xian55/HermesProxy/issues/19
https://github.com/WowLegacyCore/HermesProxy/issues/354

**Game Server**
1.12 / All / Everlook

**Hermes Proxy Version**
Hermes / Xian

Mongrul:
Flying mobs in general are handled differently in 1.14. I have a fix that works decent. However this would not have fixed Nef, He is considered "Perched" so willl require a different approach. I will have Claude try to find a similar in world Perched mob to fix.

https://github.com/jameopotato/jimsproxy/issues/48

Explanation 

The modern client handles animation rigs, skeletons, and Z-axis transitions entirely differently than the original 1.12 server core. When a mob takes off or hovers today, the modern engine expects a specific set of complex animation states that simply did not exist 20 years ago.

Our proxy is doing some heavy lifting in real-time to translate the legacy server's movement data into something the modern client understands. Because those animation skeletons aren't identical, there is no perfect 1-to-1 mapping. It plays well and looks decent, but you might occasionally spot a slightly awkward transition or a hovering animation that isn't perfectly smooth.

Pic from https://github.com/WowLegacyCore/HermesProxy/issues/354

<img width="1864" height="1076" alt="Image" src="https://github.com/user-attachments/assets/f9e17537-140a-4636-85f4-c54c14bf4e2a" />

Mongrul:
Will be waiting until BWL phase for this fix.

PR #121, opened 2026-05-03
 Fix Molten Core: Aqual/Eternal Quintessence no longer fails with "Ou…

Mongrul:
…t of range"

 Aqual (17333) and Eternal (22754) Quintessence cast spell 21358 (Dousing,
  SPELL_EFFECT_OPEN_LOCK, 5y interact range) against the seven Runes of
  Warding in Molten Core (176951–176957). On the modern 1.14 client through
  HermesProxy, the cast fails with "Out of range" before any packet leaves
  the client — Majordomo never spawns, raid can't progress to Ragnaros.
  Tracked upstream as WowLegacyCore/HermesProxy#374, classified in our
  own RESEARCH.md as Tier 3 / "Spell target validation" but never root-caused.

  Root cause: silent bit-value drift in GameObjectFlags between vanilla 1.12
  and the modern (Cataclysm-era) client.

    Vanilla 1.12 (cmangos)              Modern (TrinityCore-derived 1.14)
    ──────────────────────              ──────────────────────────────────
    0x10  GO_FLAG_NO_INTERACT           0x10  NOT_SELECTABLE
    0x20  GO_FLAG_NODESPAWN             0x20  NODESPAWN
    0x40  GO_FLAG_TRIGGERED             0x40  AI_OBSTACLE
    ─                                   0x80  FREEZE_ANIMATION

  The runes ship from the legacy server with `flags=16` (0x10). In vanilla,
  NO_INTERACT blocks right-click gossip but the spell cursor can still
  acquire the GameObject for OPEN_LOCK — which is how dousing has worked
  since 2005 (cmangos's molten_core.cpp:88-103 confirms the script never
  clears this bit; gating is handled by the Flames Circle linkedTrap GOs
  178187-178193 physically covering each rune until the matching boss dies).
  The modern client reinterprets bit 0x10 as NOT_SELECTABLE — "not selectable
  even in GM mode" — which blocks spell cursor acquisition too. The proxy
  was passing the raw uint32 through at UpdateHandler.cs:3378, so the modern
  client refused to acquire the rune; the cursor click resolved to no
  target, and the range check trivially failed → "Out of range."

  This was never noticed for the same reason the dynamic-flag translation
  was easy to miss: the bits visually match between expansions in most
  slots. The static GameObjectFlags translation simply didn't exist —
  `grep` for `GO_FLAG`/`GameObjectFlags` returned zero hits across the
  whole repo before this PR.

  Fix:

    1. Add `GameObjectFlagsLegacy` and `GameObjectFlagsModern` enums plus a `ToModern()` extension. Same-name bits (InUse, Locked, InteractCond, Transport, Nodespawn) translate via the existing CastFlags<> name-matching helper. Vanilla NoInteract maps explicitly to modern NotSelectable; vanilla Triggered is dropped (no modern equivalent — the mod ern bit at 0x40 is AiObstacle, an unrelated meaning). Mirrors the existing GameObjectDynamicFlags.cs pattern.

    2. Entry-gated override at UpdateHandler.cs: for entries 176951-176957, strip NotSelectable from the translated flags so the modern client can acquire the rune as a Quintessence target. Vanilla's gating mechanic is preserved — the Flames Circle GOs still occlude the rune until the matching boss dies, just like vanilla. Same entry-gated pattern as the existing Deeprun Tram workaround at UpdateHandler.cs:3399-3431.

    3. New `gameobject.flags.translated` JSONL event logs entry, type, legacy raw + flags, modern flags, and whether the rune override fired. Catches future flag-translation surprises across all content without needing another targeted investigation.

  Bonus correctness improvement: vanilla GO_FLAG_TRIGGERED (0x40) was
  silently being read by the modern client as AI_OBSTACLE for any
  vanilla GameObject with that bit set. Translation now drops it
  correctly. Unknown gameplay impact in the wild but at minimum it was
  mislabeling triggered/summoned objects as pathing obstacles for the
  client.

  Files:

    HermesProxy/World/Enums/GameObjectFlags.cs        new, 41 lines
    HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs:3375-3404
                                                      +27 / -1
    HermesProxy.Tests/World/GameObjectFlagsTranslationTests.cs
                                                      new, 50 lines, 8 cases

  Tests: 320/320 pass. New test cases cover same-name pass-through (5),
  explicit NoInteract→NotSelectable mapping, Triggered drop verification,
  and the actual rune template flag combination (NoInteract | Nodespawn).
  No protocol-level changes, no CSV/data changes, change reversible by
  removing the cast.

  Closes equivalent of WowLegacyCore/HermesProxy#374.

### DIFF
--- a/HermesProxy.Tests/World/GameObjectFlagsTranslationTests.cs
+++ b/HermesProxy.Tests/World/GameObjectFlagsTranslationTests.cs
@@ -1,0 +1,50 @@
+using HermesProxy.World.Enums;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+public class GameObjectFlagsTranslationTests
+{
+    // Same-named bits should round-trip to the modern enum at the same value.
+    [Theory]
+    [InlineData(GameObjectFlagsLegacy.InUse,        GameObjectFlagsModern.InUse)]
+    [InlineData(GameObjectFlagsLegacy.Locked,       GameObjectFlagsModern.Locked)]
+    [InlineData(GameObjectFlagsLegacy.InteractCond, GameObjectFlagsModern.InteractCond)]
+    [InlineData(GameObjectFlagsLegacy.Transport,    GameObjectFlagsModern.Transport)]
+    [InlineData(GameObjectFlagsLegacy.Nodespawn,    GameObjectFlagsModern.Nodespawn)]
+    public void ToModern_NameMatchedFlags_PassThrough(GameObjectFlagsLegacy legacy, GameObjectFlagsModern expected)
+    {
+        Assert.Equal(expected, legacy.ToModern());
+    }
+
+    // Vanilla NoInteract (0x10) is the rune-style "right-click blocker"; modern same-bit value
+    // means NotSelectable ("not selectable even in GM mode"). Map explicitly to NotSelectable.
+    [Fact]
+    public void ToModern_NoInteract_MapsToNotSelectable()
+    {
+        var modern = GameObjectFlagsLegacy.NoInteract.ToModern();
+        Assert.Equal(GameObjectFlagsModern.NotSelectable, modern);
+    }
+
+    // Vanilla Triggered (0x40) has no modern equivalent; modern bit 0x40 is AiObstacle, which
+    // would mislabel any vanilla GO with TRIGGERED set. Drop it during translation.
+    [Fact]
+    public void ToModern_Triggered_IsDropped()
+    {
+        var modern = GameObjectFlagsLegacy.Triggered.ToModern();
+        Assert.Equal((GameObjectFlagsModern)0, modern);
+        Assert.False(modern.HasFlag(GameObjectFlagsModern.AiObstacle));
+    }
+
+    // Combined: rune template flags = 0x10 (NoInteract) + 0x20 (Nodespawn). Modern result must
+    // include Nodespawn and NotSelectable, never AiObstacle.
+    [Fact]
+    public void ToModern_RuneTemplateFlags_TranslatesCorrectly()
+    {
+        var legacy = GameObjectFlagsLegacy.NoInteract | GameObjectFlagsLegacy.Nodespawn;
+        var modern = legacy.ToModern();
+        Assert.True(modern.HasFlag(GameObjectFlagsModern.NotSelectable));
+        Assert.True(modern.HasFlag(GameObjectFlagsModern.Nodespawn));
+        Assert.False(modern.HasFlag(GameObjectFlagsModern.AiObstacle));
+    }
+}
--- a/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
@@ -3375,7 +3375,47 @@ public partial class WorldClient
             int GAMEOBJECT_FLAGS = LegacyVersion.GetUpdateField(GameObjectField.GAMEOBJECT_FLAGS);
             if (GAMEOBJECT_FLAGS >= 0 && updateMaskArray[GAMEOBJECT_FLAGS])
             {
-                updateData.GameObjectData.Flags = updates[GAMEOBJECT_FLAGS].UInt32Value;
+                uint legacyRaw = updates[GAMEOBJECT_FLAGS].UInt32Value;
+                var legacyFlags = (GameObjectFlagsLegacy)legacyRaw;
+                var modernFlags = legacyFlags.ToModern();
+
+                // MC Runes of Warding (176951-176957): vanilla GO_FLAG_NO_INTERACT did NOT block
+                // spell-cursor OPEN_LOCK targeting (Quintessence dousing). Modern NOT_SELECTABLE
+                // does. The Flames Circle linkedTrap GOs (178187-178193) handle gating instead —
+                // they cover the rune until the boss dies. Strip NOT_SELECTABLE so the modern
+                // client can acquire the rune as a Quintessence target. Static INTERACT_COND
+                // also blocks cursor acquisition (verified empirically), so we instead inject
+                // the dynamic LO_NO_INTERACT bit (modern 0x080) — which TrinityCore uses for
+                // depleted gathering nodes (visible-but-inert). Closest analog to vanilla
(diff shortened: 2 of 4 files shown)


jameopotato:
## Deep Review: PR #121 — Fix Molten Core Quintessence (GameObjectFlags Translation)

### 1. Summary

Aqual/Eternal Quintessence fails with "Out of range" when used on MC Runes of Warding (entries 176951-176957) through the proxy. Root cause: vanilla `GO_FLAG_NO_INTERACT` (bit 0x10) was passed through raw to the modern client, where that same bit means `NOT_SELECTABLE` — which blocks spell cursor acquisition entirely. The vanilla client allowed spell-cursor OPEN_LOCK even with `NO_INTERACT` set; the modern client doesn't. This is a raid progression blocker — without dousing the runes, Majordomo never spawns and the raid can't reach Ragnaros. Closes the equivalent of upstream HermesProxy#374 and local issue #5.

### 2. Correctness

**Flag translation** (`GameObjectFlags.cs`):

The `ToModern()` method correctly leverages `CastFlags<T>` for same-named flags (InUse, Locked, InteractCond, Transport, Nodespawn), then explicitly maps `NoInteract → NotSelectable`. This is correct because:
- CastFlags maps by **name**, not by bit position
- "NoInteract" has no name match in the modern enum (modern's 0x10 is "NotSelectable")
- Without the explicit mapping, NoInteract would be silently dropped — which would be wrong (the GO should still be interaction-restricted)

**Triggered (0x40) drop** is correct. Vanilla's `TRIGGERED` means "summoned by spell/event"; modern's 0x40 is `AI_OBSTACLE` (pathing obstacle). Different semantics — better to drop than mislabel.

**MC Rune entry-gated override** is well-designed:
- Strips `NotSelectable` from static flags (restores spell cursor acquisition)
- Injects `GameObjectDynamicFlagsModern.NoInteract` (0x080) — the dynamic "visible but inert" flag — to suppress gossip/right-click
- This two-layer approach correctly mimics vanilla behavior: spell cursor works, right-click doesn't

**PrintNotification hint** in `GameObjectHandler.cs` — nice UX touch. Fires before the use packet forwards, so it's non-blocking. The CMSG_GAME_OBJ_USE still reaches the server (which silently ignores it for lock-type GOs).

**Entry range 176951-176957** — correct, these are the 7 Runes of Warding. Well-known static content.

**Edge cases:**
- Other GOs with vanilla `NO_INTERACT` now map to modern `NOT_SELECTABLE`, which is more restrictive than vanilla's behavior (blocks spell cursor too, not just right-click). If any other GOs rely on spell-cursor-while-NO_INTERACT, they'd need their own entry-gated override. Not a problem today, but the pattern is established for when/if it surfaces.

**Tests**: 8 cases covering same-name pass-through (5), explicit NoInteract→NotSelectable, Triggered drop, and the combined rune template flag scenario. Solid coverage.

### 3. Efficiency

The `ToModern()` call uses `CastFlags<T>` which has a per-type-pair FrozenDictionary cache — first call builds the mapping, subsequent calls are O(n) over set bits with cached lookups. For GameObjectFlags with typically 1-3 bits set, this is negligible.

**One concern**: The `gameobject.flags.translated` Log.Event fires on **every GO update with GAMEOBJECT_FLAGS** — not gated behind `Settings.DebugOutput`. In areas dense with GameObjects (cities, instances), this could add measurable JSONL I/O. Consider gating it behind DebugOutput, or at minimum behind a "this is actually interesting" condition (e.g., only log when `mc_rune_override_applied` is true, or when legacy and modern flags differ).

### 4. Future compatibility

- **Upstream rebase**: Adds new files (`GameObjectFlags.cs`, test file) — no conflict risk there. The `UpdateHandler.cs` change at the GAMEOBJECT_FLAGS site is a clean replacement of a single raw-passthrough line — easy rebase. Xian55's repo doesn't have GO flag translation either, so this is net-new.
- **Conventions**: Follows the `DynamicFlags.cs` enum+extension pattern exactly. Entry-gated override matches the Deeprun Tram pattern at line 3393. Good.
- **Kronos coupling**: The entry IDs are vanilla WoW content (not Kronos-custom). The flag translation itself is generic and benefits all 1.12 servers. The rune override is specific to MC content that exists on every vanilla server.
- **Nefarian (#20)**: Worth investigating whether this same class of bug (vanilla flag bit reinterpreted as modern flag bit) explains the Nefarian "out of range" issue. The GO flag translation infrastructure from this PR would make that fix trivial if so.

### 5. Three-way comparison

| Criteria | PR approach (translate + entry override) | Alt A: Strip NotSelectable from ALL NoInteract GOs | Alt B: Server-side script fix on Kronos |
|----------|------------------------------------------|----------------------------------------------------|-----------------------------------------|
| Complexity | Medium — new enum, extension, 2 site changes | Low — one-liner in ToModern() | Zero proxy changes |
| Runtime efficiency | Negligible | Negligible | N/A |
| Maintainability | Good — entry override pattern is reusable for future GO bugs | Risky — may break GOs that should be not-selectable | Not our code |
| Rebase safety | Good — new files + localized change | Good | N/A |
| Edge-case coverage | Correct for runes, conservative for everything else | Over-permissive for non-rune GOs | Server-specific, not portable |

**Recommendation**: PR approach is correct. The two-layer design (generic translation + entry-gated fixup) is the right architecture — it's reusable, testable, and conservative.

### 6. Verdict

**Approve with suggestions:**

1. **Gate the Log.Event** behind `Settings.DebugOutput` (or narrow it to only fire when the override triggers / flags actually differ). Every GO update with flags hits this — it's a hot path in GO-dense areas.
2. **Worth investigating issue #20 (Nefarian)** as a follow-up — could be the same class of flag-bit reinterpretation bug now that the translation infrastructure exists.

Ready for v5.0.9.

jameopotato (dismissed):
Approved for v5.0.9. Closes issue #5.

One suggestion: gate the `gameobject.flags.translated` Log.Event behind `Settings.DebugOutput` — it fires on every GO flag update and could add measurable JSONL I/O in GO-dense areas (cities, instances). Not a merge blocker.

jameopotato (approved):
Approved for v5.0.9. Closes #5. Gate gameobject.flags.translated Log.Event behind DebugOutput when convenient.

PR #117, opened 2026-05-03
NPC + pet scale parity, Charge fix, and quest/mount/destroy diagnostics

Mongrul:
Fixes:
  - Charge/Intercept stun visual now plays on the target instead of the caster (587cc16)
  - Hunter/warlock pets render at vanilla 1.12 size via inverse-CMS scaling, K=1.5 (fec46bb)
  - NPC scale (e.g. Varimathras) restored to vanilla rendering by setting
    UNIT_FIELD_DISPLAY_SCALE = 1/CMS and refreshing 39 stale CreatureDisplayInfo
    CMS values from wago.tools 1.14.2.42597 (2ed1c67)

  Diagnostics (additive, structured Log.Events for JSONL bundles):
  - unit.scale.context, unit.scale.display_scale_set — diagnose "X is the wrong size"
  - unit.mount.changed, unit.dynamic_flags.changed, unit.npc_flags.changed —
    diagnose mounted-quest-NPC and quest-state issues (55c2048)
  - object.destroy, object.far_object — diagnose despawn issues such as the
    Aean Swiftriver Outrunner cat-mount persistence bug (4e4ac4b)

PR #122, opened 2026-05-03
Clear mounts from dead non-player units

erkagoon:
## Summary

This fixes mounted legacy NPCs remaining visually mounted after death on modern clients.

On 1.12 servers, some mounted NPCs keep `UNIT_FIELD_MOUNTDISPLAYID` set after they die. HermesProxy was forwarding that state as-is, so the modern client continued rendering the dead unit with its mount until the object eventually went out of range.

The fix clears `MountDisplayID` for non-player units when their health reaches `0`.

## Investigation

The issue was reproduced with the mounted NPC `Marcus Bel`.

Relevant observed unit:

```text
Name: Marcus Bel
Entry: 5800
GUID: Creature-0-1-0-0-5800-0000004419
Low: 17433
DisplayID: 4347
At spawn, the server sent:

Health: 100
MountDisplayID: 2410
Flags: 524288
Flags2: 2048
During combat/death, the server sent health updates:

Health: 49
Health: 7
Health: 0
At death, the server also sent:

Flags: 0
However, it never sent:

UNIT_FIELD_MOUNTDISPLAYID = 0
So HermesProxy kept the cached mount value:

CachedMountDisplayID: 2410
The object was not destroyed immediately either. Marcus was only removed later through an out-of-range update:

object.out_of_range.debug
Entry: 5800
Health: 0
DisplayID: 4347
CachedMountDisplayID: 2410
There was no object.destroy.debug for this unit at death time.

Root Cause
The legacy server is authoritative and sends the NPC death state correctly through:

UNIT_FIELD_HEALTH = 0
But it does not clear the unit mount field:

UNIT_FIELD_MOUNTDISPLAYID
On the original 1.12 client this appears to be handled correctly, but the modern client relies on the converted update fields it receives from HermesProxy. Since HermesProxy preserved the stale mount display value, the modern client continued to render the dead NPC as mounted.

In other words, the converted modern update represented the unit as:

Health = 0
MountDisplayID = 2410
which is visually wrong for the modern client.

Fix
The fix is implemented in:

HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
When HermesProxy builds an outgoing modern unit update, it now detects this case:

non-player unit
Health == 0
MountDisplayID still set
and forces:

MountDisplayID = 0
This is done both on the outgoing ObjectUpdate data and in the legacy update dictionary used by the conversion path:

updateData.UnitData.MountDisplayID = 0;
updates[UNIT_FIELD_MOUNTDISPLAYID] = new UpdateField(0);
The behavior is intentionally scoped to non-player units only:

ObjectType.Unit
HighGuidType.Creature
HighGuidType.Pet
HighGuidType.Vehicle
It explicitly excludes:

players
active players
current player GUID
so it should not affect player mount state.

Why This Approach
This is a proxy-side compatibility fix for legacy server behavior.

The legacy server may not need to send UNIT_FIELD_MOUNTDISPLAYID = 0 for the 1.12 client to behave correctly, but the modern client does need the mount state cleared in the converted update stream.

Rather than changing object despawn behavior or synthesizing destroy packets, this fix only corrects the stale unit field at the moment the unit becomes dead.

That keeps the change small and avoids interfering with normal object lifetime handling.

Verification
Tested manually in game with the mounted NPC death case.

Before the fix:

NPC reached Health = 0
MountDisplayID stayed cached
Dead unit remained visually mounted
Object disappeared only when it went out of range
After the fix:

NPC reaches Health = 0
HermesProxy clears MountDisplayID
Dead mounted visual no longer persists
Build verification:

dotnet build HermesProxy/HermesProxy.csproj --configfile NuGet.Config -c Release
Result:

Build succeeded
0 warnings
0 errors

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
@@ -1371,6 +1371,32 @@ public partial class WorldClient
             }
         }
     }
+
+    private bool ShouldClearMountDisplayOnDeadNonPlayerUnit(WowGuid128 guid, ObjectType objectType, ObjectUpdate updateData, Dictionary<int, UpdateField> updates, int mountDisplayField)
+    {
+        if (updateData.UnitData.Health.GetValueOrDefault(-1) != 0)
+            return false;
+
+        if (guid.IsPlayer() ||
+            guid == GetSession().GameState.CurrentPlayerGuid ||
+            objectType == ObjectType.Player ||
+            objectType == ObjectType.ActivePlayer)
+            return false;
+
+        bool isNonPlayerUnit =
+            objectType == ObjectType.Unit ||
+            guid.GetHighType() == HighGuidType.Creature ||
+            guid.GetHighType() == HighGuidType.Pet ||
+            guid.GetHighType() == HighGuidType.Vehicle;
+        if (!isNonPlayerUnit)
+            return false;
+
+        int mountDisplayId = updateData.UnitData.MountDisplayID.GetValueOrDefault();
+        if (mountDisplayId == 0 && mountDisplayField >= 0 && updates.TryGetValue(mountDisplayField, out var cachedMountDisplay))
+            mountDisplayId = cachedMountDisplay.Int32Value;
+
+        return mountDisplayId != 0;
+    }
     
     private void StoreObjectUpdateInternal(WowGuid128 guid, ObjectType objectType, BitArray updateMaskArray, Dictionary<int, UpdateField> updates, AuraUpdate auraUpdate, PowerUpdate? powerUpdate, bool isCreate, ObjectUpdate updateData)
     {
@@ -1893,6 +1919,12 @@ public partial class WorldClient
             if (UNIT_FIELD_MOUNTDISPLAYID >= 0 && updateMaskArray[UNIT_FIELD_MOUNTDISPLAYID])
             {
                 updateData.UnitData.MountDisplayID = updates[UNIT_FIELD_MOUNTDISPLAYID].Int32Value;
+            }
+            if (ShouldClearMountDisplayOnDeadNonPlayerUnit(guid, objectType, updateData, updates, UNIT_FIELD_MOUNTDISPLAYID))
+            {
+                updateData.UnitData.MountDisplayID = 0;
+                if (UNIT_FIELD_MOUNTDISPLAYID >= 0)
+                    updates[UNIT_FIELD_MOUNTDISPLAYID] = new UpdateField(0);
             }
             int UNIT_FIELD_MINDAMAGE = LegacyVersion.GetUpdateField(UnitField.UNIT_FIELD_MINDAMAGE);
             if (UNIT_FIELD_MINDAMAGE >= 0 && updateMaskArray[UNIT_FIELD_MINDAMAGE])


jameopotato:
## Deep Review: PR #122 — Clear mounts from dead non-player units

### 1. Summary

Mounted NPCs (like Marcus Bel, entry 5800) keep `UNIT_FIELD_MOUNTDISPLAYID` set after death on the legacy 1.12 server. The vanilla client handled this implicitly, but the modern 1.14 client renders the dead NPC as still mounted — a visual glitch that persists until the object goes out of range. The fix clears `MountDisplayID` to 0 for non-player units when health reaches 0.

### 2. Correctness

The detection logic is sound. Using `-1` as `GetValueOrDefault` correctly handles the case where Health isn't in this update packet (returns false, no clearing). The guard only fires when Health is explicitly 0.

The entity filtering is correct — excludes players/ActivePlayer and only targets Creature/Pet/Vehicle high-GUIDs. The belt-and-suspenders approach (checking both `objectType` and `guid.GetHighType()`) is defensive but reasonable given the proxy's mixed-expansion type system.

The mount display fallback is well handled — reads from `updateData` first, then falls back to the `updates` dictionary. This correctly covers both a mount set in the current update AND a previously-cached mount that wasn't re-sent.

**One edge case worth noting**: A non-player unit that spawns dead WITH a mount display (e.g., a dead rider decoration object) would have its mount cleared on the initial CREATE_OBJECT. Unlikely to matter for any real content, but documenting for completeness.

**No risk to existing behavior**: The clearing only fires for dead non-player units. Players, living NPCs, and all other update paths are untouched.

### 3. Efficiency

`ShouldClearMountDisplayOnDeadNonPlayerUnit` is called once per unit update that includes MOUNTDISPLAYID, right after the field is read. Zero allocations, no lookups beyond what's already in scope. The guard returns false early in the common case (unit is alive). Negligible overhead.

### 4. Future compatibility

- **Upstream rebase**: Modifies `UpdateHandler.cs` (the most-touched file). Will conflict with #117 and #124 on nearby lines, but the change is a clean insertion after the MOUNTDISPLAYID read — straightforward to resolve.
- **Conventions**: Follows the existing pattern of post-read fixups in `StoreObjectUpdateInternal`. Method naming matches codebase style.
- **Kronos coupling**: Generic fix — any 1.12 server could leave mount display set on dead NPCs. Not Kronos-specific.

### 5. Three-way comparison

| Criteria | PR approach (proxy-side clear) | Alt A: Synthesize SMSG_DESTROY_OBJECT on death | Alt B: Client-side addon to hide dead mounts |
|----------|-------------------------------|------------------------------------------------|----------------------------------------------|
| Complexity | Low — single field clear | Medium — must track death state, synthesize packet | Low — but addon-dependent |
| Runtime efficiency | Negligible (field write) | Higher (packet synthesis + send) | Zero proxy cost |
| Maintainability | Clear intent, self-contained | Interleaves with object lifetime | External dependency |
| Rebase safety | Good — localized change | Poor — touches despawn path | N/A |
| Edge-case coverage | Handles all dead mounted NPCs | Over-broad — destroys the whole object | Fragile — depends on addon updates |

**Recommendation**: PR approach is the right one. Minimal correct fix.

### 6. Verdict

**Approve** — clean, focused, correct. Good investigation in the PR description. Ready for v5.0.9.

jameopotato (dismissed):
Approved for v5.0.9. Clean, focused visual fix — minimal risk.

jameopotato (approved):
Approved for v5.0.9. Clean visual fix, minimal risk.

PR #127, opened 2026-05-04
Auction sell-side: merge multi-item CMSGs, throttle, adaptive split p…

Mongrul:
…olling

Modern (3.2.2a+) clients send CMSG_AUCTION_SELL_ITEM with a list of (itemGuid, useCount) pieces and expect the server to combine them into one stack for one auction. Pre-3.2.2a vanilla servers have no per-item count field and auction the entire stack at one GUID, so the proxy now combines pieces client-side via SPLIT/SWAP into a temp slot before forwarding the legacy sell. Without this, addons like Aux that span a posted stack across multiple inventory pieces (e.g. last 6 of slot A + first 1 of slot B = a 7-stack) end up creating fragmented auctions.

Changes:
- Throttle CMSG_AUCTION_SELL_ITEM to a 500 ms minimum gap. Aux fires sells ~190 ms apart, well under what a human at the Blizzard UI could reach, and risks tripping per-socket flood thresholds. Sells are queued (handler thread sleeps), never dropped.
- Fix GUID bug in single-item split path: was sending the original source GUID to CMSG_AUCTION_SELL_ITEM after splitting to a temp slot, causing the server to auction the source remainder while the split partial stayed orphaned in the temp slot.
- Multi-item CMSG support: combine all pieces into one shared target slot, then auction the merged GUID once. Per piece:
    * useCount < srcStack: single CMSG_SPLIT_ITEM (auto-merges into a same-item-type target slot)
    * useCount == srcStack and srcStack >= 2: SPLIT (srcStack-1) then CMSG_SWAP_ITEM the leftover 1 with target. Vanilla SPLIT rejects count >= stack with EQUIP_ERR_TRIED_TO_SPLIT_MORE_THAN_COUNT; vanilla Player::SwapItem auto-stacks same-item-type slots, so the SWAP lands the final unit via merge instead of swap.
    * srcStack == 1: SWAP only (auto-stacks if target same type, else moves to empty target).
- Adaptive WaitForInventoryChange polling replaces fixed 750 ms sleeps after each SPLIT/SWAP. Snapshots dest slot pre-state, polls every 25 ms for guid/stack change, returns as soon as the SMSG inventory update lands. 750 ms ceiling kept as safety net. Observed Kronos PTR settle times cluster at 150-190 ms; saves ~9 sec on a 10-stack post.
- Structured diagnostics throughout: auction.sell.{throttled, split_sent, split_settled, posted, merge_failed, merge_skip_piece, merge_piece_sent, merge_piece_settled, merge_piece_resolved, posted_merged}.

### DIFF
--- a/HermesProxy/World/Server/PacketHandlers/AuctionHandler.cs
+++ b/HermesProxy/World/Server/PacketHandlers/AuctionHandler.cs
@@ -21,6 +21,22 @@ public partial class WorldSocket
     private const double AuctionSearchCooldownSeconds = 6.0;
     private DateTime _lastSearchTime = DateTime.MinValue;
 
+    // Minimum gap between forwarded CMSG_AUCTION_SELL_ITEM packets. Aux and similar
+    // batch-posting addons can fire sells ~190 ms apart, which is well under the
+    // cadence a human at the Blizzard UI could ever reach and risks tripping general
+    // per-socket flood thresholds on vanilla emulators. Sells are queued (not dropped)
+    // by sleeping the handler thread until the gap is satisfied.
+    private const int AuctionSellMinGapMs = 500;
+    private DateTime _lastSellTime = DateTime.MinValue;
+
+    // Maximum time to wait after a CMSG_SPLIT_ITEM / CMSG_SWAP_ITEM for the resulting
+    // SMSG_ITEM_PUSH_RESULT / object update to arrive and update GameState before we
+    // read the destination slot back. We poll adaptively (see WaitForInventoryChange)
+    // and return the moment a change is observed; this 750 ms ceiling is the safety
+    // net for a slow / stalled server. Typical update arrival is 50-200 ms.
+    private const int AuctionSplitSettleMs = 750;
+    private const int AuctionSplitPollMs = 25;
+
     [PacketHandler(Opcode.CMSG_AUCTION_HELLO_REQUEST)]
     void HandleAuctionHelloRequest(InteractWithNPC interact)
     {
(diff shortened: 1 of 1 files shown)


jameopotato (changes requested):
## Review: PR #127 — Auction sell-side: merge multi-item CMSGs, throttle, adaptive split

The merge logic and adaptive polling are well-designed. The GUID fix in the single-item split path is a real bug fix. However, several concerns need investigation before merging.

### Needs investigation

1. **Handler thread blocking.** Thread.Sleep on the WorldSocket handler thread blocks ALL modern→legacy packet processing for the session — not just auction packets. During a 10-piece batch post: 500ms throttle + 10 × ~200ms settle = ~2.5 seconds where the player can't move, cast, or interact. Aux power-sellers posting 50+ items per session will feel repeated micro-freezes. Investigate offloading the merge+sell sequence to a background Task (similar to #119's reconnect pattern) so the handler thread stays responsive.

2. **No rollback on partial merge failure.** If piece 3 of 5 fails (server rejects the SPLIT, dest slot full, server stall), pieces 1-2 are already in the temp slot. The code returns without selling or returning items, leaving a partial stack in a slot the player didn't put there. Items aren't lost but it's confusing. Consider: (a) SWAP items back to original slots on failure, or (b) sell whatever successfully merged (partial is better than nothing).

3. **CMSG_SWAP_ITEM field order.** The code writes (dstBag, dstSlot, srcBag, srcSlot) which the comment says is correct for vanilla. Verify against cmangos WorldSession::HandleSwapItem — getting this backwards silently corrupts inventory state instead of erroring. A unit test that verifies packet layout against known-good vanilla captures would add confidence.

4. **Concurrent inventory mutation.** FindEmptyInventorySlot and FindItemInInventory read GameState that the WorldClient thread updates asynchronously. The adaptive polling handles the settle-wait correctly, but there's a TOCTOU window: between FindEmptyInventorySlot returning and the first SPLIT arriving at the server, another packet (loot, trade, quest reward) could fill that slot. Consider checking the slot is still empty after the first SPLIT settles.

5. **Throttle is session-global, not per-auctioneer.** If the player visits two auctioneers in quick succession (e.g., neutral AH → faction AH), the 500ms throttle from the first carries into the second. Probably fine in practice but worth documenting.

6. **Auction search cooldown diagnostic.** The `auction.list.dropped_cooldown` event and the hypothesis about Aux hanging (CanSendAuctionQuery permanently false) is good diagnostic thinking. If confirmed, the fix should be a separate PR — queue the CMSG until cooldown expires instead of dropping it.

### What's ready

- Adaptive WaitForInventoryChange polling — correct, well-implemented, saves ~5.6 seconds per 10-stack post
- Single-item split GUID fix — real bug fix, should ship regardless
- Throttle concept — right idea, may need the gap to be configurable
- Diagnostic events — thorough, good JSONL coverage

### Suggestion

Consider splitting this PR: ship the single-item GUID fix + diagnostics as a quick win (low risk), then iterate on the multi-item merge in a follow-up with the thread-blocking and rollback concerns addressed.

jameopotato:
Added player-facing chat messages in `2473f51`:

- **Merge starting:** "Merging N item stacks for auction..." — so the player knows the proxy is working during the brief freeze
- **No empty slot:** actionable message to free up a slot
- **Merge failed (both bail paths):** "items were moved but not posted — sort your bags into clean stacks and try again"

Outstanding review concerns resolved. Ready to merge.

PR #129, opened 2026-05-04
Fix stuck button on off-GCD spell LoS failure

jameopotato:
## Summary

Off-GCD spells (Sprint, Evasion, Trinket, racials, targeted instants) bypass the hold system and defer SpellPrepare to SPELL_GO time. If the server rejects the cast before SPELL_GO (LoS, range, etc.), the client never receives SpellPrepare and has no tracking entry to resolve against the incoming SpellFailure — the action button stays in a pending/lit state permanently.

## Root cause

Found via JSONL analysis on v5.1.0-beta.2: two `spell.failure.routed` events for spell 7322 with `dequeued=false` after confirmed LoS failure. The cast was enqueued in PendingNormalCasts but the hidden-queue approach meant no SpellPrepare was sent at forward time. The client had no way to match SpellFailure to its pending button press.

## Fix

Send SpellPrepare immediately for off-GCD casts. This doesn't affect GCD sweep timing (off-GCD spells don't trigger GCD), so the hidden-queue rationale for deferring SpellPrepare on on-GCD casts doesn't apply here.

## Test plan

- [ ] Target something behind LoS, cast a targeted spell — button should release after failure
- [ ] Verify AE spam still works (GCD sweep locked at 12 o'clock)
- [ ] Verify off-GCD spells (Sprint, Evasion, Trinket) work normally
- [ ] `dotnet test` 320/320 passing

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -1478,6 +1478,8 @@ public class ClientCastRequest
     // Diagnostic only — used by spell.held_fire to compute hold duration. 0 if never held.
     public long HeldAtTickMs;
 
+    public bool HasSentPrepare;
+
     // JimsProxy: cast time (ms) reported by SMSG_SPELL_START. 0 means instant.
     // Distinguishes truly cast-time spells (Frostbolt, Polymorph) from instants that
     // *also* emit SMSG_SPELL_START on Kronos 1.12 (Arcane Explosion, Counterspell, etc.).
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -730,8 +730,9 @@ public partial class WorldClient
                 spell.Cast.SpellID = (int)pendingCast.SpellId;
 
             // For instant spells that skip SPELL_START, we need to send SpellPrepare
-            // before SpellGo so the client knows which cast completed
-            if (!pendingCast.HasStarted)
+            // before SpellGo so the client knows which cast completed.
+            // Off-GCD casts already sent SpellPrepare at forward time (HasSentPrepare).
+            if (!pendingCast.HasStarted && !pendingCast.HasSentPrepare)
             {
                 SpellPrepare prepare = new();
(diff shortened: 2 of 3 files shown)


jameopotato:
## Review: PR #129 — Fix stuck button on off-GCD spell LoS failure

### 1. Summary

Off-GCD spells bypass the hold system and deferred SpellPrepare to SPELL_GO time. If the server rejects the cast before SPELL_GO (LoS, range), the client has no SpellPrepare to match against SpellFailure and the button stays permanently lit. Fix sends SpellPrepare at forward time for off-GCD casts, with a `HasSentPrepare` flag to prevent duplicate SpellPrepare in HandleSpellGo's `!HasStarted` branch.

### 2. Correctness

Correctly scoped to the off-GCD `else` branch only — on-GCD hidden queue is untouched. The `HasSentPrepare` flag prevents a duplicate SpellPrepare when the cast succeeds (HandleSpellGo's `!HasStarted` branch checks the flag and skips). Found via JSONL analysis: two `spell.failure.routed` events for spell 7322 with `dequeued=false` after LoS failure on beta.2.

### 3. Efficiency

One SpellPrepare packet (~16 bytes) per off-GCD cast. Off-GCD spells are low-frequency. Negligible.

### 4. Future compatibility

Single-line insertion in JimsProxy-specific code. Zero upstream conflict risk. Follows existing SpellPrepare pattern.

### 5. Three-way comparison

| Criteria | PR approach (SpellPrepare at forward) | Alt A: Synthetic CastFailed on unmatched SPELL_FAILURE | Alt B: Skip status≠2 filter for off-GCD |
|----------|--------------------------------------|--------------------------------------------------------|----------------------------------------|
| Complexity | Low — 5+3 lines | Medium — new failure handler path | Low — but changes vanilla semantics |
| Runtime efficiency | One extra packet per off-GCD cast | One packet only on failure | Zero extra packets |
| Maintainability | Clear, follows existing pattern | Special case in failure handler | Risks surfacing benign Kronos notifications |
| Edge-case coverage | Handles success AND failure | Only fixes failure | May pass status=0 to client as errors |

### 6. Verdict

**Approve** — minimal, correct fix with the duplicate guard addressed.

PR #144, opened 2026-05-05
 Fix WTF wipe on logout + recreate WorldClient on relog

Mongrul:
## Summary

  Two related fixes for the logout → char-select → relog flow.

  ### Commit 1 (861c9a5) — hold InstanceSocket open through char-select

  **Bug:** SMSG_LOGOUT_COMPLETE handler was calling `InstanceSocket.CloseSocket()` synchronously after forwarding the
  packet to the modern client. The 1.14 client flushes WTF (keybinds, macros, account/config caches, SavedVariables) as
  part of its char-select transition triggered by LOGOUT_COMPLETE, and that transition takes ~1s on observed runs.
  Slamming the TCP socket microseconds after the packet pushed the client into an error-disconnect code path that
  skipped the flush — wiping the user's keybindings, macros, and account settings.

  Symptom was rare and elusive because the 20s open-world logout countdown gave WoW wide margin to finish flushing
  before whatever it was racing landed; the bug almost always reproduced on instant-logout flows (Exit Game, in-town
  logout) where there's no countdown.

  **Fix:** capture the InstanceSocket reference, null the session pointer immediately so subsequent send paths handle it
   cleanly via existing send-on-closed logic, and hand the socket to a 3s polling task that closes only if the client
  hasn't done so itself. A direct 1.12 client ↔ 1.12 server connection never sees this — the legacy server doesn't
  proactively kill the client on logout-complete; we now mirror that.

  **Diagnostics:** `session.logout_complete.*` JSONL events
  - `socket_hold_started` (`hold_ms`, `poll_interval_ms`)
  - `client_closed_first` (`elapsed_ms`) — normal path
  - `safety_net_closed` (`elapsed_ms`) — 3s expired, proxy closed
  - `safety_net_error` (`message`)

  **Validated bundle 20260505-051822:** `client_closed_first elapsed_ms=1080`, i.e. WoW needed 1.08s to finish flushing
  — under the old code the socket died within microseconds of LOGOUT_COMPLETE, well over a full second before WoW was
  done. 3s buffer leaves ~1.9s headroom.

  ### Commit 2 (04ccdf1) — recreate WorldClient on relog

  861c9a5 keeps BNet/InstanceSocket alive across logout. Side effect: the next char-login arrives as
  `CMSG_AUTH_CONTINUED_SESSION` instead of `CMSG_AUTH_SESSION`, and `HandleAuthContinuedSession` does not run the
  WorldClient-creation code that `HandleAuthSession` does. Twinstar closes the legacy world TCP shortly after each
  logout completes server-side, so by the time `CMSG_PLAYER_LOGIN` for the next char arrives the proxy's `WorldClient`
  is dead or about to be.

  **Two failure modes observed in user bundles after only commit 1:**
  - **WOW51900309** at "Enter World" — `CMSG_PLAYER_LOGIN` forwarded into a closed legacy socket; no
  `SMSG_LOGIN_VERIFY_WORLD` ever returns, modern client times out on the realm/instance handshake.
  - **NullReferenceException** at `ChatHandler.cs:227` — relog landed in world via a still-alive-but-doomed connection
  that died ~3s later, `GetSession().WorldClient` went null, the first `/say` deref'd it (the `!`-suppression doesn't
  help at runtime).

  **Fix:** mirror direct 1.12 client ↔ 1.12 server, which opens a new world TCP per character.
  - `SMSG_LOGOUT_COMPLETE` drops `session.WorldClient` (no proactive `Disconnect()` — we're on the receive thread; let
  Twinstar close on its end).
  - `HandlePlayerLogin`'s null-check now recreates `WorldClient` via `ConnectToWorldServer` instead of bailing with
  `NoWorld`. `AuthClient` is disconnected post-first-login but its session key survives (snapshot in `WorldClient.cs:90`
   already handles that case).

  **Diagnostics:**
  - `session.logout_complete.worldclient_dropped`
  - `session.relogin.worldclient_recreate_attempt`
  - `session.relogin.worldclient_recreated` (with `elapsed_ms`)
  - `session.relogin.worldclient_recreate_failed` (e.g. unknown realm)
  - `session.relogin.stale_worldclient_disconnect_error`

  **Repro bundle 20260505-120057:** char A logout at t=696s, char B login at t=706s via `CMSG_AUTH_CONTINUED_SESSION`
  (no `HandleAuthSession` path), `session.ondisconnect.suppressed` at t=709s nulled `WorldClient`, then NRE at t=758s on
   `/say`.

  ## Test plan

  - [ ] Open-world `/logout` (20s countdown): WTF preserved, relog into different char succeeds, chat works
  - [ ] In-town `/logout` (instant): WTF preserved, relog succeeds, chat works
  - [ ] Exit Game button: WTF preserved
  - [ ] Rapid relog within 3s: socket-hold doesn't interfere
  - [ ] Log-out → pick same character → enter world: chat works (no NRE)
  - [ ] Log-out → pick different character → enter world: chat works, no WOW51900309
  - [ ] JSONL bundle shows `session.relogin.worldclient_recreated { connected: true }` on every relog

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -1588,6 +1588,16 @@ public class GlobalSessionData
     public WorldSocket InstanceSocket = null!;
     public AuthClient AuthClient = null!;
     public WorldClient? WorldClient;
+    // JimsProxy: set true on SMSG_LOGOUT_COMPLETE so the next CMSG_PLAYER_LOGIN
+    // tears down and recreates WorldClient. Twinstar accepts a second
+    // CMSG_PLAYER_LOGIN on the same world TCP (the LOGIN_VERIFY_WORLD comes
+    // back fine) but then closes the connection a few seconds into the new
+    // character's session — leaving session.WorldClient null mid-game. We
+    // can't drop the WorldClient at LOGOUT_COMPLETE itself because char-select
+    // (CMSG_ENUM_CHARACTERS / CMSG_QUERY_PLAYER_NAME / etc.) is forwarded over
+    // the same WorldClient and needs it alive until the user picks a char.
+    // Cleared after the recreate succeeds.
+    public bool WorldClientNeedsRecreateOnNextLogin;
     public SniffFile ModernSniff = null!;
 
     public Dictionary<string, WowGuid128> GuildsByName = [];
--- a/HermesProxy/World/Client/PacketHandlers/CharacterHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/CharacterHandler.cs
@@ -370,32 +370,31 @@ public partial class WorldClient
         // the next character (or the same one re-logging in) starts clean.
         GetSession().ThreatTracker.Reset();
 
-        // JimsProxy (logout-wtf-flush follow-up): drop the session's reference
-        // to this WorldClient. The legacy server's world session for the
-        // outgoing character ends with SMSG_LOGOUT_COMPLETE; Twinstar closes
-        // its end of the world TCP shortly after. Because 861c9a5 keeps the
-        // BNet/InstanceSocket alive across logout, the next char-login arrives
-        // as CMSG_AUTH_CONTINUED_SESSION, which does NOT run the WorldClient-
-        // creation code in WorldSocket.HandleAuthSession. Without this nulling,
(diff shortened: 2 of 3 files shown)


jameopotato:
Superseded by #146 (strict superset — all 3 commits from this PR plus watchdog + spell failure fix). Squash-merged into beta via #146.

PR #145, opened 2026-05-05
Fix/worldclient silent stall watchdog

Mongrul:
## Summary

  The existing unplanned-reconnect path (`TryUnplannedReconnectAndPropagate`) only fires from `HandleDisconnect` /
  receive-loop catch — both require a TCP RST/FIN. Bundle `20260505-131650` caught Twinstar going silent for 150 seconds
   under Stormwind crowd load: TCP socket stayed open (no FIN, no RST), the player kept moving (modern client
  extrapolates), but every server-acknowledged action stalled. Spell casts queued via the GCD-hold mechanism waiting on
  `SMSG_SPELL_GO` that never came; after ~150s of frozen-cast ghost world the WoW client gave up and crashed.

  ## Fix

  Piggyback on the existing 30s `_keepAliveTimer`. Before sending each ping, check `Environment.TickCount -
  _lastInboundOpcodeTick` (already tracked for disconnect diagnostics). If > 60s, declare the WorldClient stale, take
  its slot via `Interlocked.CompareExchange` (so a racing real-disconnect can't double-fire), `Disconnect()` the socket,
   stop the timer, and call `TryUnplannedReconnectAndPropagate`. The reconnect path's own CAS guard is defense in depth.

  60s threshold rationale:
  - keepalive ping fires every 30s and solicits `SMSG_PONG` within ~ms
  - any normal play session sees s2c constantly (broadcasts, ticks, pongs); 30s of silence is already abnormal, 60s is
  unambiguous
  - shorter risks false positives on transient hiccups
  - longer leaves the user staring at a frozen UI

  Either reconnect succeeds and the player resumes, or the modern client gets a clean DC instead of staring at a frozen
  UI for minutes until WoW itself crashes.

  ## Diagnostics

  - `session.worldclient.silent_stall_detected` (with `last_opcode`, `ms_since_last_inbound`, `keepalive_serial`)
  - `session.worldclient.silent_stall_disconnect_error` (defensive, on Disconnect throw)
  - Existing `session.unplanned_dc.detected` / `session.unplanned_reconnect.*` log the recovery attempt

  ## Dependency

  Stacked on `Mongrul:logout-wtf-flush` (PR #144) because it touches the same WorldClient lifecycle. Once #144 merges to
   `beta`, this branch will rebase to a clean 1-commit diff. **Merge order: #144 first, then this.**

  ## Test plan

  - [ ] Long session (20+ min) with no freezes shows zero `session.worldclient.silent_stall_detected` (validated by
  bundle `20260505-134207`: 24 min, no false positives)
  - [ ] If a silent stall recurs: JSONL shows `silent_stall_detected` followed by `unplanned_reconnect.*` and either
  successful resume or clean DC (no frozen ghost world)


jameopotato:
Same commits as #146 but targeting master. #146 (targeting beta) was squash-merged instead. This content will reach master via the beta → master merge for v5.1.0.

PR #161, opened 2026-05-07
Complete in-combat popup suppression: peek instead of dequeue

Mongrul:
Summary

  - 0d873bc (PR #146) suppressed the broadcast SMSG_SPELL_FAILURE popup. A follow-up attempt overrode the inline
  CastFailed.Reason to DontReport to also suppress popup-from-CAST_FAILED — but that introduced two regressions: (a)
  action buttons stayed lit because DontReport doesn't trigger button-state cleanup on the modern client, and (b) the
  trailing real-reason SMSG_CAST_FAILED couldn't dequeue because we'd already taken the entry, so the real reason never
  reached the client.
  - This PR replaces the dequeue-and-fake-reason approach with peek (FirstOrDefault) on both PendingNormalCasts and
  PendingPetCasts. The queue entry survives for HandleCastFailed/HandlePetCastFailed, which then delivers SpellPrepare +
   CastFailed with the real server reason — clearing button-lit AND showing the correct popup (or none, for benign
  reasons).
  - Bundle jimsproxy-20260507-062946.jsonl confirmed the regression: rogue stuck-button on Pick Pocket / Stealth despite
   popup being gone. Bundle jimsproxy-20260507-065335.jsonl confirmed the fix: 5× spell.held_fire_on_failure + 2×
  spell.held_fire_on_success events, Cheap Shot SPELL_START/SPELL_GO/aura apply all firing cleanly, Kidney Shot finisher
   snapshot path intact.

  What's kept vs. removed

  Kept (from 0d873bc / PR #146):
  - skipBroadcastFailure = true for instants (no popup)
  - overrideReasonForLocalBroadcast = true for cast-time spells (broadcast SpellFailure with DontReport dismisses cast
  bar silently)
  - Ranged auto-attack carve-out (Auto Shot 75, Shoot 5019)
  - All spell.failure.routed diagnostic fields

  Removed (broken intermediate fix):
  - Inline CastFailed/PetCastFailed packets that consumed the queue entry and shipped DontReport (which doesn't unlit
  buttons)
  - Inline SpellPrepare (now sent by HandleCastFailed on the trailing SMSG_CAST_FAILED)
  - TryDequeuePendingNormalCast/TryDequeuePendingPetCast calls (now FirstOrDefault peek)

  Trade-off

  On Kronos, where the trailing SMSG_CAST_FAILED can be dropped on cast-time spells when the target dies mid-cast, the
  cast can leak in the queue (blocking HasNonStartedPendingCastForSpell for the same spellId until next press displaces
  it). The SendInterrupted hardcoded-zero path that triggers this code is overwhelmingly instants on vmangos/Twinstar
  where the trailing CAST_FAILED reliably arrives. A watchdog timer for the Kronos case can be added separately if it
  manifests.

  Test plan

  - dotnet build clean (0 warn / 0 err)
  - dotnet test 320/320 passing
  - Twinstar repro: rogue Pick Pocket macro on humanoid mobs, popup gone, buttons clear, held Cheap Shot fires on both
  PP failure and PP success
  - Twinstar Kidney Shot 2 CP: finisher.cmsg_cast snapshot fires, Cheap Shot / Kidney Shot / Sinister Strike all cast
  cleanly
  - PvP regression check: enemy-player Counterspelled cast still dismisses target-frame cast bar (non-local path
  unchanged)
  - Kronos watchpoint: monitor for HasNonStartedPendingCastForSpell blocking same-spell repeats; if observed, add
  watchdog timer in a follow-up

### DIFF
--- a/HermesProxy.Tests/World/WatchdogEvictionTests.cs
+++ b/HermesProxy.Tests/World/WatchdogEvictionTests.cs
@@ -240,4 +240,95 @@ public class WatchdogEvictionTests
 
         Assert.False(session.HasStartedNormalCast());
     }
+
+    // ---- Movement preemption tests ----
+
+    private static ClientCastRequest MakeCastTimeCast(uint spellId, uint castTimeMs, bool hasStarted)
+    {
+        return new ClientCastRequest
+        {
+            SpellId = spellId,
+            Timestamp = Environment.TickCount,
+            HasStarted = hasStarted,
+            StartedCastTimeMs = castTimeMs,
+        };
+    }
+
+    [Fact]
+    public void MarkStartedCastsMovementCancelled_StartedCastTime_Marked()
+    {
+        var session = NewSession();
+        long now = Environment.TickCount64;
+        // 2.5s Frostbolt currently casting
+        session.PendingNormalCasts.Enqueue(MakeCastTimeCast(116, castTimeMs: 2500, hasStarted: true));
+
+        int marked = session.MarkStartedCastsMovementCancelled(now + 2500);
+
+        Assert.Equal(1, marked);
+        var entry = session.PendingNormalCasts.ToArray()[0];
+        Assert.True(entry.MovementCancelled);
+        Assert.Equal(now + 2500, entry.WatchdogDeadlineMs);
+    }
+
+    [Fact]
+    public void MarkStartedCastsMovementCancelled_Instant_NotMarked()
+    {
+        // Instants don't get cancelled by movement in vanilla.
+        var session = NewSession();
+        session.PendingNormalCasts.Enqueue(MakeCastTimeCast(1953, castTimeMs: 0, hasStarted: true));
+
+        int marked = session.MarkStartedCastsMovementCancelled(Environment.TickCount64 + 2500);
+
+        Assert.Equal(0, marked);
+        Assert.False(session.PendingNormalCasts.ToArray()[0].MovementCancelled);
+    }
+
+    [Fact]
+    public void MarkStartedCastsMovementCancelled_NotStarted_NotMarked()
+    {
+        // Cast hasn't started yet (in-flight to server, awaiting SMSG_SPELL_START).
+        // Movement at this stage doesn't apply because nothing's actually casting.
+        var session = NewSession();
+        session.PendingNormalCasts.Enqueue(MakeCastTimeCast(116, castTimeMs: 2500, hasStarted: false));
+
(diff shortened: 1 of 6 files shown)


jameopotato:
**Review hold — launch week caution**

The peek-instead-of-dequeue approach correctly fixes the stuck-button regression from the DontReport override, but introduces a potential cast-pipeline stall if `SMSG_CAST_FAILED` doesn't follow `SMSG_SPELL_FAILURE` on Kronos.

**Concern:** If a cast-time spell (Frostbolt, Flash of Light, etc.) fails because the target dies mid-cast, and Kronos routes that through `SendInterrupted` without a trailing `SMSG_CAST_FAILED`, the queue entry leaks with `HasStarted=true`. `HasStartedNormalCast()` then blocks ALL subsequent casts — not just same-spell — until the player presses the same spell again to displace the leaked entry.

This isn't an edge case — mobs dying mid-cast is a very common scenario for DPS and healers.

**Before merging, needs in-game verification on Kronos:**
1. Cast Frostbolt at a mob → have someone else kill it mid-cast → immediately try casting another spell
2. Repeat several times to verify no stuck-cast state

If the trailing `SMSG_CAST_FAILED` reliably arrives for target-death on Kronos (i.e., target-death goes through `Spell::cancel()` not `SendInterrupted`), then the trade-off is safe and this can merge. If not, a watchdog timer for the Kronos case is needed first.

jameopotato:
**Alternative approaches to consider**

**Alt B — Peek + watchdog timer:** Keep the peek approach but add a watchdog timer (e.g., 2-3 seconds after SMSG_SPELL_FAILURE). If HandleCastFailed doesn't dequeue the entry within the window, the watchdog force-dequeues it and sends the ack-fail. This covers the Kronos edge case where SMSG_CAST_FAILED never arrives, while preserving the real-reason delivery when it does arrive promptly. Similar pattern to the taxi dismount CTS already in the codebase.

**Sideband investigation:** The JP_ sideband channel (CHAT_MSG_SYSTEM proxy→addon) could potentially help here. If the proxy can detect the "target died mid-cast, no SMSG_CAST_FAILED arriving" state server-side (e.g., by observing SMSG_DESTROY_OBJECT for the target while a cast is pending), it could proactively synthesize the cleanup rather than waiting for a server packet that may never come. Worth exploring whether the proxy has enough state to detect this reliably — the target's GUID is in the pending cast request, and destroy/death packets are already processed.

Issue #43, opened 2026-04-22
Re-write Spellcasting Queue


TakaNoDan:
**Describe the bug**
Rewrite the spell casting flow in the proxy so it no longer waits for a server response before allowing the next cast request.
Right now, the proxy behaves sequentially: cast → wait for result → cast again, which adds full round-trip latency to every failed or queued spell attempt.

The goal is to implement a parallel spell queue system similar to modern WoW, where multiple cast attempts can be buffered and sent independently based on predicted legality (GCD, cast state, latency window). The proxy should intercept outgoing spell packets, queue them, track spell/GCD timing locally, and dispatch them at the earliest valid moment without blocking on previous responses. Server replies should be handled asynchronously and matched back to queued requests.

This should improve responsiveness, macro execution, spam casting, and overall combat feel on a 1.12 server while using a 1.14 client. It's huge.

(SugarProxy already implemented that and the effects in-game are night and day). Link : [https://heitu.org/doc/糖糖代理.html#法术施法队列](https://heitu.org/doc/%E7%B3%96%E7%B3%96%E4%BB%A3%E7%90%86.html#%E6%B3%95%E6%9C%AF%E6%96%BD%E6%B3%95%E9%98%9F%E5%88%97)

<img width="777" height="723" alt="Image" src="https://github.com/user-attachments/assets/1ba90f3a-e0a4-452e-91db-d95d4aace5e6" />

**Game Server**
1.12 / Everlook / Kronos / All

**Hermes Proxy Version**
Hermes / Xian

Mongrul:
Looking into this

Mongrul:
## Update: what was learned, and what is still undecided

Spent some time investigating this. Leaving this issue
open because the broader question — whether to build a real spell queue in
the proxy — is worth community input before we commit.

### What we found

The proxy's current spell-cast path is more nuanced than "wait for the
server before the next cast." It already queues casts and forwards them to
Kronos without blocking. The serial behavior users feel comes from two
specific places:

1. **`HasStartedNormalCast()` proxy-side rejection.** Once Kronos sends
   `SMSG_SPELL_START` for a cast, the proxy rejects any further
   `CMSG_CAST_SPELL` with `SendCastRequestFailed`, never forwarding to
   Kronos. This is correct for mid-cast-bar spam (prevents "Another action
   in progress" on every keypress during a 2s Fireball) but **wrong for
   off-GCD spells** — Sprint, Evasion, trinkets, racials, stance changes
[...]

PR #166, opened 2026-05-07
Fix GCD sweep during cast-time spam: silent-hold pressed spell until …

Mongrul:
## Problem

  Spamming a spell while a cast bar was in progress showed no GCD sweep on the
  action button for the queued press.

  Root cause: the cast-time hold path called `SendCastFailedWithoutPrepare` on
  the queued press immediately on keydown, dropping the modern client's GCD
  anticipation. The button went idle while the hold was parked. When the held
  cast eventually fired on cast-complete, a new SpellPrepare arrived — but the
  visible idle gap made it look like no sweep.

  ## Fix

  Mirror the GCD-hold path's existing silent-hold pattern (issue #43): don't
  ack-fail the new press while it's held. Keep the button lit until SpellPrepare
  arrives at SPELL_START of the refired cast, giving continuous GCD sweep across
  the hold.

  Also closes a cleanup gap found during bug-hunter review: `ClearHeldCastTimeCast`
  now returns the dropped slot so all interrupt paths properly ack-fail any
  silently-held press when the cast bar is cancelled:
  - `HandleCastFailed` — server rejects the cast (LoS, OOM, silenced, etc.)
  - `HandleSpellFailure` — mid-cast interruption (kick, target death)
  - `HandleCancelCast` — player ESC-cancels the cast bar

  ## Testing

  - Spam Flash of Light mid-cast → GCD sweep now shows continuously
  - ESC during a hold → queued button releases cleanly, no stuck-lit button
  - Tested on Kronos PTR (vmangos 1.12.1, bundle jimsproxy-20260507-170855)

### DIFF
--- a/HermesProxy.Tests/World/WatchdogEvictionTests.cs
+++ b/HermesProxy.Tests/World/WatchdogEvictionTests.cs
@@ -240,4 +240,95 @@ public class WatchdogEvictionTests
 
         Assert.False(session.HasStartedNormalCast());
     }
+
+    // ---- Movement preemption tests ----
+
+    private static ClientCastRequest MakeCastTimeCast(uint spellId, uint castTimeMs, bool hasStarted)
+    {
+        return new ClientCastRequest
+        {
+            SpellId = spellId,
+            Timestamp = Environment.TickCount,
+            HasStarted = hasStarted,
+            StartedCastTimeMs = castTimeMs,
+        };
+    }
+
+    [Fact]
+    public void MarkStartedCastsMovementCancelled_StartedCastTime_Marked()
+    {
+        var session = NewSession();
+        long now = Environment.TickCount64;
+        // 2.5s Frostbolt currently casting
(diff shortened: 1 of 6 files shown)


jameopotato:
**Review: Approve with mandatory in-game testing**

Logic is sound — mirrors the established GCD-hold silent pattern. All interrupt paths (HandleCastFailed, HandleSpellFailure, HandleCancelCast) correctly ack-fail the silently-held press. ClearHeldCastTimeCast returning the dropped slot is a clean API improvement.

Held for this beta round (launch week) — needs in-game verification before merge:

1. Spam Flash of Light mid-cast → GCD sweep should stay continuous, no dark gap
2. ESC during a held press → button releases cleanly, no stuck-lit
3. Get kicked mid-cast while holding a press → button releases cleanly
4. OOM mid-cast while holding a press → button releases, error popup shows
5. Normal cast complete with held press → held cast fires seamlessly

Will merge after testing on a dedicated build.

PR #214, opened 2026-05-11
Spell: send SpellPrepare on hold-accept so CastFailed/SpellFailure ca…

Mongrul:
…n match

Extends the off-GCD path's "send SpellPrepare immediately so failures have something to match against" contract to the three silent-hold paths (cast-time hold, GCD hold, forwarded-pending hold).

The off-GCD path comment already documented the underlying bug:

> Without this, if the server rejects the cast (LoS, range), the client has no
> SpellPrepare to match against SpellFailure and the button stays in a pending/lit
> state permanently.

All three silent-hold paths defer SpellPrepare to SPELL_START time, which means any SMSG_SPELL_FAILURE arriving for the eventually-forwarded held cast (LoS, range, cooldown race, drain-clear) lands on a client with no ClientCastID -> ServerCastID mapping. Result: action button stays permanently lit, sound looping, /reload doesn't fix it, only full game restart.

Symptom is most prevalent on low-ping (~50ms) clients in populated zones: the proxy's hold delay (15-78ms observed) plus SMSG backlog from other casters extends the time the client spends in queued-press state past its tolerance window, and crowded zones produce more SpellFailures (LoS blocked by other players, server-side cooldown races).

Also gates the SpellPrepare emit at HandleSpellStart so held casts (which now already sent SpellPrepare at hold-accept time) don't get a duplicate SpellPrepare when SPELL_START arrives. Same pattern the SPELL_GO handler already uses (line 1273). Makes the "exactly one SpellPrepare per ClientCastID" contract explicit and cleans up packet noise.

Mirrors the off-GCD path exactly -- no new behavior, just extends "send SpellPrepare on accept" to the hold-paths.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -922,10 +922,19 @@ public partial class WorldClient
             // SMSG_SPELL_START even for these) apart from real cast-time spells.
             pendingCast.StartedCastTimeMs = spell.Cast.CastTime;
 
-            SpellPrepare prepare = new();
-            prepare.ClientCastID = pendingCast.ClientGUID;
-            prepare.ServerCastID = spell.Cast.CastID;
-            SendPacketToClient(prepare);
+            // Skip SpellPrepare if the hold-accept path (or off-GCD path) already sent
+            // one for this cast. Without this guard, held casts get a duplicate
+            // SpellPrepare (one at hold-accept, one here); the modern client treats
+            // duplicates as idempotent, but suppressing the second is cleaner and makes
+            // the contract explicit.
+            if (!pendingCast.HasSentPrepare)
+            {
+                SpellPrepare prepare = new();
+                prepare.ClientCastID = pendingCast.ClientGUID;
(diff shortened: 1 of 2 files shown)


jameopotato:
**Review: Hold for in-game testing**

Logic is sound — sends SpellPrepare at hold-accept time so the client has a ClientCastID→ServerCastID mapping to match future SMSG_SPELL_FAILURE against. HasSentPrepare flag prevents duplicate at HandleSpellStart. All three hold paths (cast-time, GCD, forwarded-pending) covered.

Held because this modifies all three hold paths — the most sensitive code in the spell system. Needs in-game verification before merge:

1. Spam Frostbolt mid-cast → GCD sweep stays continuous (no flash from early SpellPrepare)
2. Cast heal → move mid-cast while holding next press → button releases cleanly
3. Cast at target behind LoS while spamming → button doesn't stay lit after rejection
4. ESC during held cast → clean release
5. Low-ping rapid spam in populated zone → no permanent stuck-lit buttons

Mongrul:
I tested in game on paladin heal spam, made arcane explosion, frost bolt and hunter

PR #228, opened 2026-05-12
Feature/threat engine synth v2

Mongrul:
## Summary

  Adds a synthesized threat-tracking engine to the proxy so modern 1.14 client
  threat addons (Plater, ThreatPlates, Details, OmniThreat, etc.) work against
  vanilla 1.12 servers that don't emit SMSG_THREAT_UPDATE natively.

  Stack of 9 commits, in order:

  - **6.0 + 6.1** — full-party threat from observed combat log (damage + heal)
  - **Diagnostics** — emit/drop-path JSONL events for addon debugging
  - **6.2** — aggro hysteresis + leave-combat reset to prevent flicker
  - **Talent predecessor injection** — synthesize lower talent ranks the vanilla
    server unlearns via SMSG_UNLEARNED_SPELL so IsPlayerSpell(rankN) works for
    every rank the player has earned (foundation for talent-aware threat math)
  - **6.3** — passive talents: Defiance, Feral Instinct, Silent Resolve
  - **6.4** — school-gated talents: Shadow Affinity, Druid Subtlety, Imp Righteous Fury
  - **6.5** — Imp Power Word: Shield layer + threat.talent_snapshot diagnostic
  - **Talent corrections** — Druid Subtlety moved to universal passive (was
    incorrectly school-gated to Balance damage spells, missing resto-druid heal
    threat reduction); Paladin RF allowlist extended with Judgement / SoR /
    Holy Wrath / Exorcism IDs for completeness
  - **Bug-hunt fixes** — pre-beta audit caught three issues:
    - HoT ticks (Rejuvenation, Regrowth) were generating full-amount threat on
      topped-off targets instead of zero (now uses ComputeOverHealFromCache)
    - Mage Counterspell rank 2 (18469) had no threat module registered
    - Heal threat formula didn't subtract Absorbed (defensive — vanilla wire
      doesn't carry the field, but TC-1.12 forks may back-port it)

  All formulas ported from LibThreatClassic2 (LTC2). Per-class spell allowlists
  sourced from LTC2's `ClassModules/Classic/*.lua`. Talent rank IDs canonical
  per 1.14.2 Talent.dbc (wago.tools build 42597).

  Diagnostics: every threat-relevant event emits a structured JSONL entry
  (`threat.damage_added`, `threat.heal_added`, `threat.emit_update`,
  `threat.passive_modifier`, `threat.talent_snapshot`, `threat.drop_irrelevant_attacker`)
  so testers can attach a JSONL bundle with any threat-related bug report.

  ## Test plan

  - [ ] Local: 487 tests pass (no regressions in solution test suite)
  - [ ] Beta tester — warrior tank, normal threat ordering on heroic strike + sunder pull
  - [ ] Beta tester — feral druid tank, Maul/Swipe threat ratio sane
  - [ ] Beta tester — resto druid healer, Rejuvenation on full-HP tank shows zero threat tick (this fix)
  - [ ] Beta tester — paladin tank attempt, Imp RF rank N shows expected mult on Consecrate/SoR/Judgement
  - [ ] Beta tester — shadow priest, Mind Blast + Mind Flay threat reduced by Shadow Affinity (and further by Silent
  Resolve)
  - [ ] No JSONL log explosion in 40-man raids — `threat.drop_irrelevant_attacker` not pathologically chatty

### DIFF
--- a/HermesProxy.Tests/World/TalentThreatModifierTests.cs
+++ b/HermesProxy.Tests/World/TalentThreatModifierTests.cs
@@ -35,9 +35,12 @@ public class TalentThreatModifierTests
     // Rank-id arrays MUST match the constants in ThreatTracker.cs verbatim.
     // These are the spell IDs the proxy reads off the player's known-spells
     // set; changing them silently is the kind of bug only a tester catches.
-    private static readonly uint[] DefianceRanks      = { 12303, 12788, 12789, 12791, 12792 };
-    private static readonly uint[] FeralInstinctRanks = { 16947, 16948, 16949, 16950, 16951 };
-    private static readonly uint[] SilentResolveRanks = { 14523, 14784, 14785, 14786, 14787 };
+    private static readonly uint[] DefianceRanks         = { 12303, 12788, 12789, 12791, 12792 };
+    private static readonly uint[] FeralInstinctRanks    = { 16947, 16948, 16949, 16950, 16951 };
+    private static readonly uint[] SilentResolveRanks    = { 14523, 14784, 14785, 14786, 14787 };
+    private static readonly uint[] ShadowAffinityRanks   = { 15272, 15318, 15320 };
+    private static readonly uint[] DruidSubtletyRanks    = { 17118, 17119, 17120, 17121, 17122 };
+    private static readonly uint[] ImpRighteousFuryRanks = { 20468, 20469, 20470 };
 
     // === Rank detection: highest known rank wins ===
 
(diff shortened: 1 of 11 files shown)


jameopotato:
## Review — Approved with required changes

Overall: well-engineered, well-tested, clean merge against current beta. All formulas match LibThreatClassic2, comprehensive JSONL diagnostics, 345 lines of test coverage pinning every talent formula and rank ID. Ready for beta with the items below.

### Required before merge

**1. Add `_passiveCache.Clear()` to `Reset()` (one-line fix)**

`Reset()` clears `_threatLists`, `_lastHighest`, `_dirty` but not `_passiveCache`. Stale GUIDs from previous sessions won't cause incorrect threat (they won't match new session GUIDs), but they leak memory across character switches. Add to line ~486:
```csharp
public void Reset()
{
    _threatLists.Clear();
    _lastHighest.Clear();
    _dirty.Clear();
    _passiveCache.Clear();  // ← add this
}
```

**2. Toggleable via config setting**

The threat engine runs on every damage/heal/energize event for the local player + all group members. In a 40-man raid, that's substantial CPU work (per-event talent lookups, gear set counts, aura scans, hysteresis computation, SMSG emission). Players who don't use threat meters shouldn't pay this cost, and players who experience issues should be able to disable it without swapping proxy versions.

Add a `ThreatEngine` boolean to `Settings.cs` (default `true`, same pattern as `LowLatencyMode`), and gate the `OnDamage`/`OnHeal`/`OnEnergize`/`OnSpellCast` entry points:
```csharp
public void OnDamage(WowGuid128 attacker, WowGuid128 victim, int spellId, double rawDamage)
{
    if (!Settings.ThreatEngine) return;
    // ... existing code
}
```
Eight call sites in SpellHandler/CombatHandler, one early-return check each. Minimal effort, maximum flexibility.

### For Mirasu

**3. Netherwind set item ID duplicate**

Line 48 of `ThreatSetBonuses.cs` lists `16818` twice in the Netherwind array. Should be 8 unique item entry IDs for the 8-piece set. Please verify against wowhead/wago and fix the duplicate — it would cause `CountEquippedSetPieces` to undercount by 1, potentially missing the 3-piece or 8-piece threshold.

**4. `CountEquippedSetPieces` caching (future pass)**

Currently creates a `new Dictionary<int, int>()` per damage/heal event (~20 allocs/sec in a raid). Not blocking, but consider caching the result and invalidating on equipment-field changes in a follow-up. The code comment already notes this.

### Interaction with recent merges (all clean)

Evaluated every PR merged since #228 was created (May 12):

- **#290 async structured logging**: Beneficial — threat diagnostic events won't block the packet thread
- **#302 Cannibalize animation**: Suppresses SPELL_START for 20577, threat reads SPELL_DAMAGE/HEAL events — no interaction
- **#279 Quest throttle**: Entirely separate system (quest query caching) — no interaction
- **#297 Aura stack decrement** (held): Changes how aura updates emit to the client, but threat reads raw cached UNIT_FIELD_AURA slots directly — no interaction
- **#293 Gather nodes** (held): Changes DrainPendingCastsForDestroyedTarget (cast pipeline), while threat's `OnUnitDestroyed` handles threat list cleanup independently — no interaction
- **#299 Reticle cast** (held): CMSG outbound path, threat reads SMSG inbound combat log — no interaction

No problematic interactions found.

Mongrul:
Review feedback addressed in 5b1926f.

**Finding 1 — `_passiveCache.Clear()` in `Reset()`:** added.

**Finding 2 — config toggle:** added `Settings.ThreatEngine` (default `true`, mirrors `LowLatencyMode` pattern + `<add key="ThreatEngine" value="true" />` in `HermesProxy.config`). I diverged from the reviewer's suggested shape on where to place the gate:

- Placed the early-return **inside** `ThreatTracker` at the top of each per-event entry point (`OnDamage`, `OnHeal`, `OnEnergize`, `OnSpellCast`, `OnPowerWordShield`) rather than at each of the call sites in `SpellHandler`/`CombatHandler`.
- Single source of truth — can't drift if a new call site lands later.
- The talent synthesizer in `SpellHandler.cs` (`SynthesizedTalentRanks` reconcile path) deliberately runs **unconditionally** regardless of `ThreatEngine`; placing the gate at the call sites would have made it easy to mis-edit and accidentally also disable the synthesizer. Other systems consume that data even when threat output is off.
- Cleanup methods (`Reset`, `OnUnitDestroyed`, `OnLocalPlayerLeftCombat`, `ClearMob`) are intentionally **not** gated — they no-op on empty state and need to run regardless so a runtime toggle doesn't leak across sessions.
- Field default-initialized at declaration (`public static bool ThreatEngine = true`) so tests / paths that bypass `LoadAndVerifyFrom` still get the safe default.

**Finding 3 — Netherwind duplicate:** fixed. While verifying I also found the same class of bug in two more set arrays — same root cause (wrong item entry IDs miscounting equipped pieces and missing bonus thresholds):

| Set | Was | Fix | Wrong entries |
|---|---|---|---|
| Mage Netherwind (line 45) | `{16914, 16917, 16916, 16818, 16915, 16818, 16819, 16920}` | `{16818, 16912, 16913, 16914, 16915, 16916, 16917, 16918}` | 16818 dup; 16819 = Vambraces of Prophecy (Priest T2); 16920 = Handguards of Transcendence (Priest T2). Missing Boots/Gloves/Bindings. |
| Warrior Battlegear of Might (line 40) | `{16866, 16867, 16868, 16869, 16860, 16861, 16862, 16863}` | `{16861, 16862, 16863, 16864, 16865, 16866, 16867, 16868}` | 16860 = Lawbringer Gauntlets (Paladin T1!); 16869 = The Skull of Scryer (trinket). Missing Belt of Might (16864) + Breastplate of Might (16865). |
| Priest Vestments of Faith (line 65) | `…, 23064` | `…, 23061` | 23064 = Ring of the Dreamwalker (Druid T3); should be Ring of Faith (23061). |

All cross-checked against `HermesProxy/CSV/ItemSparse1.csv`.

**Finding 4 — `CountEquippedSetPieces` per-event Dict allocation:** deferred per the reviewer's "not blocking". The existing code comment already notes it; can land as a follow-up.

**Build / tests:** `dotnet build HermesProxy.sln` clean (1 pre-existing warning in `TalentThreatModifierTests.cs:276` — `irfRank` unused, not from these changes). `dotnet test HermesProxy.Tests` → **502/502 pass**, including all 50 threat tests.

PR #238, opened 2026-05-13
Vendor sold-out grey: addon-side desaturate + proxy refresh + diagnos…

Mongrul:
…tics

Tester mammouth reports sold-out vendor items not visually greyed on 1.14 Classic client. Confirmed parity break: 1.12 native client greys these reliably; modern 1.14 Classic doesn't, even though buy-block fires correctly server-side.

Tester JSONL (20260513-044049, 20260513-044704) walks through the investigation. Liferoot (item 3357) at Maria Lumere ships from vmangos with Quantity=0 in SMSG_LIST_INVENTORY both at vendor open AND on every buy-failed refresh. Two wire-side experiments to flip the grey state both confirmed delivered but ignored by the modern client:
  - Quantity=0 alone — no grey.
  - VendorItem.PlayerConditionFailed=1 on sold-out slots — no grey.

The 1.14 Classic Era client's icon-grey gate for vendor slots lives somewhere we can't reach from the wire on this build. But GetMerchantItemInfo's numAvailable return DOES expose the correct value (0 = sold out) in Lua-land, so we can apply the desaturate ourselves.

Changes:

1. JimsPlus TooltipFix.lua — FixMerchantUsability now reads numAvailable alongside the existing usability check. On sold-out slots: - icon texture SetDesaturated(true) - icon/name/border tint 0.65 grey (chosen by user; lighter than the 0.45 first pass, matches the 1.12 native client's dim) - name text 0.65 grey, overriding the quality-color path Non-sold-out slots restore desaturate=false so a button reused for a different slot doesn't carry stale grey. Core.lua build bumps 18 -> 19.

2. ItemHandler.cs HandleBuyFailed/HandleBuySucceeded — proxy-issue CMSG_LIST_INVENTORY when: - BuyFailed reason in {ItemSoldOut, ItemAlreadySold, CantFindItem} (Twinstar sends ItemAlreadySold=1, not the canonical ItemSoldOut=7, so we widen the gate) - BuySucceeded.NewQuantity == 0 The fresh SMSG_VENDOR_INVENTORY triggers MerchantFrame_Update on the modern client, which fires the addon's grey logic with up-to-date numAvailable. Without this, a player buying the last stack or hitting a slot another player drained mid-session wouldn't see the grey update until they closed and reopened the vendor.

3. NPCHandler.cs HandleVendorInventory — diagnostic Log.Event per VendorItem (slot, item_id, quantity, stack_count, price, extended_cost) and a sold_out_count summary. Confirms wire state for any future "vendor behaves wrong" report without re-instrumenting.

All 502 tests pass.

### DIFF
--- a/Addons/JimsPlus/Core.lua
+++ b/Addons/JimsPlus/Core.lua
@@ -1,6 +1,6 @@
 local ADDON_NAME, namespace = ...
 
-namespace.BUILD = 18
+namespace.BUILD = 19
 
 print("|cFF00FF00[JimsPlus]|r Core loaded (build " .. namespace.BUILD .. ")")
 
--- a/Addons/JimsPlus/TooltipFix.lua
+++ b/Addons/JimsPlus/TooltipFix.lua
@@ -557,6 +557,22 @@ local function FixMerchantUsability()
         local index = (page - 1) * perPage + i
         local button = _G["MerchantItem" .. i .. "ItemButton"]
         if button and index <= total then
+            -- Sold-out detection: GetMerchantItemInfo's 5th return is numAvailable
+            -- (0 = sold out, -1 = unlimited, N = stock left). The 1.14 Classic
+            -- Era client ignores the wire's VendorItem.Quantity=0 for icon-grey
+            -- rendering, so we apply the grey here. 1.12 native client greys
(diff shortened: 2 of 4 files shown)


Mongrul:
conflict fixed

PR #260, opened 2026-05-18
Fix/trainer buy predecessor defense

Mongrul:
## Summary

 **Unlearned spell defense**: legacy server's trainer-buy supersede path calls `RemoveSpell(prev)` unconditionally but gates the
  `SMSG_SUPERCEDED_SPELL` send on `m_session->IsInWorld()`. When that gate is briefly false (server-side transient), the
   predecessor disappears from the spellbook silently. The next `CMSG_CAST_SPELL` for the predecessor rank slips past
  the existing cast-block-unknown-spells guard and trips an autoban. Repro'd 2026-05-18 on a rogue training Stealth R3 —
   server log: `HandleCastSpellOpcode: Player ... casts spell 1785 which he shouldn't have`. Proxy now mirrors the
  server-side removal locally so the cast-block guard catches the post-buy cast attempt and converts it into a
  `CastFailed` instead of a real ban.
  - **Trainer list refreshes after a buy**: vanilla server sends `SMSG_TRAINER_BUY_SUCCEEDED` (1.12-only opcode the
  modern client has no slot for), then nothing else. The modern client never refreshes its trainer panel from
  `SMSG_LEARNED_SPELL` alone. Proxy now intercepts the legacy opcode and emits a fresh `CMSG_TRAINER_LIST` back to the
  legacy server (same pattern AuctionHandler uses post-operation).
  - **Filter + intercept already-known buys**: empirically the modern Classic 1.14 trainer panel does NOT visually
  differentiate Known from Available via the state byte alone, and does NOT honor mid-interaction `SMSG_TRAINER_LIST`
  refreshes (panel renders from the initial open). Two complementary fixes — initial list filters Known entries before
  forwarding; the buy handler intercepts clicks on already-known entries (direct, supersede-known, and rapid
  double-click races) and force-closes the gossip panel so reopening produces a clean list.

  ## Changes

  | File | What |
  |---|---|
  | `HermesProxy/CSV/SpellRankChain.csv` (new) | 3140 `spell_id,prev_spell` pairs derived from 1.12.1 Spell.dbc
  Name+Rank grouping |
  | `HermesProxy/World/GameData.cs` | `SpellRankPredecessor` FrozenDictionary + `LoadSpellRankChain` loader |
  | `HermesProxy/GlobalSessionData.cs` | `PendingTrainerBuy*` (ban defense state), `LastTrainerListKnownSpells`
  (supersede-known cache), `InFlightTrainerBuy*` (duplicate-detection state) |
  | `HermesProxy/World/Server/PacketHandlers/NPCHandler.cs` | `HandleTrainerBuySpell`: stale-click/duplicate-click
  intercept + speculative predecessor removal |
  | `HermesProxy/World/Client/PacketHandlers/NPCHandler.cs` | `HandleTrainerList`: filter Known + populate Known cache;
  `HandleTrainerBuySucceeded`: re-request fresh list; `HandleTrainerBuyFailed`: restore predecessor on explicit fail |
  | `HermesProxy/World/Client/PacketHandlers/SpellHandler.cs` | clear ban-defense + in-flight state on
  `SMSG_LEARNED_SPELL` / `SMSG_SUPERCEDED_SPELLS` |
  | `HermesProxy/World/KnownBenignOpcodes.cs` | remove `SMSG_TRAINER_BUY_SUCCEEDED` (now has a real handler) |

  ## Diagnostics added (all low-volume)

  - `spell.trainer_buy.predecessor_speculatively_removed` — ban defense fired
  - `spell.trainer_buy.predecessor_restored_on_failed` — ban defense reverted on explicit fail
  - `spell.trainer_buy.refresh_list_requested` — fresh trainer list re-requested
  - `spell.trainer_list.known_filtered` — count of hidden Known spells per list
  - `spell.trainer_buy.intercepted_already_known` — stale/duplicate click caught with `in_known_spells` /
  `in_trainer_known_set` / `is_in_flight_duplicate` flags

  ## Test plan

  - [x] `dotnet build` clean
  - [x] `dotnet test` 452/452 pass (talent-rank-injection tests still green)
  - [x] In-game: trainer flow with all four scenarios validated 2026-05-18
    - [x] Normal multi-spell training (60+ buys in one session) — no false intercepts
    - [x] Stealth chain train (R1→R2→R3→R4) — ban defense fires per rank-up, predecessor speculatively removed,
  server-side state matched
    - [x] Apprentice → Journeyman riding train, then click stale Apprentice — intercept fires with
  `in_trainer_known_set: true`, panel closes
    - [x] Rapid double-click same spell — intercept fires with `is_in_flight_duplicate: true`
  - [x] No regression: existing `cast-block-unknown-spells` guard still catches removed predecessors

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -228,6 +228,27 @@ public sealed class GameSessionData
     // a real ban. On explicit SMSG_TRAINER_BUY_FAILED we restore the predecessor.
     public uint PendingTrainerBuySpellId;
     public uint PendingTrainerBuyRemovedPredecessor;
+    // JimsProxy: real spell ids that the most recent SMSG_TRAINER_LIST marked
+    // with TrainerSpellState=Known. Authoritative server view of "this spell
+    // is effectively learned by you" — covers both direct ownership (HasSpell)
+    // and supersede-chain ownership (HasSpell on any higher rank). Without this
+    // we can't intercept stale-click buys for spells that were just removed
+    // from CurrentPlayerKnownSpells by a SMSG_SUPERCEDED_SPELLS for the new
+    // rank: e.g., learn Apprentice Riding, then Journeyman, then click stale
+    // Apprentice in the UI — Apprentice is gone from KnownSpells but the
+    // server still treats it as effectively-known and rejects the buy with
+    // FAILED. Repopulated on every new SMSG_TRAINER_LIST so this set always
+    // reflects the latest server view.
+    public System.Collections.Generic.HashSet<uint> LastTrainerListKnownSpells = new();
+    // JimsProxy: in-flight trainer-buy tracking. Set on every forwarded
+    // CMSG_TRAINER_BUY_SPELL, cleared on the response (LEARNED, SUPERCEDED, FAILED,
+    // or a fresh trainer-list refresh). Used to drop rapid same-spell double-clicks
+    // where the second CMSG races with the first's response — both reach the
+    // server, the first succeeds, the second gets FAILED. Distinct from
+    // PendingTrainerBuy* (which only tracks buys that triggered speculative
+    // predecessor removal for the ban defense).
+    public uint InFlightTrainerBuySpellId;
+    public long InFlightTrainerBuyTickMs;
     // JimsProxy: per-unit HP cache used to compute overhealing on legacy servers
     // that don't include OverHeal in SMSG_SPELL_HEAL_LOG (1.12 vanilla). Authoritative
     // source is UNIT_FIELD_HEALTH / UNIT_FIELD_MAXHEALTH from SMSG_UPDATE_OBJECT;
--- a/HermesProxy/World/Client/PacketHandlers/NPCHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/NPCHandler.cs
@@ -168,6 +168,14 @@ public partial class WorldClient
         trainer.TrainerID = trainer.TrainerGUID.GetEntry();
         trainer.TrainerType = packet.ReadInt32();
         int count = packet.ReadInt32();
+        int filteredKnown = 0;
+        // JimsProxy: reset the authoritative-Known set for this fresh trainer
+        // list, then repopulate as we walk the entries. The intercept in
+        // HandleTrainerBuySpell consults this set to catch stale UI clicks for
+        // spells that the server considers Known via supersede chain (and so
+        // are absent from CurrentPlayerKnownSpells).
+        var lastKnownSet = GetSession().GameState.LastTrainerListKnownSpells;
+        lastKnownSet.Clear();
         for (int i = 0; i < count; ++i)
         {
             TrainerListSpell spell = new();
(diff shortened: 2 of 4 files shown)


jameopotato (commented):
Good work on the ban defense mechanism — the speculative predecessor removal + restoration on fail is solid, and the server-log evidence makes the case clearly. A few things to address before this merges:

## Request changes

### Known spell filter removes information (UX regression)

The `continue` in `HandleTrainerList` for `TrainerSpellState.Known` entries strips them entirely from the list sent to the modern client. In 1.12 native, Known spells render greyed out in the trainer panel — players use this to see what they've already learned when planning a training session. Hiding them entirely is a behavior change, not a bug fix. 

Suggestion: keep Known entries in the list with their state byte intact. If the 1.14 client truly can't render them differently from Available, we can add an addon-side visual hook (similar to the vendor sold-out grey approach) rather than stripping data at the packet level.

### GossipComplete may not close the trainer frame

The stale-click intercept sends `GossipComplete` to close the panel. But trainers use the trainer frame, not the gossip frame — does `GossipComplete` reliably close a trainer window on the 1.14 client? If not, the player sees "You already know that spell" but the trainer panel stays open with stale entries, which is worse UX than letting the failed buy round-trip through. Please verify this in-game.

## Suggestions (non-blocking, can be follow-ups)

### Comment verbosity

The Kronos IsInWorld race condition is explained in full (10-15 lines) in four separate locations — `GlobalSessionData.cs`, both `NPCHandler.cs` files, and `SpellHandler.cs`. One canonical explanation + short references elsewhere would keep the code navigable. Not blocking, but a trim pass would help readability.

### 2500ms in-flight duplicate window

The `2500` magic number has no named constant or rationale comment. On high-latency connections a legitimate retry could be intercepted. Consider tying to observed RTT or at minimum extracting to a named constant with documented reasoning.

### SpellRankChain.csv generation

3140 entries with no documented extraction script. If anyone needs to verify or regenerate this data, they'd have to reverse-engineer the derivation. The extraction script (or at least the methodology) should live in `scripts/` alongside other CSV generators.

Mongrul:
The known spell is preserved, you need to filter for it like in 1.12.  The gossip interact flows smoothly you can't even see it, it instantly refreshes the trainer list getting rid of the spell already known. Under spam click training on the PTR (tested roughly 10 time on 10 chars) I have ran into zero regressions or issues. 

image

Mongrul:
Re: both concerns —

**GossipComplete + trainer frame**: confirmed in-game, GossipComplete doesn't close the trainer panel on 1.14, but the trainer list refreshes and the spell that errored moves into the **Trained** category. Net UX: the panel stays open (so the player keeps shopping), no stuck error, and the player can still see what they've learned via the Trained filter. The code comment is slightly misleading on the mechanism — happy to clean it up in a follow-up.

**Known-spell filter**: tied to the above — Known entries don't disappear from the player's view, they show up in the Trained category after the refresh. So the information isn't being hidden, just routed to the right bucket. Tested ~10x across 10 chars on PTR with spam-clicking trainer entries, zero regressions.

jameopotato (approved):
UI hygiene layer on top of the already-merged ban defense. Purely additive (+140, -0).

Known-spell filter: correctly strips server-marked Known entries before forwarding to the 1.14 client, which renders them indistinguishable from Available. Set repopulated on every fresh trainer list, real spell IDs normalized via GetRealSpell.

Stale-click intercept: three predicates (direct known, trainer-known-set, in-flight duplicate) cover all common stale-click scenarios. Gated on ExpansionVersion <= 1. In-flight state cleared on all three response paths (LEARNED, SUPERCEDED, FAILED). No false-positive path found.

Field-tested: chain training, stale clicks, rapid double-clicks, 60+ buy sessions.

PR #279, opened 2026-05-20
Quest query throttle: cache + dedupe + token-bucket

Mongrul:
Questie's filter-toggle SmoothReset bursts ~1500 unique CMSG_QUERY_QUEST_INFO in under a second when re-evaluating quest tags across its DB. Kronos's anti-flood (~80-100 packets/sec) drops the connection mid-burst. Field test captured 1427 unique CMSGs in ~700ms (>2000/sec) before SocketException.

Triages each incoming CMSG_QUERY_QUEST_INFO before forwarding:
- Cache hit (GameData.QuestTemplates already has it): synthesize the modern SMSG locally from the cached QuestTemplate. QueryQuestInfoResponse.Write() already serializes from a QuestTemplate so no new wire path is needed.
- Negative cache (legacy server returned masked-entry for this id earlier): reply Allow=false locally. Covers TBC/Wrath quest ids that Questie's DB includes but the 1.12 server has never heard of.
- In-flight dedupe: a CMSG for the same id is already pending. The modern client's quest cache is keyed by id, so one SMSG satisfies all waiters.
- Otherwise gate through a per-session token bucket: 10/sec sustained, 20-token initial burst. Excess goes onto a FIFO queue drained by a single async pump (CAS-gated, restarts on tail enqueues). The drain re-checks the cache on dequeue so any id that got warmed by a parallel proxy-issued query skips the redundant server round-trip.

Response handler clears the in-flight bit and seeds the negative cache on masked-entry replies, so the gating self-heals across the session.

JSONL events for tuning: quest.query.cache_hit / negative_cache_hit / in_flight_drop / bucket_enqueue / bucket_drain_send / bucket_drain_cache_hit.

Field-tested: connection survives a full Questie PvP-icon toggle burst.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -245,6 +245,28 @@ public sealed class GameSessionData
     public Dictionary<uint, WowGuid128> CachedPetNumbers = new();
     // Tracks quest ids the proxy has issued its own CMSG_QUERY_QUEST_INFO for.
     public HashSet<uint> ProxyIssuedQuestInfoQueries = new();
+    // JimsProxy: client-originated CMSG_QUERY_QUEST_INFO gating. Questie's filter-toggle
+    // SmoothReset iterates ~10k quest IDs and fires GetQuestTagInfo on each, generating
+    // thousands of CMSG_QUERY_QUEST_INFO in a single burst. Without dedupe, Kronos's
+    // anti-flood drops the connection. InFlight tracks queries already forwarded so
+    // duplicates within the round-trip window are dropped (one modern SMSG response
+    // satisfies all client-side waiters for the same quest). Negative cache tracks
+    // quest IDs the legacy server returned masked-entry on — typically TBC/Wrath
+    // quests in Questie's DB that the 1.12 server has never heard of.
+    public HashSet<uint> InFlightClientQuestInfoQueries = new();
+    public HashSet<uint> NegativeQuestInfoCache = new();
+    // JimsProxy: token-bucket rate limiter for CMSG_QUERY_QUEST_INFO. Questie's
+    // cold-start scan bursts ~1500 UNIQUE quest IDs in ~700ms (so in-flight dedupe
+    // does nothing on the first pass); this lands at >2000/sec, well past Kronos's
+    // ~80-100/sec WorldPacketLimit, and the connection gets dropped. The bucket
+    // smooths the burst to a sustained 10/sec with a 20-token initial allowance.
+    // PendingQuestQueryQueue holds IDs awaiting a token; QuestQueryDrainerRunning
+    // is a 0/1 CAS flag so at most one async drainer pumps the queue per session.
+    public System.Collections.Concurrent.ConcurrentQueue<uint> PendingQuestQueryQueue = new();
+    public double QuestQueryTokens = 20.0;
+    public long QuestQueryLastRefillMs = 0;
+    public int QuestQueryDrainerRunning = 0;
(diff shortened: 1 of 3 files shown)


jameopotato (changes requested):
Good design — the layered cache → negative-cache → dedupe → token-bucket approach is the right way to solve the Questie disconnect. Holding on one concurrency issue before merge.

### Thread-safety of HashSet fields (blocking)

`InFlightClientQuestInfoQueries` and `NegativeQuestInfoCache` are plain `HashSet<uint>` but are accessed from three threads concurrently:
- **WorldServer handler thread** — `Add`, `Contains` (Server/QueryHandler.cs)
- **WorldClient handler thread** — `Remove`, `Add` (Client/QueryHandler.cs)  
- **Drainer thread pool thread** — `Remove` (DrainQuestQueryBucketAsync)

This codebase explicitly avoids plain collections for cross-thread access — `QuestItemObjectiveProgress` has a comment documenting why it uses `ConcurrentDictionary` ("Plain Dictionary risks torn-state corruption on cross-thread enumeration"), and `UnitHealthCache`, `CachedPetCreatureFamily`, `UnitChannelSpells` all follow the same convention.

Under the exact high-concurrency scenario this PR handles (1500 queries + responses streaming back), concurrent `Add`/`Remove` on `HashSet` will corrupt internal bucket arrays.

**Fix**: Swap both to `ConcurrentDictionary<uint, byte>` (`TryAdd`/`TryRemove`/`ContainsKey`), matching the existing codebase pattern.

### Minor items

- **No exception handling in drainer**: Existing `Task.Run` patterns in this codebase (CharacterHandler, MovementHandler) wrap their bodies in try/catch with logging. The drainer should do the same — if the session disconnects mid-drain, `SendPacketToServer` throws and the exception goes unobserved on the discarded task.

- **Double GameData lookup in drain cache-hit path**: `GameData.GetQuestTemplate(questId)` is called twice — store in a local variable.

Mongrul:
- fix/quest-query-throttle → pushed commit 6cd631e with the three review fixes (ConcurrentDictionary swap, drainer
  try/catch, dedup'd GameData lookup). GitHub will refresh PR #279.
  - local/all-prs-stacked → cherry-picked the same fix as a28fb27, on top of the existing stack. Build clean, 502/502
  tests pass.
  - upstream/beta is ahead of expectations: one of the trainer prep commits (7b8d903 — SpellRankChain CSV+loader) was
  already merged upstream, which is what caused the rebase conflicts I aborted. Worth a closer look next session to see
  if any other stacked PRs landed and the stack needs pruning.

Mongrul:
@
Review feedback addressed in `6cd631e`:

- **Thread-safety (blocking):** `InFlightClientQuestInfoQueries` and `NegativeQuestInfoCache` are now `ConcurrentDictionary<uint, byte>` (`TryAdd` / `TryRemove` / `ContainsKey`), matching the `QuestItemObjectiveProgress` / `UnitHealthCache` convention. No more torn bucket arrays under concurrent Add/Remove.
- **Drainer exception handling:** `DrainQuestQueryBucketAsync` body is wrapped in try/catch — a mid-drain `SendPacketToServer` throw on a disconnected session now logs `quest.query.bucket_drain_error` instead of becoming an unobserved task exception. The `finally` still resets the drainer flag and re-checks the queue race.
- **Double GameData lookup:** the drainer cache-hit path hoists `GameData.GetQuestTemplate(questId)` into a `warmTemplate` local and reuses it.

Build clean, all tests pass. Ready for re-review.
@

PR #280, opened 2026-05-20
Allow shop bought mounts to be useable on all races

KianFitz:
Kronos has shop bought mounts that should be usable on all races.

On the 1.12 client, this information is brought down from SMSG_ITEM_QUERY_SINGLE_RESPONSE and stored in the WDB cache. On the 1.14 client, this information is stored in the ItemSparse CSV.

This change simply allows the mounts from the store to be usable on both factions.


jameopotato (approved):
Verified the 16 changed items are an exact 1:1 match with the Kronos shop mount catalog. Every shop mount is covered, no extras included.

The fix is correct: the 1.14 client uses its built-in ItemSparse CSV for AllowableRace instead of honoring the server's SMSG_ITEM_QUERY_SINGLE_RESPONSE, so shop-bought cross-faction mounts get blocked client-side. Changing AllowableRace to -1 in the CSV removes the client-side restriction while the server still controls actual mount usage via spell validation — no exploit risk.

Mechanically clean diff: only the AllowableRace field is modified on each line, all other fields are byte-identical.

PR #282, opened 2026-05-20
Cleanup: extract mana gem skip condition + reorder exclusion check

### LINKED ISSUE #277: Cleanup: extract mana gem skip condition to named method
In #270, the mana gem slot-mismatch skip is a 4-conjunct inline condition:

```csharp
if (ManaGemItemEntries.Contains(item.Entry) &&
    item.TriggeredSpellIds[slot] > 0 &&
    (effect == null || effect.SpellID != item.TriggeredSpellIds[slot]) &&
    FindItemEffectBySpellId(item.Entry, item.TriggeredSpellIds[slot], slot) == null)
    return null;
```

This has been patched 3 times now (original exclusion → hotfix item IDs → narrowed condition). Extract to a named method like `ShouldSkipManaGemMutation(item, slot, effect)` for readability.

Also: `GetItemEffectByItemId` was moved before the blanket exclusion check, so spellstone/orb items now do a wasted linear scan of ItemEffectStore before hitting `return null`. Move the blanket check back before the lookup, or pass `effect` as a lazy parameter.

Mongrul:
Closes #277.

## Changes

- **`ShouldSkipManaGemMutation(item, slot, effect)`**: replaces the 4-conjunct inline condition with a named method. Each guard is its own early-return so the failure mode is obvious next time this gets patched. The rationale comment now lives with the method instead of inline at the call site.

- **Move blanket exclusion check above `GetItemEffectByItemId`**: spellstone / Orb of Soran'ruk entries now skip the `ItemEffectStore` scan entirely instead of doing a wasted linear search before returning null.

## Verification

Behavior identical — full test suite 452/452 passes on this branch.

### DIFF
--- a/HermesProxy/World/GameData.cs
+++ b/HermesProxy/World/GameData.cs
@@ -4105,23 +4105,34 @@ public static partial class GameData
         return messages;
     }
 
+    // Mana gems: skip mutation only when the server's slot SpellID doesn't match
+    // any CSV record we have — otherwise the rewrite/new-record path binds the
+    // legacy SpellID (10053/10054) and the modern client strips the Use: line.
+    private static bool ShouldSkipManaGemMutation(ItemTemplate item, byte slot, ItemEffect? effect)
+    {
+        if (!ManaGemItemEntries.Contains(item.Entry))
+            return false;
+        int serverSpellId = item.TriggeredSpellIds[slot];
+        if (serverSpellId <= 0)
+            return false;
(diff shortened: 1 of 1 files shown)


jameopotato (approved):
Extracted method is logically equivalent (traced all four conditions). Reorder of exclusion check is safe (no side effects). Rebased onto beta to resolve GameData.cs conflict with #284 — trimmed the method comment to match.

PR #283, opened 2026-05-20
Cleanup: drop redundant M2NativeRatio re-index

### LINKED ISSUE #276: Fix: TryGetValue + re-index in pet scale M2NativeRatio lookup
In #261, `M2NativeRatio.TryGetValue(modelIdForRatio, out var mRatio)` captures the value in `mRatio`, but the multiplication uses `M2NativeRatio[modelIdForRatio]` instead — a redundant dictionary lookup. This pattern is duplicated in both:

- `UpdateHandler.cs` (isLocalPet branch, ~line 4363)
- `QueryHandler.cs` (ResolvePendingPetScalesForEntry, ~line 471)

Fix: use `mRatio` directly instead of re-indexing.

Also: add a null-check on `GetDisplayInfo((uint)displayId)` — if a DisplayID has no entry in the display info table, `.ModelId` will NRE. The `modelIdForRatio > 0` guard catches 0 but not a missing entry.

Mongrul:
Closes #276 (partially — see note on the second item).

## Change

Both pet-scale call sites used `TryGetValue` to capture `mRatio` but then re-indexed `M2NativeRatio[modelIdForRatio]` in the multiplication. Now they use the captured `mRatio` directly.

`mRatio` is pre-declared with a `1f` default because the `TryGetValue` lives inside a short-circuiting `&&` chain — without a definite-assignment fallback the compiler rejects the use in the ternary. The default value is unreachable in practice (the ternary branch only fires when `modelRatioApplied` is true, which requires `TryGetValue` to have succeeded), but the explicit `1f` is also the identity multiplier so any future restructure stays safe.

Two files touched:
- `UpdateHandler.cs` — isLocalPet branch
- `QueryHandler.cs` — `ResolvePendingPetScalesForEntry`

## Note on the second item in #276

The issue also mentions a potential NRE on `GetDisplayInfo((uint)displayId).ModelId` if the DisplayID has no entry. After checking `GetDisplayInfo` (GameData.cs:547), this isn't reachable:

\`\`\`csharp
public static CreatureDisplayInfo GetDisplayInfo(uint displayId)
{
    if (CreatureDisplayInfos.TryGetValue(displayId, out var info))
        return info;
    return new CreatureDisplayInfo(0, 1.0f);
}
\`\`\`

It returns a non-null fallback record with `ModelId = 0`, and the existing `modelIdForRatio > 0` guard catches the missing-entry case. No defensive null-check needed.

## Verification

Behavior identical — full test suite 452/452 passes.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/QueryHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/QueryHandler.cs
@@ -464,11 +464,12 @@ public partial class WorldClient
             int modelIdForRatio = p.DisplayId > 0
                 ? (int)GameData.GetDisplayInfo(p.DisplayId).ModelId
                 : 0;
+            float mRatio = 1f;
             bool modelRatioApplied = modelIdForRatio > 0
                                      && vanillaCmsK.HasValue
-                                     && WorldClient.M2NativeRatio.TryGetValue(modelIdForRatio, out var mRatio);
+                                     && WorldClient.M2NativeRatio.TryGetValue(modelIdForRatio, out mRatio);
             float k = modelRatioApplied
-                ? vanillaCmsK!.Value * WorldClient.M2NativeRatio[modelIdForRatio]
+                ? vanillaCmsK!.Value * mRatio
                 : (vanillaCmsK ?? familyTableK ?? (p.IsWarlockPet ? 0.75f : 1.5f));
 
             float emit = (p.Cms > 0)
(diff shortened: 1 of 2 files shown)


jameopotato (approved):
Eliminates redundant FrozenDictionary re-index at two pet-scale sites — uses the out variable from TryGetValue instead. Identical logic, cleaner code.

PR #293, opened 2026-05-22
Fix gather nodes stuck after the first tick

Mongrul:
## Problem

Mining/herb nodes "do nothing" after the first successful gather — the player has to run out of range and back before the node works again.

## Cause

Gathering a node casts the gather spell at the node GameObject. Clicking again while the first tick is still casting queues a second cast. When the first tick completes the depleted node despawns and the legacy server sends `SMSG_DESTROY_OBJECT` for it. The destroy-hook fast path (`DrainPendingCastsForDestroyedTarget`) then evicts the queued second cast — which the legacy server had **already started** (`SMSG_SPELL_START` sent, client showing a channel bar) — and sends the client `SMSG_CAST_FAILED`.

`SMSG_CAST_FAILED` cannot tear down an already-started cast/channel bar; only `SMSG_SPELL_FAILURE` can. The client is left with a phantom gather channel and refuses to gather again until `SMSG_DESTROY_OBJECT` for the node resets its state client-side — i.e. the player runs away and back.

The destroy-hook fast path exists to fast-fail casts that have **not started yet**, replacing the slow 2.5s watchdog with a precise `BadTargets` reason. It should never touch a cast the legacy server has already started: the server owns those and sends a real `SMSG_SPELL_GO` / `SMSG_SPELL_FAILURE` that `HandleSpellGo` / `HandleSpellFailure` route to the client, and `HandleSpellFailure` arms the watchdog as a leak backstop.

Verified against a capture: the server reliably sends `SMSG_SPELL_FAILURE` + `SMSG_CAST_FAILED` ~600–970ms after `SMSG_SPELL_START` for the interrupted gather; the old eviction at ~+50ms raced and discarded the cast so that real resolution matched nothing.

## Fix

`DrainPendingCastsForDestroyedTarget` now evicts only not-yet-started casts (normal and pet queues). Started casts stay queued for the server's authoritative resolution.

## Testing

- `dotnet test` — 502/502 pass; updated two destroy-hook tests that encoded the old behavior and added a regression guard.

### DIFF
--- a/HermesProxy.Tests/World/WatchdogEvictionTests.cs
+++ b/HermesProxy.Tests/World/WatchdogEvictionTests.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Frozen;
 using HermesProxy;
 using HermesProxy.Enums;
 using HermesProxy.World;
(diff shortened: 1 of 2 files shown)


jameopotato:
**Hold — needs deeper investigation before merge.**

A few concerns:

### 1. Title / repro accuracy
"Gather nodes stuck after the first tick" implies a common failure, but we've been gathering extensively on Kronos without ever hitting this. The description reads like a specific double-click timing scenario (spam-click node during active channel → second CMSG queued → server starts it → node depletes → destroy-hook evicts). Can you share:
- Exact repro steps: single click or deliberate double-click? Which node type?
- Was this observed on Kronos, Twinstar, or another fork?
- The JSONL bundle showing the actual packet sequence (SPELL_START for the second cast arriving before SMSG_DESTROY_OBJECT)

### 2. Blast radius wider than the symptom
The fix changes `DrainPendingCastsForDestroyedTarget` to skip ALL started casts, not just gather channels. The fast-fail for started casts currently provides immediate BadTargets feedback when a mob dies mid-cast (Frostbolt on a dying mob). With this change, those cases now wait for the server's SPELL_FAILURE (up to ~1s) or the 2.5s watchdog timeout instead of getting instant feedback. That's a noticeable UX regression for a common combat scenario.

### 3. Interaction with prior fix
We already touched this area in `c56cc5d` (movement preemption: exclude channeled spells from `MarkStartedCastsMovementCancelled`). That fix addressed gathering channels getting suppressed by a different path. Both fixes modify "what happens to started casts" and we want to make sure we're not layering fixes that interact in unexpected ways.

### 4. Could this be scoped tighter?
If the real symptom is specifically gather channels, could the `HasStarted` guard be narrowed to `HasStarted && IsChanneledSpell(cast.SpellId)` instead of all started casts? That would preserve the instant BadTargets feedback for normal combat while fixing the specific gather phantom-channel scenario.

Looking forward to the bundle data — that'll make the path forward clear.

Mongrul:
▎ Good catches across the board. Let me address each:
  ▎
  ▎ 1. Repro / scope: The reporter was on Kronos with mining. The specific sequence is the double-click one you
  ▎ described — first click starts the gather channel, second click while the channel is still ticking queues a second
  ▎ cast that the server picks up and starts (SPELL_START on the wire ~50ms before the depleted node's DESTROY_OBJECT
  ▎ arrives). I don't have a clean JSONL bundle to share for this particular repro — the original observation was a
  ▎ tester report on Kronos, validated by reading the c56cc5d-era cast flow. I should have flagged in the PR description
  ▎  that the title overstates how common the failure mode is. Will reword.
  ▎
  ▎ 2. Blast radius: You're right — for non-channeled started casts (Frostbolt on a dying mob) the synthetic
  ▎ SMSG_CAST_FAILED does tear down the cast bar (different from a channel bar), so the instant BadTargets feedback was
  ▎ a real UX improvement over native. Broadening the gate removed that.
  ▎
  ▎ 3. Interaction with c56cc5d: Same code area, same IsChanneledSpell predicate. The two fixes compose cleanly —
  ▎ c56cc5d keeps movement preemption from suppressing the channel-failure path; this one keeps the destroy-hook from
  ▎ racing the same channel-failure path. They share the principle "channels need the server's real SMSG_SPELL_FAILURE
  ▎ to flow through, intermediaries can't substitute."
  ▎
  ▎ 4. Tighter scope: Done. Pushed e572852 narrowing the gate to HasStarted && IsChanneledSpell(cast.SpellId). Started
  ▎ non-channeled casts now still get evicted with BadTargets as before. Only started channels are kept for the server's
  ▎  authoritative resolution. Test renamed and updated to verify both halves (a started Frostbolt-style cast still gets
  ▎  evicted; a started Mining channel does not). 452/452 tests pass.

PR #297, opened 2026-05-23
Fix stack-aura decrement (Lightning Shield + sibling charge-consumed buffs)

Mongrul:
## Summary
- **Observed bug**: Lightning Shield charges don't decrement on the modern client — show 3 through every hit until the aura vanishes whole. JSONL repro from a shaman tester.
- **Root cause**: the aura-slot loop in \`UpdateHandler\` gated on \`UNIT_FIELD_AURA + i\` only — per-slot, fires on spell-ID change. Stack updates change only \`UNIT_FIELD_AURAAPPLICATIONS\` (quad-packed, 4 slots in one uint32). The proxy dropped them entirely. The original narrow gate exists to avoid the \"warlock recasts one DoT, all 4 visible DoTs refresh together\" bug.
- **Fix**: cache the last \`AuraDataInfo\` emitted per \`(unit, slot)\`. When the quad mask fires but the per-slot spell-ID mask doesn't, extract this slot's app byte and compare to the cached value. If different, synthesize a fresh \`AuraInfo\` from the cache with the new Applications count. Adjacent slots whose byte didn't change are still skipped — the warlock-DoT protection is preserved.

## Scope note
The fix targets any aura where the vanilla server updates \`UNIT_FIELD_AURAAPPLICATIONS\` without restating \`UNIT_FIELD_AURA[slot]\`. Lightning Shield is the verified case. Sibling charge-consumed self-buffs (Flurry, etc.) likely share the same wire shape and are also covered, but only Lightning Shield was reproduced end-to-end. Stack-up debuffs like Sunder Armor were *not* validated — if the server restates the spell ID on each re-apply, those go through the existing \`_maskAura\` path and don't touch the new code. \`CloneAuraDataForStackChange\` is direction-agnostic, so the new path is a no-op if it does happen to fire on an increment that was already handled elsewhere.

## Test plan
- [x] Cast Lightning Shield, take 3 melee hits → stacks decrement 3→2→1→aura clears
- [x] Warlock multi-DoT scenario (Corruption + Curse + Immolate on one target) → no spurious refreshes on unrelated slot updates
- [x] \`dotnet test\` — 452/452 pass
- [x] Build clean

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -460,6 +460,13 @@ public sealed class GameSessionData
     public Dictionary<WowGuid128, Dictionary<byte, int>> UnitAuraDurationFull = [];
     public Dictionary<WowGuid128, Dictionary<byte, WowGuid128>> UnitAuraCaster = [];
 
+    // MIRASU (stack-aura-decrement): last AuraDataInfo emitted for each (unit, slot).
+    // Used to detect AURAAPPLICATIONS-quad-only updates where a single slot's app
+    // byte changed (Lightning Shield charge consumed, Sunder Armor stack added,
+    // Devouring Plague tick stacked) and re-emit just that slot without spuriously
+    // refreshing the other three slots packed into the same uint32.
+    public Dictionary<WowGuid128, Dictionary<byte, AuraDataInfo>> UnitAuraLastEmitted = [];
+
     // JimsProxy (Rupture-DoT-Lingering-Icon): combo-point cache + finisher-cast snapshot.
     // Vanilla servers don't send aura duration for enemy debuffs, and CP-scaling finishers
     // (Rupture, Kidney Shot) compute their duration server-side as (base + perCp × CP).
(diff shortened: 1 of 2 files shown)


jameopotato:
**Review question — Sunder Armor in the title/description:**

Does Sunder Armor actually hit the new `appsOnlyChanged` path? When a new Sunder stack is applied, does the vanilla server restate `UNIT_FIELD_AURA` (spell ID) in the update mask, or does it only touch `UNIT_FIELD_AURAAPPLICATIONS`?

If the server restates the spell ID on each stack application (which we believe it does — Sunder stack count was already updating correctly on base HermesProxy), then `_maskAura = true` and the existing code handles it. Sunder would never enter the new stack-change path.

The fix is clearly correct for **charge-consumption decrements** where the spell stays but the app byte drops (Lightning Shield charges, Flurry stacks). We just want to make sure no reasoning in the code depends on Sunder hitting the new path — e.g., the `CloneAuraDataForStackChange` increment direction (wire apps going up) being exercised and tested.

Can you confirm which auras you actually observed hitting the new `appsOnlyChanged` branch in your testing? The diagnostic `aura.slot.update` events in the bundle should show whether Sunder's updates came through `_maskAura` or `appsOnlyChanged`.

Mongrul:
You're right — I didn't separately test Sunder. Updated title and description to match what was actually reproduced
  ▎ (Lightning Shield decrement, with Flurry as a strongly-suspected sibling case). The new code path is harmless if
  ▎ Sunder does happen to enter it (cached AuraDataInfo clone is direction-agnostic), but I won't claim it as a tested
  ▎ fix.

PR #303, opened 2026-05-23
Fix Kronos CMSG_GM_TICKET_CREATE wire format (supersedes PR #301 ticket fix)

Mongrul:
## Summary

The \`CMSG_GM_TICKET_CREATE\` shape from PR #301 was wrong. Kronos's parser actually reads:

\`\`\`cpp
uint32 map;
float x, y, z;
cstring ticketText;
uint32 unk1 (= 0);
uint32 unk2 (= 1);
uint32 unk3 (= 0);
\`\`\`

**No leading category byte. Three trailing uint32s.**

PR #301 kept the leading \`uint8(2)\` (older mangos-zero layout) and only added a single trailing \`uint32\`. That shifted every field after the type byte by 1 *and* left the parser 8 bytes short of EOF, so Kronos still threw \`ByteBufferException\` on every ticket submit.

This PR replaces the prior PR #301 ticket fix with the correct shape (confirmed against Kronos's server-side parser snippet). TBC+ branch unchanged.

### Note on PR #301
PR #301 also fixed AH list trailing \`getAll\` byte and AUTH_SESSION trailing byte — those are independent and remain correct. Only the ticket portion needs this replacement.

If Jim merges PR #301 first, this PR should still apply cleanly (it modifies the same lines further). If Jim merges this first, PR #301's ticket commit will conflict and should be dropped during merge.

## Test plan

- [x] In-game on Kronos: right-click player → Report Player → submit. Server log shows no opcode 517 ByteBufferException; ticket lands in Kronos's GM queue.

### DIFF
--- a/HermesProxy/World/Server/PacketHandlers/SupportTicketHandler.cs
+++ b/HermesProxy/World/Server/PacketHandlers/SupportTicketHandler.cs
@@ -38,19 +38,25 @@ public partial class WorldSocket
 
         if (LegacyVersion.RemovedInVersion(ClientVersionBuild.V2_0_1_6180))
         {
-            packet.WriteUInt8(2); // GMTICKET_BEHAVIOR_HARASSMENT
+            // MIRASU (Kronos ticket parser 2026-05-23 — REVISED): Kronos's
+            // CMSG_GM_TICKET_CREATE handler reads:
+            //   uint32 map; float x; float y; float z;
+            //   cstring ticketText;
+            //   uint32 unk1 (= 0);
+            //   uint32 unk2 (= 1);
+            //   uint32 unk3 (= 0);
+            // No leading category byte. Three trailing uint32s. The prior
+            // shape (uint8 category + cstring "" trailer) was the older
+            // mangos-zero layout; PR #301 corrected the 1-byte trailer to
+            // 4 bytes but the leading category byte still threw the offsets
+            // off by 1, and Kronos wants 3 trailing words not 1. This
+            // matches the parser snippet shared by Kronos directly.
             packet.WriteUInt32(complaint.Header.SelfPlayerMapId);
             packet.WriteVector3(complaint.Header.SelfPlayerPos);
             packet.WriteCString(ticketText);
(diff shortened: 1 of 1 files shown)


Mongrul:
FYI GMs respond vie Mail or /w per Kronos Devs

jameopotato:
**Hold — server-specific branching needed.**

This PR removes the leading `uint8(2)` category byte and changes the trailing data to three uint32s. This matches Kronos's parser, but the wire layout applies to ALL vanilla servers (`RemovedInVersion(V2_0_1_6180)`). The old mangos-zero layout with the category byte may work on vmangos/Twinstar — removing it shifts field offsets and could break ticket submission on those forks.

**Proposed approach: `ServerType` config setting**

Add to `Settings.cs`:
```csharp
public enum ServerType { Kronos, Generic }
public static ServerType ServerFork = ServerType.Kronos;
```

Config: `<add key="ServerType" value="Kronos" />`
Default `Kronos` (this launcher is Kronos-first). Users on vmangos/Twinstar set `Generic`.

Then in the ticket handler:
```csharp
if (Settings.ServerFork == ServerType.Kronos)
{
    // Kronos format: no category byte, three trailing uint32s
    packet.WriteUInt32(map);
    packet.WriteVector3(pos);
    packet.WriteCString(text);
    packet.WriteUInt32(0);
    packet.WriteUInt32(1);
    packet.WriteUInt32(0);
}
else
{
    // Generic/vmangos: category byte + single trailing uint32
    packet.WriteUInt8(2);
    packet.WriteUInt32(map);
    packet.WriteVector3(pos);
    packet.WriteCString(text);
    packet.WriteUInt32(0);
}
```

Same pattern as `LowLatencyMode` / `ThreatEngine` — static bool/enum in Settings, read from HermesProxy.config, gated at the divergence point. Future Kronos-specific wire quirks get gated the same way.

We'll add a launcher UI dropdown to control this setting. Please update the PR with this branching.

jameopotato:
Superseded — the Kronos ticket format fix has been implemented directly on beta with a ServerType config branch (commit bf3c162). The Kronos format is gated on Settings.ServerType == Kronos; Generic/vmangos uses the original mangos-zero layout. Thanks for the server dev parser snippet that made this possible.

PR #309, opened 2026-05-25
Fix My Auctions tab capped at 50 items on 1.14 client

### LINKED ISSUE #305: Failed to learn Spell 3422 (Reason 2)
**Describe the bug**
A weird message "Failed to learn Spell 3422 (Reason 2)" appears when trying to learn Crippling Poison (Rank 1). I assume I cannot learn it because I did not finish the required rogue poison quest. I think I shouldnt be able to click "Train" button or the skill should be red instead of what is happening now but I am not sure.

**Game Server**
Kronos 5

**Hermes Proxy Version**
v5.1.5-beta.1

Image
Image

Mongrul:
Closes #305 (related).

## Problem

The 1.14 Classic Era client's My Auctions tab caps at 50 displayed items. Players with more than 50 active auctions can see the first 50 in the FauxScrollFrame but have no way to reach items 51+. Root cause is a Blizzard regression in the 1.14 client UI — `Blizzard_AuctionUI.xml` defines `BrowsePrevPageButton`/`BrowseNextPageButton` for the Browse tab but **never adds the equivalent buttons for the Auctions tab**. The Lua code at `Blizzard_AuctionUI.lua:1349` iterates display slots using `AuctionFrameAuctions.page * NUM_AUCTION_ITEMS_PER_PAGE` but no UI element exists to flip `page`. Vanilla 1.12 had these buttons (confirmed in PTR-1.12 native client testing: shows ``Items 1 - 50 (65 total)`` with `< Prev` / `Next >` buttons in same position as Browse tab).

Additionally, the modern client's `GetNumAuctionItems(\"owner\")` returns `total=batch` — the wire `totalcount` field is not surfaced to Lua — so even if we got past the 50-item display limit, the pagination math (``totalAuctions > NUM_AUCTION_ITEMS_PER_PAGE``) couldn't trigger anyway.

## Fix

Proxy walks all owner-list pages from the legacy server and combines into a single SMSG that the 1.14 client's `FauxScrollFrame` can scroll through normally. Addons (aux, Auctionator) work too because they read from the same client buffer the walk populates.

**Proxy side** (`HermesProxy/World/{Server,Client}/PacketHandlers/AuctionHandler.cs`):
- `HandleAuctionListOwnerItems` arms a walk on `Offset==0` CMSG (vanilla only); non-zero Offset still passes through so any addon doing per-page fetches still works
- `HandleAuctionListMyItemsResult` accumulates server response pages, **dedupes by `AuctionID`** (Kronos sometimes sends a duplicate of the first page as a second SMSG via its auto-pagination quirk), and finalizes when the server's `TotalItemsCount` is reached or a short page arrives. 1000-item safety cap.
- 2-second post-walk suppression window drops late server responses to walk-issued CMSGs (would otherwise overwrite the modern client's display since the client REPLACES rather than appends on each SMSG)
- Post-cancel/sell auto-refresh path also arms a walk so the auto-refresh produces a combined SMSG (otherwise the 2s suppression would eat it entirely)

**JP sideband** (proxy → addon channel via `CHAT_MSG_ADDON`):
- After each forwarded OWNED result, proxy sends body ``AH_TOTAL\t<total>\t<batch>`` with prefix ``JP``. Addons that want to display the real total count or drive custom pagination UI can subscribe.

**JimsPlus addon** (``Addons/JimsPlus/AuctionsPaginationFix.lua``):
- Listens for JP sideband, parses ``AH_TOTAL``
- On ``AUCTION_HOUSE_SHOW`` seeds ``_G.locked = {}`` as a fallback so aux's ``cancel_auction`` (in ``aux/tabs/auctions/core.lua:45``) doesn't crash with ``attempt to index global 'locked' (a nil value)`` when the user clicks Cancel before aux's ``AUCTION_OWNED_LIST_UPDATE`` handler has fired. aux's module env inherits from ``_G`` via ``__index`` (per ``aux/libs/package.lua``), so the fallback is invisible once aux's own handler runs and writes its own ``locked`` into the module env. **No aux modification required** (aux is third-party and bundled by many players, can't be forked).

## Test plan

- [x] Blizzard My Auctions UI with 65 auctions (50 mithril + 15 thorium) — all visible via scroll
- [x] aux Auctions tab — same 65 visible, cancel button works without ``locked`` crash
- [x] Auctionator — visible count matches
- [x] Cancel an auction (any position) — display auto-updates to 64 within ~1-2s
- [x] Player with ≤ 50 auctions — walk finalizes immediately on first short page, no overhead
- [ ] BC/WotLK servers (ExpansionVersion > 1) — walk skipped entirely, passthrough preserved

### DIFF
--- a/Addons/JimsPlus/AuctionsPaginationFix.lua
+++ b/Addons/JimsPlus/AuctionsPaginationFix.lua
@@ -1,0 +1,42 @@
+-- JimsPlus AuctionsPaginationFix
+-- Listens to JP_AH_TOTAL sideband from the proxy. The proxy walks all owner pages
+-- server-side and combines them into a single SMSG response, so pagination buttons
+-- on the 1.14 client UI are not needed; the FauxScrollFrame just scrolls through
+-- the full list. This file's job is the diagnostic command + sideband tracking;
+-- if it turns out the 1.14 client caps display at 50 items per SMSG, we'll re-add
+-- the injected Next/Prev buttons here as a fallback.
+
+local sidebandTotal = nil
+local sidebandBatch = nil
+
+local function HandleJpSideband(prefix, message)
+    if prefix ~= "JP" then return end
+    local cmd, totalStr, batchStr = strsplit("\t", message)
+    if cmd ~= "AH_TOTAL" then return end
+    sidebandTotal = tonumber(totalStr) or 0
+    sidebandBatch = tonumber(batchStr) or 0
+end
+
+if C_ChatInfo and C_ChatInfo.RegisterAddonMessagePrefix then
+    C_ChatInfo.RegisterAddonMessagePrefix("JP")
+end
+
+local f = CreateFrame("Frame")
+f:RegisterEvent("CHAT_MSG_ADDON")
+f:RegisterEvent("AUCTION_HOUSE_SHOW")
+f:SetScript("OnEvent", function(_, event, arg1, arg2)
+    if event == "CHAT_MSG_ADDON" then
+        HandleJpSideband(arg1, arg2)
+    elseif event == "AUCTION_HOUSE_SHOW" then
+        -- aux fallback: aux's tabs/auctions/core.lua expects a `locked` table to exist before
+        -- cancel_auction() runs, but only initializes it inside its AUCTION_OWNED_LIST_UPDATE
+        -- handler. When the proxy's walk-and-combine delays the first OWNED_LIST_UPDATE,
+        -- aux's Cancel button can race ahead and crash with "attempt to index global 'locked'".
+        -- aux's module env inherits from _G via __index (per aux/libs/package.lua), so seeding
+        -- _G.locked={} satisfies the read until aux's own handler runs and writes to its module
+        -- env. Same for `refresh` so the on_update path doesn't read nil if it runs first.
+        if _G.locked == nil then _G.locked = {} end
+        if _G.refresh == nil then _G.refresh = false end
+    end
+end)
+
--- a/Addons/JimsPlus/JimsPlus.toc
+++ b/Addons/JimsPlus/JimsPlus.toc
@@ -13,6 +13,7 @@ TaxiFix.lua
 TooltipFix.lua
 ZoneSync.lua
 MoonkinSound.lua
+AuctionsPaginationFix.lua
 castbars/PoolManager.lua
 castbars/AnchorManager.lua
 castbars/Data.lua
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -343,6 +343,12 @@ public sealed class GameSessionData
     public List<World.Server.Packets.AuctionItem> AuctionReplicateAccumulator = new();
     public readonly Lock AuctionReplicateLock = new();
     public DateTime AuctionReplicateStartTime;
+    // JimsProxy (issue #305-ah): walk owner-items pages and combine into one SMSG; see memory.
+    public bool AuctionOwnerWalkInProgress;
+    public WowGuid128 AuctionOwnerWalkAuctioneer = WowGuid128.Empty;
+    public List<World.Server.Packets.AuctionItem> AuctionOwnerWalkAccumulator = new();
+    public readonly Lock AuctionOwnerWalkLock = new();
+    public long AuctionOwnerWalkLastFinalizedTickMs;
     public uint LastWhoRequestId;
     public WowGuid128 CurrentPetGuid;
     public WowGuid64 CurrentAttackTarget;        // active CMSG_ATTACK_SWING victim, cleared on ATTACK_STOP/CANCEL_COMBAT
--- a/HermesProxy/World/Client/PacketHandlers/AuctionHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/AuctionHandler.cs
@@ -101,7 +101,127 @@ public partial class WorldClient
         auction.TotalItemsCount = packet.ReadInt32();
         if (LegacyVersion.AddedInVersion(ClientVersionBuild.V2_3_0_7561))
             auction.DesiredDelay = packet.ReadUInt32();
+
+        // JimsProxy (issue #305-ah): walk all owner pages and combine into one SMSG.
+        if (auction.GetUniversalOpcode() == Opcode.SMSG_AUCTION_LIST_OWNED_ITEMS_RESULT
+            && LegacyVersion.ExpansionVersion <= 1)
+        {
+            var gs = GetSession().GameState;
+            const int PageSize = 50;
+            lock (gs.AuctionOwnerWalkLock)
+            {
+                if (!gs.AuctionOwnerWalkInProgress)
+                {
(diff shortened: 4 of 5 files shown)


jameopotato (approved):
## Review

### Summary

Two-prong fix for a Blizzard 1.14 client regression where `Blizzard_AuctionUI.xml` only defines pagination buttons for the Browse tab, leaving "My Auctions" capped at 50 visible items (vanilla 1.12 had Next/Prev buttons on both tabs).

1. **Proxy side** — on vanilla (`ExpansionVersion <= 1`), the proxy intercepts `CMSG_AUCTION_LIST_OWNED_ITEMS` with `Offset==0`, walks all owner-list pages server-side, dedupes by `AuctionID` (Kronos auto-pagination sometimes duplicates page 1), and emits a single combined `SMSG_AUCTION_LIST_OWNED_ITEMS_RESULT` with `Items` set to the full combined list. 1000-item safety cap. 2-second post-walk suppression window drops late server responses to walk-issued CMSGs.
2. **Addon side** — `Addons/JimsPlus/AuctionsPaginationFix.lua` listens for `JP\tAH_TOTAL\t<total>\t<batch>` sideband (currently captured but not displayed; reserved for future fallback UI), and on `AUCTION_HOUSE_SHOW` seeds `_G.locked = {}` and `_G.refresh = false` to satisfy aux's `cancel_auction` global reads before aux's own `AUCTION_OWNED_LIST_UPDATE` handler initializes them.

### Correctness

Read all 5 files end-to-end plus the surrounding `HandleAuctionListMyItemsResult` and the regular CMSG forwarder.

**Verified correct:**
- `LegacyVersion.ExpansionVersion <= 1` gate confines all walk logic to vanilla. BC/WotLK passthrough preserved (regular `WorldPacket(...)` flow at the end of the Server handler is unchanged).
- `auction.Offset == 0` gate in Server handler means addons doing per-page fetches still work (their offset != 0 → passthrough).
- `auction.GetUniversalOpcode() == Opcode.SMSG_AUCTION_LIST_OWNED_ITEMS_RESULT` correctly distinguishes OWNED from BIDDER (the handler is shared between the two `SMSG_AUCTION_LIST_*_ITEMS_RESULT` opcodes).
- Walk-continuation CMSG wire format (`WriteGuid(auctioneer.To64()) + WriteUInt32(totalSoFar)`) matches the existing passthrough format.
- `AuctionOwnerWalkAuctioneer` cached at walk-start, reused for continuation — robust against the client closing/reopening AH mid-walk.
- Dedupe HashSet (rebuilt per page) — cheap at the actual scale (≤1000 items × O(50) inserts per page).
- Finalize copies accumulator into a new `List<AuctionItem>` before assigning to `auction.Items`, then clears accumulator. Defensive.
- Walk state lives in `GameState`; session rebuild on disconnect cleans naturally.
- Cancel/sell auto-refresh path (`Client/AuctionHandler.cs` ~line 451) arms the walk before sending the refresh CMSG, so the refresh produces a combined SMSG instead of being dropped by the suppression window — clever, and the comment explains why.

**Concerns (not blocking):**

**a. Lock scope crosses I/O.** `AuctionOwnerWalkLock` is held across `SendPacketToClient` (passthrough path), `SendPacketToServer` (walk-continuation), and another `SendPacketToClient` (finalize). The lock is only contended between the CMSG handler (`Server/AuctionHandler.cs`) and SMSG handler (`Client/AuctionHandler.cs`); if those run on the same per-session thread, the lock is technically unnecessary. If they can run on different threads, holding it across socket I/O is a potential stall point — not a deadlock since no nested locks, but a fairness concern. Worth a quick check of the proxy's session-thread model; if single-threaded per session, the lock could be removed for clarity.

**b. `_G.locked` and `_G.refresh` global pollution.** Names generic enough that any other addon writing to these globals between AH-shows would have its data wiped on the next show. The PR's defense (aux's module env shadows `_G` reads via `__index`) handles aux specifically, but a third addon using these names would collide. Tradeoff is necessary — names are dictated by aux's source. Risk is real but probably small. The nil-guards (`if _G.locked == nil then _G.locked = {}`) already preserve any existing value — good.

**c. Captured-but-unused sideband data.** `sidebandTotal` and `sidebandBatch` are stored in `HandleJpSideband` and never read by any other code in the file. Per the file's docstring, this is preparation for a future fallback UI. Acceptable as a hook; worth knowing it's currently inert.

**d. 2000ms post-walk suppression window.** Drops late server responses to walk-issued CMSGs. For typical RTT (~100ms) this is generous, but if Kronos's auto-pagination quirk delays a duplicate-page response significantly (>2s), it slips through and overwrites the client display with a partial page. Hard to hit in practice; observable if it ever does. Worth grepping tester JSONL bundles after merge for `auction.owner_walk.passthrough_outside_walk` showing partial-page values within ~5s of a finalize.

**e. `auction.TotalItemsCount = auction.Items.Count` after combine.** Replaces the wire `TotalItemsCount` with the actual combined count. PR body notes the 1.14 Lua doesn't surface this field anyway — purely for any addon that reads raw packet fields. Harmless.

No correctness bugs found.

### Efficiency

Walk runs once per AH open per session (or per auto-refresh). Per page: O(50) hashset operations, one allocation for the page-local existingIds set. For the 1000-cap pathological case: ~20K hashset operations total across 20 pages, microseconds. Sideband packet is a single ChatPkt — same as the established `prefix=JP, body=<CMD>\t<args>` pattern.

Heavy AH users (hundreds of auctions) get a few extra server round trips to fetch all pages, then a single combined SMSG. Trade: latency on AH open goes up (~RTT × pages_needed), but display correctness becomes possible. Worth it.

### Future compatibility

- Walk logic gated on `ExpansionVersion <= 1` — silently does the right thing on BC/WotLK (no walk).
- New `AuctionOwnerWalk*` fields in `GameState` are private to this code path; no surface exposed elsewhere.
- JP sideband uses the established `prefix=JP, body=<CMD>\t<args>` pattern. Future sideband consumers can add prefix matchers cleanly.
- If Blizzard ever fixes the 1.14 client UI to re-add Auctions pagination, this PR remains compatible (the addon's stored sideband data + the proxy walk both still work; client just won't need the walk).
- DBC-style server-quirk handling is locked behind ExpansionVersion gate, but Kronos-specific (auto-pagination dupe) behavior assumes the dedupe is harmless on non-Kronos vanilla emulators. Should be — dedup-by-AuctionID is universally safe.

### Verdict — Approve with suggestions

The PR addresses a real player-visible bug (heavy AH user can't reach items 51+), is well-scoped per file, has substantial test coverage (5/6 plan items including aux + Auctionator interop), and the multi-component architecture is justified (proxy walk + sideband + aux shim each handle a distinct concern). The concerns above are non-blocking polish.

**Optional suggestions for follow-up (not before merge):**
- Verify the lock-vs-thread model. If session packet handlers are single-threaded per session, the lock can be removed in a cleanup PR.
- After a few sessions on beta, grep tester JSONL bundles for `auction.owner_walk.passthrough_outside_walk` to validate the 2000ms suppression window is wide enough.
- The captured-but-unused sideband fields in the addon could either be commented "reserved for future fallback UI" or deleted entirely until the fallback is needed.

PR #124, opened 2026-05-04
Fix GCD sweep drift: synthesize SMSG_SPELL_COOLDOWN

jameopotato:
## Summary

- After `BeginGcd` in `HandleSpellGo`, synthesize `SMSG_SPELL_COOLDOWN` with `SPELL_COOLDOWN_FLAG_INCLUDE_GCD` (0x01) to force the 1.14 client to anchor GCD start to packet arrival time
- Without this, the client chains `gcd_start = old_start + 1500ms` but actual cast-to-cast is ~1600ms (RTT overhead), causing the GCD sweep animation to drift clockwise and wrap every ~15 casts
- Confirmed: 42/42 paired `gcd.begin` + `gcd.cooldown_synth` events, sweep locked at 12 o'clock during sustained AE spam

## Test plan

- [x] Spam instant-cast AoE (Arcane Explosion) for 30+ casts
- [x] Verify GCD sweep stays at 12 o'clock (no progressive clockwise drift)
- [x] Check JSONL for 1:1 `gcd.begin` / `gcd.cooldown_synth` event pairing
- [x] Verify no regression on cast-time spells (Frostbolt, Fireball)
- [x] Verify no regression on off-GCD spells (Counterspell)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #313, opened 2026-05-25
Spell: configurable GCD/cast-time hold window (SpellQueueWindow, default 1300ms)

jameopotato:
## Summary

The proxy was holding presses for the entire GCD (1500 ms) and the entire cast bar (3 s for Frostbolt, longer for Pyroblast/Hearth/etc.). The 1.14 client's `SpellQueueWindow` CVar (default 400 ms) tells the client to expect the server to queue presses only in the LAST 400 ms of an active GCD or cast bar. Earlier presses are not queued — they're forwarded, and the server returns NOT_READY / SpellInProgress, which the client renders as standard "Spell not ready yet" / "Spell is already in progress" feedback.

Our 1500 ms / full-cast hold was over-doing the 1.14 server's job. Players mashing in the early portion of a cast saw their press silently absorbed and fired up to 2.5 s later — disconnected from the keypress and contributing to "feels stuck" reports.

## Behavior after this PR

- Instant pressed in last 400 ms of previous GCD → queued, fires at GCD expiry (same as today)
- Instant pressed earlier than that → forwarded → server NOT_READY → client shows standard popup (player learns to time presses)
- Cast-time press in last 400 ms of cast bar → queued, fires on completion (same as today)
- Cast-time press earlier in cast → forwarded → server SpellInProgress → client shows standard popup

Matches the 1.14 retail server's `SpellQueueWindow=400` semantics exactly.

## Implementation

Adds two narrow-window variants alongside the existing methods (so other callers that need full-window semantics aren't affected):

- `IsInGcdQueueWindow()` — true only when GCD has ≤400 ms remaining
- `HasStartedCastInQueueWindow()` — true only when a cast-time cast has ≤400 ms remaining

Adds a `StartedAtTickMs` field on `ClientCastRequest`, set in `TryMarkPendingNormalCastStarted` / `TryMarkPendingPetCastStarted` when SPELL_START arrives. The cast-time gate computes `(StartedAtTickMs + StartedCastTimeMs) - now` and gates on `remaining <= QueueWindowMs`.

Only the `HandleCastSpell` hold gates (lines 340 + 365) use the new variants. Other callers keep the wider checks:
- `HasStartedNormalCast` at `Server/SpellHandler.cs:592` (item-use duplicate guard)
- `IsGcdHoldActive` at `Client/SpellHandler.cs:427` (held-cast-on-failure release)

Adds a constant `QueueWindowMs = 400` in `GlobalSessionData.cs` for the window width.

## What this PR does NOT do

- No NOT_READY suppression. The server's response flows through to the client unchanged. Players who don't want the "Spell not ready" popup can disable "Suppress Error Speech and Text" in client settings (or set `SuppressSpellCastErrors=true` in `HermesProxy.config` — existing feature, currently active in Low Latency Mode).
- The OUTER hold path (`HasForwardedPendingCast` at L441) unchanged.
- The GCD sweep synthesis (PR #124 / commit 3c2778f) unchanged — that invariant stays, fires at SPELL_GO time, drives the action-button GCD bar.

## Recommended testing strategy

This PR changes player-visible UX (NOT_READY popups should appear more often during heavy mashing). It is **not** an integrity fix; it is a feel/architectural alignment with 1.14 retail server behavior. Recommend testing in a separate bundle from the integrity fixes so any feedback about feel can be cleanly attributed to this change rather than confounded with the others.

## Test plan

- Instant spam during GCD: should see NOT_READY popups for early-GCD presses, but smooth queue→fire for last-400ms presses.
- Cast Frostbolt → press Fire Blast at various times:
  - t=0.5s into cast → popup, no queue
  - t=2.7s into cast (last 400ms) → silent queue, fires at completion
- A/B feel: compare to current beta. Last-400 ms behavior unchanged; early-cast/early-GCD behavior changes from silent absorption to visible NOT_READY popups.

### DIFF
--- a/HermesProxy/Configuration/Settings.cs
+++ b/HermesProxy/Configuration/Settings.cs
@@ -73,6 +73,12 @@ public static class Settings
     public static bool IdentityPinnedCastIds;
     // Effective gate for the T1 mechanism: the sub-toggle is live only under LowLatencyMode.
     public static bool IdentityPinnedCastIdsActive => LowLatencyMode && IdentityPinnedCastIds;
+    // JimsProxy (#313): width (ms) of the spell-queue hold window. A press arriving in the
+    // last SpellQueueWindowMs of an active GCD or cast bar is held and fired at expiry; earlier
+    // presses are forwarded and the server arbitrates (NOT_READY / SpellInProgress). Mirrors the
+    // 1.14 SpellQueueWindow contract. The launcher exposes 400 (retail-accurate) / 1000 / 1300
+    // (smoothest, closest to the old full-hold); lower values are allowed, capped at 1300 ms.
+    public static int SpellQueueWindowMs;
     // JimsProxy (PR #228 follow-up): synthesize SMSG_THREAT_UPDATE / HIGHEST /
     // CLEAR so the modern client's native threat APIs (UnitDetailedThreatSituation,
     // UNIT_THREAT_LIST_UPDATE) populate. Default on. Disable for players who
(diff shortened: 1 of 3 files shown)


jameopotato:
Reworked + rebased onto beta:

- Hold-window width is now configurable via `Framework.Settings.SpellQueueWindowMs` (config key `SpellQueueWindowMs`), **default 1300ms, capped at 1300, lower values allowed** — instead of the original hardcoded 400ms.
- The launcher will expose this as an advanced setting (presets 400 retail-accurate / 1000 / 1300 smoothest) — separate launcher PR.
- Rebased onto current beta; resolved a conflict where #334's `HasStartedCastOnGameObject` and the off-GCD `IsOffGcd` field collided with #313's additions — both preserved.
- Build clean (0 warnings / 0 errors), 535 tests pass.

PR #315, opened 2026-05-25
Fix: clear cast state on transfer abort

jameopotato:
## Summary

`HandleTransferAborted` clears `IsWaitingForNewWorld` but never swept the cast queues. Any `CMSG_CAST_SPELL` forwarded between `TransferPending` and `TransferAborted` becomes an orphan — `HasForwardedPendingCast()` returns true permanently and the OUTER hold path blocks all subsequent casts for the rest of the session.

Found during an audit of all state transitions after shipping #312 (the NEW_WORLD equivalent of this fix).

## Triggers

- Instance full (raid at capacity)
- Level too low for instance
- Transfer cancelled by server for any reason

## Fix

Same three-line cleanup as `HandleNewWorld` (#312): `ResetInFlightCastState` + `CancelGcdHold` + `ClearHeldCastTimeCast`. Diagnostic event only fires when state was actually cleared.

## Test plan

- [x] `dotnet build` clean
- [ ] Attempt to enter a full instance → transfer aborted → verify casting still works

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
@@ -314,6 +314,24 @@ public partial class WorldClient
 
         SendPacketToClient(transfer);
         GetSession().GameState.IsWaitingForNewWorld = false;
+
+        var clearedCounts = GetSession().GameState.ResetInFlightCastState();
+        var droppedGcdHold = GetSession().GameState.CancelGcdHold();
+        var droppedCastTimeHold = GetSession().GameState.ClearHeldCastTimeCast();
+        if (clearedCounts.normalCasts > 0 || clearedCounts.petCasts > 0 ||
+            droppedGcdHold != null || droppedCastTimeHold != null)
+        {
+            Log.Event("session.transfer_aborted.cast_state_cleared", new
+            {
+                aborted_map_id = transfer.MapID,
+                reason = transfer.Reason.ToString(),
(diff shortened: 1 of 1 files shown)


Mongrul:
Heads-up — the red `test`/`verify` here is the same pre-existing `WatchdogEvictionTests.DrainPendingCastsForDestroyedTarget_Match_EvictsAllExceptStartedChannels` failure that's blocking #316 too, not anything in your change. All three platform `build` jobs passed clean.

Posted the root-cause analysis + a 2-spot diff to fix it on #316: https://github.com/jameopotato/jimsproxy/pull/316#issuecomment-4534185917

Once that drain-fix lands on beta (folded into #316 or as a separate commit), this PR's CI should clear on the next push/rebase.

jameopotato:
Holding open as audit-found defensive fix. The race window is narrow: CMSG_CAST_SPELL forwarded → SMSG_TRANSFER_PENDING arrives → `HandleTransferPending` sends `SuspendToken` (`MovementHandler.cs:270`) freezing further client input. The only cast that can be in flight at TRANSFER_ABORTED time is one sent before TRANSFER_PENDING, and at that point the server is still on the source map (transfer was rejected, not completed), so the in-flight cast should still resolve normally.

No confirmed orphan symptom on TRANSFER_ABORTED yet — unlike #312 (NEW_WORLD) which had the 6h warlock BG lockout. Merge this if anyone reports a casting lockout right after walking into a full raid / level-gated instance.

PR #316, opened 2026-05-25
Fix unit-targeted spells failing out-of-range while moving

### LINKED ISSUE #314: Unit-targeted spells fail out-of-range while moving (position lag)
## Problem

Unit-targeted spells (Fire Blast, Counterspell, etc.) cast at near-max range while moving frequently fail with `SPELL_FAILED_OUT_OF_RANGE` even though the 1.14 client believes the player is in range. Same root cause as the reticle fix in PR #299: the modern client's `CMSG_CAST_SPELL` carries no movement info in the vanilla wire format, so the legacy server checks range against its last-cached position which trails movement heartbeats by up to ~500ms.

The cast-reject-retry cycle at ~220ms RTT creates a ~500ms dead window where the player is visually in range but every attempt is rejected against the stale position.

## Evidence

`jimsproxy-20260524-224528.jsonl` — Fire Blast (spell 10199) approach-spam sequences:
- Lines 1920-2180: three consecutive `spell.failure.routed` (reason=1, broadcast_skipped=true) before a successful cast
- Line 5662: single press while approaching, rejected 247ms later, player stops 93ms after rejection

All failures show `dequeued:false`, `foundActiveCastId:false` pattern. Smoothed RTT was 219ms throughout.

## Proposed fix

Extend `SyncLegacyServerPositionBeforeCast` (added in PR #299) to fire for unit-targeted casts while moving, not just reticle/ground-targeted casts. When `MoveUpdate` is present and the target is a unit (not self), synthesize `MSG_MOVE_HEARTBEAT` to the legacy server before the cast packet so the server's position view is current.

The reticle fix's `DstLocation != null` gate currently excludes unit targets. Options:
1. **Remove the DstLocation gate entirely** — fire heartbeat sync for any cast while moving. Simple, slightly more packets but MSG_MOVE_HEARTBEAT is cheap.
2. **Add unit-target distance check** — only fire when computed distance to target > threshold. Requires reading target position from cached UpdateObject data. More precise but more complex.
3. **Option 1 with a melee exclusion** — skip when target distance < 5 yds (melee range never has this problem). Middle ground.

The dest-nudge portion of PR #299 does not apply to unit targets (no dest to nudge). Only the heartbeat sync matters here.

## Test plan

- Approach a mob while spamming Fire Blast at max range (36 yds) — should land on first in-range press
- Melee abilities while moving in melee range — should be unaffected
- Standing-still casts at max range — should be unaffected (no MoveUpdate)

jameopotato:
## Summary

Closes #314.

- Extends `SyncLegacyServerPositionBeforeCast` (PR #299) to fire MSG_MOVE_HEARTBEAT for **unit-targeted** casts while moving, not just reticle/ground-targeted casts
- The `DstLocation != null` gate previously excluded spells like Fire Blast, Counterspell — the server checked range against its stale cached position (~500ms behind) and rejected casts the client considered in-range
- Dest-nudge (1.5yd pull toward player) remains reticle-only — no change to existing reticle behavior
- Self-casts excluded (no range check to fail)
- New `cast.move_sync_for_unit_target` diagnostic event for JSONL tracing

## Test plan

- [ ] Approach mob while spamming Fire Blast at ~36 yds — should land on first in-range press
- [ ] Melee abilities while moving in melee range — unaffected
- [ ] Standing-still casts at max range — unaffected (no MoveUpdate)
- [ ] Reticle casts while moving (Distract, grenades) — unchanged behavior from PR #299
- [ ] Channeled spells while moving — still correctly interrupted

### DIFF
--- a/HermesProxy/World/Server/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Server/PacketHandlers/SpellHandler.cs
@@ -535,25 +535,15 @@ public partial class WorldSocket
     /// </summary>
     private void SyncLegacyServerPositionBeforeCast(SpellCastRequest castRequest, string source)
     {
-        // Gate as tightly as possible — only run for the exact case this fix targets:
-        // reticle (ground-targeted) casts at long range while the player is moving.
-        // Everything else (unit-targeted, self-cast, short-range reticle, standing-
-        // still, no MoveUpdate on the wire) goes through the proxy's existing,
-        // unchanged forwarding path. This avoids touching the normal cast pipeline
-        // and minimizes the surface for any regression on standard player casting.
         if (castRequest.MoveUpdate == null)
             return;
-        if (castRequest.Target.DstLocation == null)
-            return;
 
-        var player = castRequest.MoveUpdate.Position;
-        var dest = castRequest.Target.DstLocation.Location;
-        float dx = dest.X - player.X;
-        float dy = dest.Y - player.Y;
-        float dz = dest.Z - player.Z;
-        float distance = MathF.Sqrt(dx * dx + dy * dy + dz * dz);
-        if (distance <= 15f)
-            return; // close-range reticle casts (Goblin Sapper at 5 yds, etc.) don't need the fix
+        bool hasUnitTarget = !castRequest.Target.Unit.IsEmpty()
+            && castRequest.Target.Unit != GetSession().GameState.CurrentPlayerGuid;
+        bool hasDstLocation = castRequest.Target.DstLocation != null;
+
+        if (!hasUnitTarget && !hasDstLocation)
+            return;
 
         try
         {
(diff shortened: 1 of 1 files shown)


Mongrul:
Heads-up on the CI red — it's not your change. The `test` job is failing on a pre-existing test bug on `beta` that blocks every PR right now:

```
HermesProxy.Tests.World.WatchdogEvictionTests.DrainPendingCastsForDestroyedTarget_Match_EvictsAllExceptStartedChannels
Expected: 2, Actual: 1
```

All three platform `build` jobs (macos/ubuntu/windows) passed clean — only `test` and `verify` (which gates on `test`) failed. The test exists on `beta` HEAD; same failure reproduces with no changes applied.

## Root cause

The test's name documents the intended behavior: **evict all matching casts EXCEPT started channels**. But the implementation in `GlobalSessionData.cs:1385` short-circuits on `!cast.HasStarted`, so any started cast — channel or not — survives the drain:

```csharp
// current — evicts only NOT-STARTED matches; keeps ALL started casts
if (!cast.HasStarted && !cast.TargetGuid.IsEmpty() && cast.TargetGuid == destroyedGuid)
```

The test enqueues a started non-channel (spell 100, `hasStarted: true`) aimed at the destroyed target and expects it to be evicted. Implementation keeps it.

## Proposed fix

Replace the `!cast.HasStarted` short-circuit with an explicit "started AND channeled" guard, so non-channeled started casts get drained but real channels (where the server resolves the cast) survive:

```diff
         var keepNormal = new List<ClientCastRequest>();
         while (PendingNormalCasts.TryDequeue(out var cast))
         {
-            if (!cast.HasStarted && !cast.TargetGuid.IsEmpty() && cast.TargetGuid == destroyedGuid)
+            bool targetMatches = !cast.TargetGuid.IsEmpty() && cast.TargetGuid == destroyedGuid;
+            bool isStartedChannel = cast.HasStarted && GameData.IsChanneledSpell(cast.SpellId);
+            if (targetMatches && !isStartedChannel)
             {
                 normalEvicted.Add(cast);
                 // DIAGNOSTIC ...
             }
             else
                 keepNormal.Add(cast);
         }
```

Same shape for the pet loop at line 1409. Two-spot change.

Verified locally: all 18 `WatchdogEvictionTests` go green, full 502-test suite passes, no regressions.

Closing my parallel #318 since your #316 is the cleaner take on #314 — wanted to leave you this drain-fix as a head start so CI clears for both. Happy to send it as a separate PR if you'd rather review it in isolation.

PR #318, opened 2026-05-25
Fix unit-targeted spells failing out-of-range while moving (#314)

### LINKED ISSUE #314: Unit-targeted spells fail out-of-range while moving (position lag)

Mongrul:
Extends the position-sync fix from PR #299 (reticle/ground-target casts) to also cover unit-targeted casts. Same root cause: legacy 1.12 CMSG_CAST_SPELL carries no movement info, so the server checks range against its last-cached position which trails movement heartbeats by up to ~500ms — at 200ms+ RTT this creates a ~500ms dead window where the client believes you're in range but every press gets rejected.

## Change

Moved the `MSG_MOVE_HEARTBEAT` sync out of the `DstLocation != null` gate so it fires for **any** cast where `MoveUpdate` is present, not just reticle. The dest-nudge (1.5 yd safety margin against vmangos anticheat) stays reticle-only since unit-targeted casts have no destination to nudge — the server resolves target position from its own object cache.

Renamed the diagnostic event from `cast.move_sync_for_reticle` to `cast.position_sync` with a generic payload (`target_kind: "reticle" | "unit"`, `dest_distance`, `nudged`). Lets tester bundles confirm the sync engaged on both code paths.

## Cost

Adopts Option 1 from the issue: always fire heartbeat when MoveUpdate is present, no melee-range exclusion. Costs an extra `MSG_MOVE_HEARTBEAT` per moving melee swing (~2 extra pps during sustained melee) — invisible noise compared to the 60+ pps a busy session already sees.

## Validation

In-game tested on Twinstar. JSONL confirms `cast.position_sync` fired for:
- Frostbolt R9 (spell 10181) — `target_kind: "unit"`, `nudged: false`
- Scorch R3 (spell 10207) — `target_kind: "unit"`, `nudged: false`

Sustained press-while-moving spam in the bundle matches the failure pattern described in the issue. No regression in reticle-cast behavior (PR #299 dest-nudge path still gates on `DstLocation != null` and `distance > 15f`).

## Test plan

- [x] Spam Frostbolt/Scorch at max range while approaching a mob — lands on first in-range press
- [x] Reticle spells at max range while moving — still land (PR #299 path unchanged)
- [ ] Melee abilities while moving in melee range — should be unaffected (extra heartbeat per swing is harmless)
- [ ] Standing-still casts at max range — should be unaffected (no MoveUpdate → early return)

Closes #314.

### DIFF
--- a/HermesProxy/World/Server/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Server/PacketHandlers/SpellHandler.cs
@@ -512,51 +512,39 @@ public partial class WorldSocket
     }
 
     /// <summary>
-    /// MIRASU (cast-while-moving-out-of-range 2026-05-23): reticle-targeted spells
-    /// (Distract, Goblin Sapper Charge, Iron Grenade, Flare, etc.) fail with
-    /// SpellCastResult.OutOfRange when cast at max range while the player is moving.
-    /// The modern 1.14 CMSG_CAST_SPELL / CMSG_USE_ITEM packets carry a MoveUpdate with
-    /// the client's current position; the legacy 1.12 protocol's CMSG_CAST_SPELL has
-    /// no movement info, so the server uses its last-known position for the
-    /// caster-to-dest range check. Because periodic movement heartbeats run on
-    /// ~500 ms intervals, the legacy server's view of the player position can lag
-    /// ~hundreds of ms behind the client when the cast lands — at max range that's
-    /// enough to fail the range check by a few yards. Native 1.12 clients don't hit
-    /// this because they compute the dest position from the same lagged server-side
-    /// position the engine uses for everything.
+    /// MIRASU (cast-while-moving-out-of-range): casts at long range while the player
+    /// is moving fail with SpellCastResult.OutOfRange because the legacy 1.12
+    /// CMSG_CAST_SPELL has no movement info — the server checks range against its
+    /// last-cached position which trails movement heartbeats by up to ~500ms.
     ///
-    /// Fix: when MoveUpdate is present, fan it out to the legacy server as
-    /// MSG_MOVE_HEARTBEAT first so the server's position view is in sync with
-    /// whatever position the client used to pick the dest. The heartbeat is the
-    /// cheapest movement opcode (no flag transitions, no state-machine effects) so
-    /// it's safe to inject ad-hoc. Channeled spells still get correctly interrupted
-    /// by ongoing movement because the heartbeat carries the moving MovementFlags;
-    /// only the range check benefits.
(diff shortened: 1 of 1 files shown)


Mongrul:
Closing in favor of @jameopotato's #316 — same fix, his self-cast exclusion (`Target.Unit != CurrentPlayerGuid`) is the cleaner gate and preserving the existing `cast.move_sync_for_reticle` event name keeps backwards compat for tester-bundle filters. Sorry for the parallel work — should have checked open PRs before picking up the issue.

PR #319, opened 2026-05-25
Restore targeted /spit emote

jameopotato:
## Summary

Blizzard modified the modern 1.14 client's `EmotesTextData` DB2 table (post-2019) to replace all targeted `/spit` text variants with "spits on the ground" — effectively neutering the emote even when targeting another player.

- **DB2 hotfix**: Pushes corrected `EmotesTextData` records at login to restore the original vanilla strings (`"%s spits on %s."`, `"%s spits on you."`, `"You spit on %s."`) for the three targeted relationship flags
- **Target injection**: Caches the player's current selection from `CMSG_SET_SELECTION` in `GameState.CurrentSelection`, and injects it into `CMSG_SEND_TEXT_EMOTE` when emote 89 (spit) arrives with an empty target GUID — safety net in case the client also strips the target, not just the display text

## Test plan

- [ ] Target a player, `/spit` — verify chat shows "You spit on PlayerName." (not "on the ground")
- [ ] Other player sees "YourName spits on you." in their chat
- [ ] Nearby players see "YourName spits on PlayerName."
- [ ] `/spit` with no target — still shows "You spit on the ground." (untargeted path unchanged)
- [ ] Other emotes (`/wave`, `/dance`, `/rude`) still work correctly with targets
- [ ] Verify hotfix loads at login (check proxy log for EmotesTextData hotfix records)

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -349,6 +349,7 @@ public sealed class GameSessionData
     public DateTime AuctionReplicateStartTime;
     public uint LastWhoRequestId;
     public WowGuid128 CurrentPetGuid;
+    public WowGuid128 CurrentSelection;
     public WowGuid64 CurrentAttackTarget;        // active CMSG_ATTACK_SWING victim, cleared on ATTACK_STOP/CANCEL_COMBAT
     public bool WaitingForAttackStart;           // true between CMSG_ATTACK_SWING and SMSG_ATTACK_START
     public bool DeferredAttackStop;              // CMSG_ATTACK_STOP received while waiting for SMSG_ATTACK_START
--- a/HermesProxy/World/GameData.cs
+++ b/HermesProxy/World/GameData.cs
@@ -2410,6 +2410,7 @@ public static partial class GameData
     // on top of the universal hotfix at startup when Settings.ServerType names
     // a known emulator. Distinct ID range so it doesn't collide with the
     // universal HotfixItemSparseBegin block.
+    public const uint HotfixEmotesTextDataBegin = 310000;
     public const uint HotfixItemSparseEmulatorBegin = 300000;
     public static Dictionary<uint, HotfixRecord> Hotfixes = [];
     public static void LoadHotfixes()
@@ -2438,6 +2439,7 @@ public static partial class GameData
         LoadCreatureDisplayInfoHotfixes();
         LoadCreatureDisplayInfoExtraHotfixes();
         LoadCreatureDisplayInfoOptionHotfixes();
+        LoadEmotesTextDataHotfixes();
     }
 
     public static void LoadAreaTriggerHotfixes()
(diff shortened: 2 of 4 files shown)


jameopotato:
## Review

### Summary

Blizzard altered the 1.14 client's `EmotesTextData` DB2 (post-2019) to replace targeted /spit variants with "spits on the ground" regardless of target — confirmed via wago.tools CSV for build 42597. PR has two prongs:
1. **DB2 hotfix** restoring vanilla strings for the three relationship-flag rows (458, 459, 460).
2. **Target injection** caching `CMSG_SET_SELECTION` and substituting on `CMSG_SEND_TEXT_EMOTE` (emote 89) when the client sends an empty target — safety net in case the client also strips the target GUID alongside the text.

### Correctness

**Verified correct:**
- IDs 458/459/460 + relationship flags (0/1/2 mapping) confirmed via wago.tools CSV. Current Classic strings are all "spits on the ground"; vanilla strings (`%s spits on %s.`, `%s spits on you.`, `You spit on %s.`) match the strings in `spitFixes`.
- `TEXTEMOTE_SPIT = 89` is the correct vanilla TextEmoteData ID for /spit.
- `CurrentSelection` lifecycle: written on every `CMSG_SET_SELECTION`, read only when an empty-target /spit arrives. Empty selection → empty cache, override doesn't apply. GameState rebuild on session/disconnect wipes naturally. No lock needed (CMSG handlers are single-threaded per session).
- `HotfixEmotesTextDataBegin = 310000` distinct from other hotfix ID bases (`HotfixItemSparseEmulatorBegin = 300000`).
- `LoadEmotesTextDataHotfixes()` wired into `LoadHotfixes()`.
- Target-injection gate (`emote.EmoteID == TEXTEMOTE_SPIT && target.IsEmpty()`) only acts on the specific case; other emotes and targeted spits pass through unchanged.

### 🛑 Issue — hotfix payload is incomplete

The DBD definition for `EmotesTextData` in Classic Era builds (`WoWDBDefs/EmotesTextData.dbd`, layout `2682E49B` covering 1.13.x and 2.5.x) is:

```
ID                       int32  $noninline,id$    (record key, not in payload)
Text_lang                locstring                (CString in single-locale)
RelationshipFlags        u8                       (inline)
EmotesTextID             u16    $noninline,relation$  (separate section, FK to EmotesText.ID)
```

The PR writes only `WriteCString(text)`. Missing **`RelationshipFlags` (u8 inline)**.

Comparing to a known-working loader like `LoadItemEffectHotfixes` (`GameData.cs:5073-5108`) — every inline field of the table is written in order, typed correctly. The pattern is "write the full inline record"; partial writes likely cause the modern client to reject or misparse the hotfix. The CSV shows distinct relationship-flag values per row (458=0, 459=1, 460=2), so they're meaningful, not just padding.

The `EmotesTextID` noninline relation is a separate concern — existing hotfix loaders don't appear to write noninline-section data, so it may be excluded from the payload by design. But the inline `RelationshipFlags` definitely needs to be there.

**Suggested fix:**

```csharp
ReadOnlySpan<(uint id, string text, byte relationshipFlags)> spitFixes =
[
    (458, "%s spits on %s.",  0),
    (459, "%s spits on you.", 1),
    (460, "You spit on %s.",  2),
];
foreach (var (id, text, relFlags) in spitFixes)
{
    counter++;
    HotfixRecord record = new()
    {
        TableHash = DB2Hash.EmotesTextData,
        HotfixId = HotfixEmotesTextDataBegin + counter,
        Status = HotfixStatus.Valid,
        RecordId = id,
    };
    record.UniqueId = record.HotfixId;
    record.HotfixContent.WriteCString(text);
    record.HotfixContent.WriteUInt8(relFlags);  // <-- missing
    Hotfixes.Add(record.HotfixId, record);
}
```

### Efficiency

3 hotfix records added once at startup; one extra `WowGuid128` field write per `CMSG_SET_SELECTION`; one extra check per `CMSG_SEND_TEXT_EMOTE`. All cold paths. Negligible.

### Future compatibility

- Hotfix mechanism is the established pattern in `GameData.cs` — fine.
- DBD layout has been stable for EmotesTextData across all Classic builds (1.13.x → 2.5.x same layout). Future rebases shouldn't conflict.
- `CurrentSelection` field is a clean addition; no API surface conflicts.

### Verdict — Request changes

The missing `RelationshipFlags` u8 likely breaks the DB2 hotfix in-game. Without it, the modern client either rejects the hotfix (best case: stays on "spits on the ground") or misparses neighboring records (worst case: corrupts other emotes' display).

**Before merge:**

1. **Add `WriteUInt8(relationshipFlags)`** after the CString write per the snippet above.

2. **In-game verify** the hotfix actually applies:
   - Log into a character on the proxy
   - Target another player, `/spit` → should show "You spit on \<PlayerName\>." instead of "...on the ground."
   - If still showing "ground", the hotfix is being rejected — check proxy log for warnings; may need to also write the noninline EmotesTextID section.

3. **Negative test:** `/wave` and `/dance` to a target should still work normally (regression check that the EmotesTextData hotfix didn't break neighboring records).

4. **Untargeted /spit** (no selection, type `/spit`) → should still show "You spit on the ground." Target injection's `IsEmpty()` gate covers this; just confirm.

The target-injection logic (CurrentSelection + ChatHandler) is sound and worth keeping independent of the DB2 fix — it's the belt to the DB2 suspenders. If the DB2 hotfix doesn't work but the client still sends the target GUID, the targeted spit still propagates server-side; only the local chat text suffers.

PR #327, opened 2026-05-27
Fix: synthesize ControlUpdate restore when server omits it after BG exit

jameopotato:
## Summary

When a BG ends, the server sends `SMSG_CONTROL_UPDATE(HasControl=false)` to freeze the player for the scoreboard. After the world transfer back, the server should send `SMSG_CONTROL_UPDATE(HasControl=true)` to restore control. In rare cases the restore never arrives — permanent lockout (black action bars, can't move, must relog).

When `SMSG_MOVE_UNROOT` arrives for the local player within 5s of a world transfer and `LastObservedHasControl` is still false, synthesize `ControlUpdate(HasControl=true)` + run speed reassert.

## Evidence

602k-line JSONL log (C*******L, Kronos V, 2026-05-26): 4/5 BG exits had paired CONTROL_UPDATE false/true. 5th exit had false with no true for the remaining ~2000 lines until logout. Zero CMSG_MOVE_* from the player after the bug — total lockout confirmed.

## Safety

- **Flight paths** do not use CONTROL_UPDATE — they use the taxi handler's own control synthesis. PTR-verified.
- **Death/spirit release** does not use CONTROL_UPDATE. PTR-verified.
- **Login spawn burst** always sends its own CONTROL_UPDATE(true) before any UNROOT, so `LastObservedHasControl` is already true by the time this check runs.
- **5s time window** prevents triggering during normal gameplay root/unroot cycles (frost nova, hamstring root, etc.).
- **Redundant HasControl=true** is a no-op if the server already restored control normally.
- Emits `movement.control_restore.synthesized` structured event for monitoring.

## Open questions

- Root cause is likely server-side (Kronos). The trigger appears to be accumulated state from earlier transitions (mage portal + flight path chain before the BG), not the destination position. This fix is a defensive safety net, not a root cause fix.
- Should we also log the `HasControl` value in `HandleControlUpdate` for future diagnostics? Currently we track it in GameState but don't emit it to JSONL.

## Test plan

- Enter and exit multiple BGs. Verify no `movement.control_restore.synthesized` events in normal operation (server sends its own restore).
- If the bug reproduces (rare), verify the synthesized event fires and the player regains control without needing to relog.
- Check that flight path takeoff/landing is unaffected.
- Check that death/rez near water is unaffected.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -103,6 +103,10 @@ public sealed class GameSessionData
     // by hardcoding 7.0f. Default true: a player who has never been CC'd is
     // always in control, so the natural login state is "has control already."
     public bool LastObservedHasControl = true;
+    // JimsProxy (bg-exit-control-stuck): TickCount64 of the most recent
+    // HandleNewWorld inside the IsWaitingForNewWorld path. Gates the
+    // defensive ControlUpdate synth to a narrow post-transfer window.
+    public long WorldTransferTimestampMs;
     // JimsProxy (speed-stuck-after-fear-while-mounted): cached for reassert; see memory.
     public float LastKnownPlayerRunSpeed = 7.0f;
     // JimsProxy (speed-stuck-after-bg-end-while-mounted): deferred reassert flag; see memory.
--- a/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
@@ -319,6 +319,7 @@ public partial class WorldClient
         {
             GetSession().GameState.IsWaitingForNewWorld = false;
             GetSession().GameState.IsWaitingForWorldPortAck = true;
+            GetSession().GameState.WorldTransferTimestampMs = Environment.TickCount64;
 
             // JimsProxy (zone-transfer cast-state cleanup): the source map's server
             // state is torn down on transition (BG entry, instance change, zep arrival,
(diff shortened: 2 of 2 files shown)


jameopotato:
## Status: hold pending fresh repro logs after #340 ships

The C******L log (`jimsproxy-20260526-184810.jsonl`, 5 BG exits, 5th broken) re-analyzed against the Kronos dev's `HandleWorldPort` packet sequence overturns the original premise:

| Exit | Pre-LEAVE CONTROL | Post-LEAVE CONTROL | Post-LEAVE UNROOT |
|---|---|---|---|
| #1 (works) | -7733ms | +678ms | +1672ms |
| #2 (works) | -3271ms | +514ms | +1567ms |
| #3 (works) | -874ms | +724ms | +1684ms |
| #4 (works) | -2061ms | +625ms | +1597ms |
| **#5 (BROKEN)** | -1921ms | **+1043ms** | +2031ms |

The server *did* send a post-LEAVE `CONTROL_UPDATE` on the broken exit — opcode and size are identical to working exits. We can't tell from the JSONL whether it carried `HasControl=true` or `false` because the byte isn't logged today; PR #340 (just merged) fixes that. Once a new beta ships and someone reproduces the wedge, the new log entries will resolve the disposition.

**What's actually different on the broken exit:** destination position is underwater (Z=-39.5, map 0 Eastern Kingdoms) vs. working exits all above-ground on Kalimdor. Reporter independently confirmed underwater at teleport. Post-LEAVE the client sent 0 movement and 0 cast packets but kept pinging — consistent with a client-side input wedge, not a server-side missing restore.

Working hypothesis: the proxy's own deferred `movement.transport_clear.synthesized` packet conflicts with the destination's UpdateObject when destination carries swim flags from an underwater position. If true, this PR's defensive synth is redundant at best and could mask the actual proxy bug.

Converting to draft until a new repro log with the `HasControl` byte (PR #340) captured arrives. If that byte is `true`, this PR is unnecessary and the fix shape moves to addressing the deferred-synth at underwater destinations. If it's `false`, the defensive synth here may still help — but the proxy-synth path would still need a separate fix.

Related issues: #328, #329, #330, #331, #332 may share the same family (all `/reload`-fixable post-world-transfer wedges).

jameopotato:
Closing after re-evaluation against the redone world-transition model (evidence-first, 6 healthy captures) — **branch `fix/bg-exit-control-restore` is preserved**, this is a rethink, not a rejection of the problem.

Why closed: the premise (server omits the post-BG `CONTROL_UPDATE(true)`) rests on N=1 and was marked overturned in the 2026-06-27 session; and the fix is a wall-clock-gated fabrication of movement-control state — the same class as the transport-clear synth that is the one *proven* wedge in this bug family (#329/#331/#332). A false fire would inject a fabricated `SET_RUN_SPEED(MoveCounter=0)` into the move-counter chain, risking a new wedge to fix a rare one.

The standing rule for this family is **no fix code until an EU wedged capture**. #328 remains the open tracker; when the capture lands and shows what's actually missing on the wire, this branch's shape (including the `movement.control_restore.synthesized` diagnostic) is salvageable if the missing-restore theory turns out true.

PR #338, opened 2026-05-27
Erkagon misc fixes

erkagoon:
## Summary

This PR contains two small client/proxy compatibility fixes for Classic 1.14 clients connected to legacy 1.12 servers.

## Ranged auto-repeat after LoS failures

When using ranged auto-repeat spells such as `Shoot`, if the initial cast failed because the target was behind line of sight, the proxy cleared the current auto-repeat state after receiving the legacy spell failure.

On a native 1.12 client this behavior recovers once the target becomes visible again, but with the 1.14 client through HermesProxy the client did not send another cast request. The character stayed visually ready to attack, but ranged auto-repeat never resumed.

The fix keeps retryable ranged auto-repeat casts active for LoS/range-style failures and schedules a replay of the original held cast packet based on the player's ranged attack speed. Explicit `CMSG_CANCEL_AUTO_REPEAT_SPELL` still clears the state normally.

## Whisper target normalization

Some clients can produce reply targets in modern cross-realm format, for example `Player-Realm`. In the reported broken case the realm suffix was empty, producing `Player-`.

HermesProxy previously forwarded that target string directly to the legacy server, causing the server to look for a character literally named `Player-`, so the whisper failed.

The fix normalizes whisper targets before writing the legacy `CMSG_MESSAGECHAT` packet. If the target contains a realm separator (`-`), only the local player name before the separator is sent to the legacy server.

This is safe for 1.12-style servers because player names do not contain hyphens and the legacy whisper packet expects a local character name.

## Testing

- Built successfully with `dotnet build HermesProxy/HermesProxy.csproj -c Release`.
- Ranged auto-repeat LoS retry was tested in game and confirmed working.
- Whisper normalization is a defensive proxy-side fix for a reported issue that could not be reproduced locally.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -383,6 +383,7 @@ public sealed class GameSessionData
     private bool _gcdTimerHasFired;                     // true after OnGcdTimerElapsed runs; prevents orphaned holds
     private uint _lastFiredSpellId;                     // spell ID forwarded by the timer; used to drop same-spell late presses
     public Action<ClientCastRequest>? OnGcdHeldCastFire; // set by WorldSocket at attach time; invoked on a ThreadPool thread at GCD expiry
+    public Action<ClientCastRequest>? OnAutoRepeatRetry; // set by WorldSocket; refires Shoot/Auto Shot after retryable legacy failures
 
     // JimsProxy: cast-time spell queue. While a cast-time spell is in progress
     // (HasStartedNormalCast), presses are held here instead of dropped. Fired
@@ -1700,6 +1701,7 @@ public sealed class GameSessionData
         // path that normally nulls it.
         CurrentClientNextMeleeCast = null;
         CurrentClientAutoRepeatCast = null;
+        OnAutoRepeatRetry = null;
         return (normalCount, petCount, otherCount);
     }
 
--- a/HermesProxy/World/Client/PacketHandlers/ChatHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/ChatHandler.cs
@@ -563,7 +563,7 @@ public partial class WorldClient
                 packet.WriteCString(msg);
                 break;
             case ChatMessageTypeVanilla.Whisper:
-                packet.WriteCString(to);
+                packet.WriteCString(NormalizeLegacyWhisperTarget(to));
                 packet.WriteCString(msg);
                 break;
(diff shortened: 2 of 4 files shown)


jameopotato (changes requested):
Thanks for these two — re-reviewing now that some time has passed.

**Whisper target normalization:** this has since landed on `beta` independently via #348 ("Normalize legacy whisper targets") — same `NormalizeLegacyWhisperTarget` logic (strip at the first `-`). So that half of this PR is already shipped; please **rebase onto current `beta` to drop the duplicate** whisper commit.

**Ranged auto-repeat LoS/range retry:** this part is still unique and not in beta, but it needs some rework before it can merge:

1. **Rebase onto current beta.** This PR predates a month of cast/auto-repeat changes (#365, #367, #372, and especially **#374** which reworks observed auto-repeat). `HeldPacketForReplay` is now shared by more paths (GCD hold, low-latency forward), so the retry path needs re-checking against them.
2. **Reconsider the `Task.Delay(...).ContinueWith(...)` retry.** Firing a re-cast from an ad-hoc ThreadPool continuation is fragile — please route it through the existing timer infrastructure (the `OnGcdHeldCastFire` / session-owned timer pattern) instead, so teardown/cancellation is handled consistently. The `ReferenceEquals` guard + `try/catch` help, but the check-then-send isn't atomic with session teardown.
3. **Confirm clean stop in all teardown paths.** The unbounded retry-at-cadence relies entirely on `CurrentClientAutoRepeatCast` being nulled (you correctly added that to `HandleCancelAutoRepeatSpell`); please double-check reconnect / zone / death paths also stop it.
4. **In-game validation** of the retry cadence (a time-based re-fire is something we review carefully): confirm Auto Shot / wand Shoot resumes correctly when LoS/range clears, and that it doesn't double-fire or linger after the target dies or you cancel.

Happy to re-review once it's rebased down to just the auto-repeat retry with the above addressed.

jameopotato:
Closing as split/superseded — both halves of this PR are accounted for, with credit:

- **Whisper target normalization** → shipped via #348 (same normalization, in beta since v5.1.6-era and now stable in v5.1.7).
- **Ranged auto-repeat retry after LoS failures** → extracted to **#398** with your authorship preserved (cherry-picked onto current beta and reviewed at full rigor — verdict: hold for in-game testing, code is approve-quality). Thanks @erkagoon — the swing-cadence replay mirroring the native client's retry behavior was the right design, and it survived review unchanged.

Issue #330, opened 2026-05-27
Stuck after flight path landing: bars darkened, full input lockout, /reload fixes


jameopotato:
After a flight path landing (Eastern Plaguelands, departure unknown), the player can't move, can't rotate, action bars are darkened, but **can** cast Blink. `/reload` reliably fixes it.

## Evidence available
- JSONL log: `jimsproxy-20260526-230159.jsonl` (43 MB, beta.3, 12 taxi events, 0 world transfers, 6 control_updates).
- Multiple players have hit this. Reporter has had `/reload` fix it every time before.
- Reporter explicit: rotation also blocked.
- Reporter explicit: bars darkened.

## Note on Blink
Blink working does **not** prove `HasControl=true` — Blink has usable-while-impaired spell attrs that bypass control checks. So 'bars dark + Blink works' is internally consistent with `HasControl=false` on the client side.

## Hypothesis
`/reload` fixing it points to client-side state corruption (since /reload reinitializes the UI but does not change server state). Could be the taxi handler's own landing-side state synthesis missing a packet.

## Evidence needed
- A successful flight-path landing log for diff comparison.

## Related (possibly same family — not confirmed)
- #329 (mage teleport wedge)
- Other post-world-transfer movement wedges currently being tracked separately.

jameopotato:
## Likely root cause identified (2026-05-27)

Traced through the FP-user's session logs against the reporter's video. Strong evidence points to a specific mechanism — recording it here while it's fresh.

### Evidence: JSONL log `jimsproxy-20260526-230159.jsonl`

**Wedge window:** `t=1779829573997` (Hinterlands→EPL zone-sync at +254.8s into session) to `t=1779829675506` (second relog Login synth that cleared it). Duration ~101.5s.

**Reporter video verified against this slice:**
- Player lands in EPL via flight path ✓
- ~75 seconds of trying things ✓
- Tried Blink once ✓

**Client-side c2s packets during the 101s wedge window:**
| Activity | Count |
|---|---|
| `CMSG_MOVE_SET_FACING` (character rotation) | **0** |
| `CMSG_MOVE_START_FORWARD/STRAFE/BACKWARD/JUMP` (linear) | **0** |
| `CMSG_CAST_SPELL` | **1** (likely the Blink attempt) |
| `CMSG_LOGOUT_REQUEST` | yes at +75s after EPL entry |
| Passive packets (DB queries, pings, addon msgs) | normal |

[...]

PR #343, opened 2026-05-29
Restore movement after a taxi flight resumed across login (#330)

Mongrul:
**Problem** — If a player disconnects mid-taxi (or the server saves them mid-flight) and logs back in while still flying, the resumed flight completes but the 1.14 client stays a taxi-spline passenger: after landing you can't move — stuck/suspended in air, only Blink works (#330).

**Root cause** — A taxi resumed across login is delivered as the flight spline embedded in the player's CREATE block, not as per-leg `SMSG_ON_MONSTER_MOVE` packets. So `HandleMonsterMove` — which schedules the flight-end dismount (control restore + gravity + clear-fly + unroot) for a normal in-session taxi — is never invoked for a resumed flight. The vanilla server also sends no teleport or control-update at flight end (control is already `true` on the resume), so nothing tells the modern client to leave the passenger state.

**Fix** — The CREATE-block spline parse already recognizes the taxi spline (`Runmode|Flying`) and reads its remaining duration (`SplineTimeFull - SplineTime`). Use that to schedule the same dismount the normal path uses, mirroring the `HandleMonsterMove` dismount Task (CTS + atomic claim + cancel-on-disconnect). Gated to the player's own CREATE block on vanilla, so normal in-session flights (which flow through `HandleMonsterMove`) are untouched.

**Validation** — In-game on a 1.12 server: resumed a flight mid-air (81s flight, 27s elapsed -> 53.7s remaining); the dismount fired at the computed landing (+54.0s), the client ACKed gravity/unroot/unset-fly, reported spline-done, and regained movement. Separately confirmed a normal in-session taxi is unaffected (the resume path never fires; the existing dismount handles it). 509/509 tests pass.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/MovementHandler.cs
@@ -224,6 +224,62 @@ public partial class WorldClient
         }
     }
 
+    // JimsProxy (taxi-resume-control-stuck #330): packets the 1.14 client needs to leave the taxi/passenger state — control restore + gravity + clear-fly + unroot. Mirrors the dismount Task; used by the resumed-taxi SPLINE_ENABLED-clear path in ReadMovementUpdateBlock which never reaches HandleMonsterMove. See memory.
+    public void SendTaxiDismountRestore(WowGuid128 guid)
+    {
+        ControlUpdate control = new ControlUpdate();
+        control.Guid = guid;
+        control.HasControl = true;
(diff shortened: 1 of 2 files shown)


jameopotato:
Superseded by #351, merged to beta in v5.1.5-beta.5 (closes #330). #351 recreates this on an origin branch for CI and adds the dismount Task's finally (dispose CTS + clear bookkeeping) to match HandleMonsterMove. Authorship preserved. Thanks @Mongrul!

PR #360, opened 2026-06-05
JimsPlus v1.1.0: name dash, target cast-bar durations, handshake reconnect, versioning

jameopotato:
Full JimsPlus update (started as the cross-realm name-dash PR, expanded at request).

- **Name dash:** suppress the trailing cross-realm dash on emotes, mail, friends, inbox, and mail autofill, plus the AH "(*)" seller glyph (`NameDashFix.lua`, via `Ambiguate` + targeted hooks).
- **Target cast-bar durations:** fix the one-cast lag where an observed player's cast bar showed the *previous* cast's duration (e.g. rank-1 time on a rank-7 cast). The proxy sends `JP_CS` just after `SMSG_SPELL_START`, so the bar's COMBAT_LOG start reads the prior cast's `proxyCastTimes`. On `JP_CS`, re-apply the correct duration to a just-started player cast in place and consume it. **Verified in-game.**
- **Handshake reconnect:** re-send the JP sideband handshake on `PLAYER_ENTERING_WORLD` so cast-time/channel sidebands survive a reconnect (fresh proxy session) without `/reload`.
- **Versioning:** → `1.1.0`; load message + Options subtitle read the `.toc` Version via `GetAddOnMetadata` (single source of truth); removed the hardcoded `BUILD`.

### DIFF
--- a/Addons/JimsPlus/Core.lua
+++ b/Addons/JimsPlus/Core.lua
@@ -1,8 +1,8 @@
 local ADDON_NAME, namespace = ...
 
-namespace.BUILD = 20
+namespace.VERSION = (GetAddOnMetadata and GetAddOnMetadata(ADDON_NAME, "Version")) or "?"
 
-print("|cFF00FF00[JimsPlus]|r Core loaded (build " .. namespace.BUILD .. ")")
+print("|cFF00FF00[JimsPlus]|r v" .. namespace.VERSION .. " loaded")
 
 C_ChatInfo.RegisterAddonMessagePrefix("JP")
 
--- a/Addons/JimsPlus/JimsPlus.toc
+++ b/Addons/JimsPlus/JimsPlus.toc
@@ -4,7 +4,7 @@
 ## SavedVariables: JimsPlusDB, JimsPlusCastbarsDB
 ## SavedVariablesPerCharacter: JimsPlusCastbarsCharDB
 ## Author: JimsProxy Team (CastBars based on ClassicCastbars by Wardz)
-## Version: 1.0.1
+## Version: 1.1.0
 
 Core.lua
 Options.lua
--- a/Addons/JimsPlus/JimsPlus.toc
+++ b/Addons/JimsPlus/JimsPlus.toc
@@ -15,6 +15,7 @@ ZoneSync.lua
 MoonkinSound.lua
 AuctionsPaginationFix.lua
 MailNameFix.lua
+NameDashFix.lua
 KeyringClamp.lua
 castbars/PoolManager.lua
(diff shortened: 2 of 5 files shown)


jameopotato:
Follow-up surface found on a diag build: the **trade request** line still shows the trailing dash — `You have requested to trade with Budge-.` in chat.

The sections in this PR cover emote, friends, mail inbox/autofill, and AH `(*)`, but none hook `CHAT_MSG_SYSTEM`. The trade line is formatted from `ERR_TRADE_REQUEST_S = "You have requested to trade with %s."` with the raw target name, and that path never goes through `Ambiguate`, so it slips past every existing hook here.

**Recommendation** — add a small template-driven `CHAT_MSG_SYSTEM` filter to `NameDashFix.lua`, matching by global string (not raw English) so it stays locale-safe and no-ops if a constant is absent:

```lua
-- 7) Trade request system messages (ERR_TRADE_REQUEST_S etc.) — formatted with the raw
--    target name; bypasses Ambiguate.
local function templateToPattern(t)
    local M = "\001"
    t = t:gsub("%%s", M):gsub("([%(%)%.%%%+%-%*%?%[%]%^%$])", "%%%1"):gsub(M, "(.-)")
    return "^"..t.."$"
end
local rules = {}
for _, g in ipairs({ "ERR_TRADE_REQUEST_S", "ERR_TRADE_BLOCKED_S" }) do
    local t = _G[g]
    if type(t) == "string" and select(2, t:gsub("%%s","")) == 1 then  -- single %s = the name
        rules[#rules+1] = { tmpl = t, pat = templateToPattern(t) }
    end
end
ChatFrame_AddMessageEventFilter("CHAT_MSG_SYSTEM", function(_, _, msg, ...)
    if type(msg) ~= "string" or not msg:find("-", 1, true) then return end
    for _, r in ipairs(rules) do
        local name = msg:match(r.pat)
        if name then
            local clean = StripRealm(name)
            if clean ~= name then return false, r.tmpl:format(clean), ... end
        end
    end
end)
```

Safe because single-realm + letters-only 1.12 names mean any `-` in the captured name is the realm separator (same assumption the rest of this file relies on); it only fires on a dash + exact template match; and there's no re-entrancy (system filters don't re-fire on their own output).

Two things to confirm when implementing: (1) whether a transient **UIErrorsFrame** red-text copy also dashes — if so it needs a separate `hooksecurefunc(UIErrorsFrame, "AddMessage", …)`; (2) `ERR_TRADE_BLOCKED_S` availability on 1.14.2 (build-resolved → just no-ops if off). The rule table also extends trivially to other name-bearing system lines (e.g. duel challenge) if we want them.

PR #364, opened 2026-06-08
Refresh paladin Judgement debuff timer on the caster's melee hit

Mongrul:
## Problem
A paladin's Judgement debuff (Judgement of Light / Wisdom / the Crusader /
Justice) shows a duration timer on the target, but on the 1.14 client that
timer counted down from the initial judge and never reset — even though the
paladin's melee hits refresh it server-side. The proc/effect still worked;
only the displayed timer was wrong.

## Cause
Vanilla mechanic (cmangos `Unit.cpp` `DealMeleeDamage`): when a paladin
lands a melee hit, the server resets their OWN Judgement debuff (ALWAYS_HIT,
paladin-family, caster == attacker) on the victim to full duration via
`RefreshHolder()`.

The 1.12 server doesn't echo enemy-debuff durations (only self-cast auras get
`SMSG_UPDATE_AURA_DURATION`), and the refresh doesn't change the
`UNIT_FIELD_AURA` slot — so the proxy's aura-discovery loop never re-fires and
the modern client's timer just keeps draining.

## Fix
Synthesize the refresh, mirroring the server: on the local player's
`SMSG_ATTACKER_STATE_UPDATE`, if they're a paladin, scan the victim's aura
slots for the player's own Judgement debuff and re-emit a full-duration
`AuraUpdate` (via `SendAuraRefreshUpdate` with a new `forceFullRemaining` flag,
default false so existing recast callers are unchanged).

- Judgement debuff IDs verified against the 1.14.2 client `SpellName` table and
  cmangos; the heal/mana **proc** effects (20267/20268/…) and intermediate
  triggers (1826/5373) are excluded so only the debuff refreshes.
- JoL/JoW reuse the existing Improved-Judgement-aware duration table; JoC/JoJ
  durations come from `AuraDurations1.csv` (10 s).
- Melee-only (the cmangos refresh is in the melee-damage path); gated to a
  paladin local player.
- Diagnostic event `judgement.refresh.synth`.

## Testing
Clean build; `dotnet test` green — 536 total, including 22 new tests guarding
the Judgement ID set (debuff ranks included; proc/intermediate/cast IDs
excluded). Mechanic and IDs cross-checked against cmangos `Unit.cpp` and the
1.14.2 client `SpellName` table.

### DIFF
--- a/HermesProxy.Tests/World/JudgementDebuffSetTests.cs
+++ b/HermesProxy.Tests/World/JudgementDebuffSetTests.cs
@@ -1,0 +1,46 @@
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+public class JudgementDebuffSetTests
+{
+    // The set must contain every paladin Judgement DEBUFF aura (the ones applied to the
+    // target) so the local paladin's melee hit can synthesize the duration refresh.
+    [Theory]
+    [InlineData(20185u)] // Judgement of Light R1
+    [InlineData(20344u)] // Judgement of Light R2
+    [InlineData(20345u)] // Judgement of Light R3
+    [InlineData(20346u)] // Judgement of Light R4
+    [InlineData(20186u)] // Judgement of Wisdom R1
+    [InlineData(20354u)] // Judgement of Wisdom R2
+    [InlineData(20355u)] // Judgement of Wisdom R3
+    [InlineData(21183u)] // Judgement of the Crusader R1
+    [InlineData(20188u)] // Judgement of the Crusader R2
+    [InlineData(20300u)] // Judgement of the Crusader R3
+    [InlineData(20301u)] // Judgement of the Crusader R4
+    [InlineData(20302u)] // Judgement of the Crusader R5
+    [InlineData(20303u)] // Judgement of the Crusader R6
+    [InlineData(20184u)] // Judgement of Justice
+    public void IsJudgementDebuff_DebuffAura_ReturnsTrue(uint spellId)
+    {
+        Assert.True(GameData.IsJudgementDebuff(spellId));
+    }
+
+    // These must NOT be in the set: the JoL/JoW proc effects (heal/mana cast on the
+    // attacker, not a debuff on the target), the intermediate triggers, the Judgement
+    // cast itself, and unrelated spells. Including a proc id would refresh a non-debuff.
+    [Theory]
+    [InlineData(20267u)] // Judgement of Light R1 — heal proc effect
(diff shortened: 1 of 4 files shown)


jameopotato:
## Review: solid mechanism — gate the diagnostic + in-game verify before merge

Verified against `beta`:

- **Trigger correctly scoped** — `RefreshOwnJudgementOnVictim` is called inside `if (attack.AttackerGUID == CurrentPlayerGuid)` (CombatHandler L145), so it fires only on the local player's own swings and returns immediately for non-paladins. Per-hit overhead is near-zero except on the local paladin's swings. ✓
- **`durationFull` resolves for Judgements** — the enemy-debuff branch in `SendAuraRefreshUpdate` falls back cache → `TryGetTalentDuration` (Improved-Judgement-aware for JoL/JoW) → `GetAuraSpellDuration` (JoC/JoJ 10s), and `forceFullRemaining: true` sets `Remaining = durationFull`. ✓
- **ID set + tests** — the 22 tests correctly exclude the heal/mana procs (20267/20268…), intermediates (1826/5373), and the cast (20271). Good debuff-vs-proc distinction. ✓
- `forceFullRemaining` is additive (default `false`) — existing recast callers unchanged. ✓

### Requested change — gate the diagnostic
`judgement.refresh.synth` (CombatHandler) is **ungated** and fires on *every* paladin melee swing while a Judgement is up — a steady per-swing stream into the JSONL. Please gate it behind `Settings.DebugOutput`, consistent with how we treat hot-path logging.

### Before merge
1. Gate the diagnostic (above).
2. **In-game verify** (not done yet): with each seal up (JoL / JoW / JoC / JoJ), Judge the target, then melee — confirm the debuff timer resets to full on each swing with the right duration. This validates the ID set end-to-end too.
3. Recreate on an origin branch (authorship preserved) so CI runs — fork PRs don't get CI here.

### Minor (non-blocking)
The untracked-caster heuristic (`caster == default` treated as ours) could cosmetically refresh another paladin's *untracked* Judgement on your own client — rare and harmless; fine as-is.

**Verdict:** hold for testing — mechanism is correct and well-unit-tested; the diagnostic gate + an in-game pass are the gates to merge.

Mongrul:
This was tested in game. 

judgement.refresh.synth` is now wrapped in `if (Framework.Settings.DebugOutput)`,
  matching how we gate the other per-event hot-path diagnostics (e.g. `cast.received`).
  Only the `Log.Event` is gated — the `SendAuraRefreshUpdate(..., forceFullRemaining: true)`
  behavior is unchanged, so the refresh still fires on every qualifying swing; it just no
  longer streams to the JSONL unless DebugOutput is on.

PR #365, opened 2026-06-10
Fix engineering-malfunction cast jam (Goblin Mortar backfire -> can't cast)

jameopotato:
## Summary
Engineering devices that can malfunction (Goblin Mortar confirmed) leave the player **unable to cast** until relog. The device's item-use cast is forwarded to the server, but on a malfunction the server casts a *substitute* spell instead (Goblin Mortar → **Malfunction Explosion 13261**) whose `SMSG_CAST_FAILED` arrives with `status != 2` and is discarded by `HandleCastFailed`. The original item-use cast (13237) then never gets a `SPELL_START` or its own failure, so it sits **forwarded-unstarted** in `PendingNormalCasts` forever — `HasForwardedPendingCast()` stays true and every later press is held (`spell.held_pending` with no release).

Root confirmed by hex-dumping the backfire's CAST_FAILED: it is for **13261, status 0**, not the mortar's 13237.

## Fix
On the discarded `status != 2` path, evict the first forwarded-unstarted **item-use** cast (`ItemGUID` set, `SpellId != the discarded spell`) and release its button silently. Instant, deterministic (fires on the real malfunction packet), **no timeout**.
- **Item-use-gated** → normal `CMSG_CAST_SPELL` casts can never be evicted.
- **Procs can't trigger it** (they `GO`, not `CAST_FAILED`).
- Emits `cast.item_use_orphan_evicted`.

## Tests
**519/519** (514 baseline + 5 new in `ItemUseMalfunctionEvictionTests`).

## In-game verification: PENDING
Trigger one Goblin Mortar backfire → expect `cast.item_use_orphan_evicted (evicted=13237, trigger=13261)` and `held_pending = 0` (no jam). Optional: a Rocket Boots / Sapper malfunction to confirm the same CAST_FAILED shape.

### DIFF
--- a/HermesProxy.Tests/World/ItemUseMalfunctionEvictionTests.cs
+++ b/HermesProxy.Tests/World/ItemUseMalfunctionEvictionTests.cs
@@ -1,0 +1,86 @@
+using HermesProxy;
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (engineering-malfunction jam): TryEvictForwardedItemUseCast clears a forwarded-but-
+// unstarted ITEM-use cast that a server-side substitute spell preempted — e.g. Goblin Mortar (13237)
+// orphaned when the malfunction substitutes Malfunction Explosion (13261), whose CAST_FAILED arrives
+// status != 2 and is discarded, so the mortar's forwarded cast never starts/fails and jams
+// HasForwardedPendingCast() forever. These pin the guards: item-use only, no self-eviction on the
+// item's own spell, and started/normal casts are left intact.
+public class ItemUseMalfunctionEvictionTests
+{
+    private static GameSessionData NewSession() => GameSessionData.CreateForTesting();
+
+    private static ClientCastRequest ItemUse(uint spellId, bool started = false) => new()
+    {
+        SpellId = spellId,
+        HasStarted = started,
+        ItemGUID = new WowGuid128(1, 2), // non-empty => item-use cast
+    };
+
+    private static ClientCastRequest SpellCast(uint spellId, bool started = false) => new()
+    {
+        SpellId = spellId,
+        HasStarted = started,
+        // ItemGUID left default (empty) => normal CMSG_CAST_SPELL
+    };
+
+    [Fact]
+    public void EvictsForwardedItemUse_PreemptedBySubstitute()
+    {
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(ItemUse(13237)); // Goblin Mortar, forwarded-unstarted
+
+        Assert.True(s.TryEvictForwardedItemUseCast(13261, out var orphan)); // 13261 = Malfunction Explosion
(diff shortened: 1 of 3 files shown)


jameopotato:
## Review: narrow the eviction to a known-malfunction set

Mechanism is correct and the bug is real — verified against `beta`:

- `spellId` at the `status != 2` block is the discarded **substitute's** id, and that block is vanilla-only (`RemovedInVersion(V2_0_1_6180)`), so the trigger is right.
- `SendCastRequestFailed(orphan, false, DontReport)` = player cast, silent button release — correct.
- Guards are tight: evicts only the first `!HasStarted && !ItemGUID.IsEmpty() && SpellId != trigger`; normal `CMSG_CAST_SPELL` casts and started casts are never touched. The 5 tests pin these well.
- Composes cleanly with #362 (GO-path `PlayerForwardedCastIds` recall vs. this CAST_FAILED-path `PendingNormalCasts` eviction — different packets/structures, no double-handling).

### Concern: false-positive eviction on unrelated item use
The eviction fires on **any** vanilla `status != 2` CAST_FAILED whenever **any** forwarded-unstarted item-use cast is queued. The victim guard is just `!ItemGUID.IsEmpty()`, so the victim can be a **potion/trinket** in its brief (~RTT) unstarted window — not only a malfunction device. If an unrelated spell throws a `status != 2` right as you use a potion/trinket, that item's cast gets evicted. Recoverable (the item still fires; the button/cooldown just glitches), but it can hit common item use.

### Recommendation
Gate the eviction to a known-malfunction set rather than "any `status != 2` + any item-use orphan." Simplest: only run the eviction when the **trigger** is a known malfunction substitute — seed the set with **Malfunction Explosion `13261`** (Goblin Mortar) and expand as more are confirmed.

Repro notes: only **Goblin Mortar** reproduces the jam. **Rocket Boots** didn't backfire in 25 casts; **Goblin Sapper** only self-damages (no substitute-spell jam observed) — so neither belongs in the set yet.

This keeps the confirmed Mortar fix, removes the false-positive entirely, stays deterministic (no timeout), and is trivially expandable. (If Mortar turns out to roll substitutes other than 13261, gate on the **device** instead — orphan `SpellId == 13237` with a broad trigger.) Add a short code comment noting the set is seeded with Mortar and expected to grow.

**Verdict:** narrow to the Mortar set before merge; otherwise solid.

jameopotato:
Re-reviewed the update — this fully addresses the concern. The `MalfunctionSubstituteToDevice` map gates on **both** the known substitute trigger (`13261`) **and** the mapped device victim (`13237`), so an unrelated `status != 2` (or `13261` while a different item is in-flight) evicts nothing — the false-positive surface is gone. The early `FrozenDictionary` lookup also skips the queue drain on the common path, and `DoesNotEvict_OnUnrelatedTrigger` / `DoesNotEvict_WhenOrphanIsNotTheMappedDevice` pin exactly that. Mortar behavior is unchanged, so the in-game verification still holds.

LGTM — approving. The seed-and-expand comment is a good call for adding devices later.

PR #367, opened 2026-06-10
Release predicted pet-button GCD on failed pet cast (warlock imp Firebolt LoS)

Mongrul:
## Problem
Pressing a pet ability the pet can't use — e.g. a warlock's imp Firebolt while the imp is out of line of sight — plays the failure sound but runs a **full 1.5s GCD sweep** on the pet action button, even though nothing was cast. Spamming the button just re-triggers the bogus sweep.

## Root cause
A pet-bar press arrives as `CMSG_PET_ACTION`, which creates no `PendingPetCast`. The modern 1.14 client locally predicts a GCD on the pet button at press time. On Kronos the failure is reported via `SMSG_SPELL_FAILED_OTHER` (plus a trailing `SMSG_CAST_FAILED` that `HandleCastFailed` drops, since there's no matching normal cast) — **never `SMSG_PET_CAST_FAILED`**, which is the only packet that releases the client's predicted pet-button GCD. So the sweep ran to completion.

`HandlePetCastFailed` already synthesizes a release for this exact "no pending pet cast" case, but only when the failure arrives as `SMSG_PET_CAST_FAILED` — which Kronos doesn't send for pet-bar actions.

## Fix
Synthesize `SMSG_PET_CAST_FAILED(DontReport)` for own-pet failures via a shared `SendPetGcdRelease` helper:
- Called from `HandleSpellFailedOther` (the path Kronos uses) and defensively from `HandleSpellFailure`.
- `DontReport` is silent, so the existing `SpellFailure` still plays the error sound — no double sound.
- Also fired inside the `SMSG_SPELL_FAILED_OTHER` 500 ms dedup branch so rapid double-clicks each get released (the release is cheap/idempotent; only the `CancelSpellVisual`/`SpellFailure` storm needs throttling).
- All three call sites are gated to **unqueued** presses (`!HasQueuedPetCast`) so a queued `CMSG_PET_CAST_SPELL` — which receives its own trailing real-reason `SMSG_PET_CAST_FAILED` — isn't double-failed.

Only fires for the player's own pet (`CurrentPetGuid == casterUnit`); player and mob/enemy cast paths are untouched.

## Testing
- Validated in-game on Twinstar (Kronos) with a warlock imp: Firebolt out of LoS no longer GCD-sweeps; double-click spam is clean after the dedup-branch fix.
- Log analysis (33 presses = 33 failures, all `CMSG_PET_ACTION`): release fires once per failure, no flood; no overlap with successful casts; no double-send (Kronos never sends `SMSG_SPELL_FAILURE`/`SMSG_PET_CAST_FAILED` for pets).
- Builds clean (0 warnings, 0 errors).

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -624,6 +624,20 @@ public partial class WorldClient
         SendPacketToClient(failed);
     }
 
+    // JimsProxy (warlock-pet-gcd-on-failure): synth SMSG_PET_CAST_FAILED(DontReport) so the modern client releases its predicted pet-button GCD sweep for a failed pet cast. Silent (the fizzle already played); the client matches by SpellID since CMSG_PET_ACTION presses carry no client CastID.
+    void SendPetGcdRelease(uint spellId, WowGuid128 castId)
+    {
+        PetCastFailed petFailed = new PetCastFailed();
+        petFailed.SpellID = spellId;
+        petFailed.Reason = (uint)SpellCastResultClassic.DontReport;
+        petFailed.CastID = castId;
+        SendPacketToClient(petFailed);
+    }
+
+    // JimsProxy (warlock-pet-gcd-on-failure): true if a CMSG_PET_CAST_SPELL for this spell is still queued — those get their own trailing SMSG_PET_CAST_FAILED, so the synthesized GCD release is gated to unqueued CMSG_PET_ACTION presses to avoid a double-fail.
+    bool HasQueuedPetCast(uint spellId) =>
+        GetSession().GameState.PendingPetCasts.Any(c => c.SpellId == spellId || (c.LegacySpellId != 0 && c.LegacySpellId == spellId));
+
     [PacketHandler(Opcode.SMSG_SPELL_FAILED_OTHER)]
     void HandleSpellFailedOther(WorldPacket packet)
     {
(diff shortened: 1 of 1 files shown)


jameopotato:
## Review: clean, in-game-verified pet-GCD release — approve

Verified against `beta`:
- **Own-pet only** — `casterIsPet = CurrentPetGuid == casterUnit` (L718 / L1064); the dedup branch checks it inline. Never fires for enemy/other pets. ✓
- **Mirrors the established release** — `SendPetGcdRelease` (PetCastFailed + DontReport + deterministic/tracked CastID) is exactly what `HandlePetCastFailed` already synthesizes for the no-pending case (L515-517). This just routes it to the FAILED_OTHER / SPELL_FAILURE paths Kronos actually uses. ✓
- **`castId` always defined** — no-pending path resolves via tracked CastIDs → deterministic seed (L713); the dedup branch computes the same seed; the client matches by SpellID anyway. ✓
- **`!HasQueuedPetCast` gating** — a queued `CMSG_PET_CAST_SPELL` gets its own trailing `PetCastFailed`, so no double-fail. ✓
- **No double-sound** (`DontReport` silent; `SpellFailure` still plays the fizzle); **no double-send** on Kronos (only FAILED_OTHER fires; the SPELL_FAILURE site is defensive; a double-release is idempotent). ✓
- **No new diagnostics** — only a `sentPetCastFailed` field on existing logs. ✓

Minor (non-blocking): the dedup-branch release also fires during a pet auto-cast retry storm (5+/sec) — harmless no-ops (auto-casts carry no player-predicted GCD) and cheap/idempotent.

**Verdict: approve.** Purely additive, well-gated, mirrors the proven release, in-game verified (Twinstar warlock imp) + log-analyzed. Merge path: recreate on an origin branch (authorship preserved) so CI runs green before the admin-merge.

PR #368, opened 2026-06-11
Fix Bagnon keyring sort loop (bag-family guard) + JimsPlus custom sort order

jameopotato:
## Problem

Bagnon auto-sort loops forever spamming "That item cannot go in that container" (recurrence of the bug class #357 fixed). Log forensics on `jimsproxy-20260611-091806.jsonl`:

- **Zero** `SMSG_INVENTORY_CHANGE_FAILURE` from Kronos in the whole session — the error toasts were the proxy's own `SendKeyringBounce`, which is invisible in the JSONL (`packet.translated` fires even for blocked packets; synthesized s2c packets are never logged).
- Final burst: 54 identical CMSG_SWAP_ITEM at an exact 50ms cadence with total server silence; the loop re-armed 7s after a `/reload` and only died on the second `/reload`. The character's keyring was **empty** afterward — the move never landed.

## Root cause

The proxy contradicts itself. Kronos marks some non-Key-class items keyring-able via the vanilla keyring **bag-family** bit (0x8) — the repo's own Kronos overlay ships `18645 Alarm-O-Bot` (class 7) with `BagFamily=256`, and `ItemTemplate.ReadFromLegacyPacket` translates the bit to modern 0x100 and advertises it to the client via ItemSparse hotfixes. So the client reports `GetItemFamily(id)==256`, Bagnon (correctly) sorts the item into the keyring — and the #357 guard, judging by `Class != 13` **only**, bounces every attempt. Bagnon has no retry cap: identical move every 50ms until `/reload`. The vanilla server itself accepts keyring placement by bag family, not class, so the forwarded move would have succeeded.

## Fix

**Proxy** (`World/Server/PacketHandlers/ItemHandler.cs`):
- `IsKnownNonKey`: keyring-eligible = `Class == 13` **or** `BagFamily & 0x100` — same rule the server enforces and the proxy advertises.
- Same-itemId endpoint moves (stack merges) are exempt — the old guard made an orphaned stack impossible to merge out of the keyring onto an existing bag stack.
- Every bounce now emits `item.keyring_bounce` (containers/slots/item ids) so this is diagnosable from the JSONL next time.
- Handlers now Resolve → Block → Remember: blocked moves no longer clobber `LastMoveItem0/1`, so the #357 empty-GUID backfill can't be poisoned by bounce spam during sorts.

**JimsPlus 1.2.0** (`Addons/JimsPlus/BagnonSort.lua`, new):
- **Loop breaker** (always on): hooks Bagnon's Wildpants `Sorting` module at runtime (no Bagnon files modified; type-checked no-op if Bagnon is absent or restructured). After 3 identical attempted moves in one sort run the item is benched for the run (`sorted=true`, `stack=nil` in the in-memory model), so *any* future rejection source — old proxy builds, server-side rejections, data mismatches the proxy can't see — degrades to three error toasts and a completed sort instead of an infinite loop. Convergence traced against Bagnon 9.1.5.
- **Curated sort order** (Options toggle, default on): hearthstone first, then profession/gathering tools, quest items, soulbound (BoP) gear, other gear, consumables, everything else, junk last; ties fall back to Bagnon's original comparator. Applies to bank sort too.
- Bag Audit now leaves keyring-bag-family non-keys (e.g. Alarm-O-Bot) in place, consistent with the proxy rule.

## Verification

- `dotnet build` clean; `dotnet test` 535/535 passing.
- Multi-agent adversarial review: hook targets, loop-breaker convergence, comparator strict-weak-ordering, and 1.14.2 API availability verified against the installed Bagnon 9.1.5 source; minor hardening findings applied (late-load `ADDON_LOADED` fallback, docs for bank-sort scope and the future-`SortBags` bypass).
- Root-cause chain verified in repo data: `CSV/Hotfix/ItemSparse1.kronos.csv` row 18645 is the only BagFamily-drift row (0 → 256) among all 585 overlay rows; `ItemTemplate.cs` stores the translated mask.

### DIFF
--- a/Addons/JimsPlus/BagnonSort.lua
+++ b/Addons/JimsPlus/BagnonSort.lua
@@ -1,0 +1,177 @@
+-- JimsPlus BagnonSort
+-- Two hooks into Bagnon's (Wildpants) client-side bag sorter. Bagnon's own files
+-- are never modified, so this survives Bagnon updates, and both hooks no-op when
+-- Bagnon isn't installed.
+--
+-- 1. LOOP BREAKER (always on): Bagnon's sorter has no retry cap or backoff — a
+--    move that never takes effect is recomputed and re-sent every 50ms forever
+--    (Wildpants/api/sorting.lua: Iterate -> Move -> Delay(0.05, 'Run')). Any move
+--    the proxy or server keeps rejecting becomes endless "That item cannot go in
+--    that container" spam until /reload. After 3 identical attempted moves in one
+--    sort run, the item is blacklisted for the rest of the run (marked sorted and
+--    made unstackable in the in-memory model) so the sort converges and finishes.
+--
+-- 2. CUSTOM SORT ORDER (Options toggle, default on): replaces the sort comparator
+--    with a category ranking — permanent fixtures (hearthstone), then profession/
+--    gathering tools, quest items, soulbound (BoP) gear, other gear, consumables,
+--    everything else, junk last. Within a category Bagnon's original ordering
+--    applies. Lower rank lands closer to backpack slot 1. The Sorting module is
+--    shared, so the order also applies to Bagnon's bank (and guild bank) sort.
+--
+-- Note: Bagnon prefers the server-side SortBags() when the client provides it
+-- (Wildpants/classes/inventory.lua, serverSort default true). 1.14.2 has no
+-- SortBags, so the client-side sorter — and these hooks — always run for the
+-- inventory; if a future client build adds SortBags, both hooks silently stop
+-- applying to the inventory sort and this file needs revisiting.
+
+local _, namespace = ...
+
+local FIXTURES = {
+    [6948] = true, -- Hearthstone
+}
+
+local TOOLS = {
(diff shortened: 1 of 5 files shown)


jameopotato:
⚠️ **Merge-ordering note (paired with #369).** Both this PR and #369 bump `Addons/JimsPlus/JimsPlus.toc` `## Version:` — this one to **1.2.0**, #369 to **1.1.1** — so whichever merges *second* will conflict on the `.toc` (the version line, plus this PR's added `BagnonSort.lua` in the load list). It's the **only** conflict: `BagnonSort.lua` and #369's `NameDashFix.lua` don't overlap.

Recommended: merge **#369 first**, then this one; resolve the `.toc` to `## Version: 1.2.0` and **keep `BagnonSort.lua` in the load list**. End state = JimsPlus 1.2.0 carrying both the keyring guard/loop fix (here) and #369's Promote/NameDash fix, bundled with this PR's proxy guard fix.

Do **not** cut a version with #369 merged but this still open — that ships JimsPlus without the keyring guard fix (the sort-loop reproduces).

PR #369, opened 2026-06-12
Fix JimsPlus guild Promote block: drop global Ambiguate taint wrap

jameopotato:
## Summary
- JimsPlus v1.1.0's `NameDashFix` replaced the global `Ambiguate()` to strip the realm dash "broadly". `Ambiguate` is read inside secure UI paths (the unit/guild right-click dropdown), so the insecure replacement tainted that path and blocked **guild Promote** with `ADDON_ACTION_FORBIDDEN`.
- There is no taint-safe way to wrap a secure transform function (`hooksecurefunc` can't change a return value), so this removes the global wrap entirely.
- Chat sender names are re-covered with an **author-only** `ChatFrame_AddMessageEventFilter` for the player chat events (say/yell/whisper/+inform/guild/officer/party/+leader/raid/+leader/+warning/channel). Author-only so user-typed message bodies are never rewritten.
- Mail, Friends, inbox, mail autocomplete, and the AH `(*)` indicator already use their own taint-safe handlers (`hooksecurefunc` + fontstring rewrites, editbox hooks) and are unchanged.
- `.toc` bumped 1.1.0 -> 1.1.1.

## Test plan
- [x] Verified in-game 2026-06-12: right-click guild member -> **Promote works** with JimsPlus enabled.
- [ ] `/console taintLog 1` -> no JimsPlus taint on Ambiguate paths.
- [ ] Whisper / guild / party / channel sender names render with no trailing `-`.
- [ ] Emote names ("X laughs.") still strip the dash (unchanged §2 path).

## Notes
- Out of scope (deferred, verify-first): unit-frame and player-tooltip name dashes — not in the original reported set; confirm they actually show the dash before adding targeted passes.
- Aside: two other global wraps live in JimsPlus (`SendMail` in MailNameFix, `GetContainerNumSlots`/`...FreeSlots` in KeyringClamp); neither is on a protected-action secure path today, but worth a separate taint audit.

### DIFF
--- a/Addons/JimsPlus/JimsPlus.toc
+++ b/Addons/JimsPlus/JimsPlus.toc
@@ -4,7 +4,7 @@
 ## SavedVariables: JimsPlusDB, JimsPlusCastbarsDB
 ## SavedVariablesPerCharacter: JimsPlusCastbarsCharDB
 ## Author: JimsProxy Team (CastBars based on ClassicCastbars by Wardz)
-## Version: 1.1.0
+## Version: 1.1.1
 
 Core.lua
 Options.lua
--- a/Addons/JimsPlus/NameDashFix.lua
+++ b/Addons/JimsPlus/NameDashFix.lua
@@ -18,15 +18,35 @@ local function StripRealm(name)
     return name
 end
 
--- 1) Broad coverage. Ambiguate() is the function most name-display code routes a
---    realm-qualified name through; wrapping it drops the realm everywhere at once
---    (unit frames, many chat paths, tooltips, etc.). Safe here -- single realm, so the
---    "short" form is always what we want and the realm suffix is always empty.
-if type(Ambiguate) == "function" then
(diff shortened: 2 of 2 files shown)


jameopotato:
⚠️ **Merge-ordering note (paired with #368).** Both this PR and #368 bump `Addons/JimsPlus/JimsPlus.toc` `## Version:` — this one to **1.1.1**, #368 to **1.2.0** — so whichever merges *second* will conflict on the `.toc`. It's the **only** conflict: this PR's `NameDashFix.lua` and #368's `BagnonSort.lua` don't overlap.

Recommended: merge **this first**, then #368; resolve the `.toc` to `## Version: 1.2.0` and **keep `BagnonSort.lua` listed**. This PR's NameDashFix change rides along untouched.

Make sure **#368 also lands before cutting a version** — otherwise the keyring sort-loop guard fix won't ship.

PR #371, opened 2026-06-13
Drop stray trailing movement for just-destroyed / out-of-range units

Mongrul:
Vanilla broadcasts a moving unit's trailing `MSG_MOVE_*` (and `MONSTER_MOVE`) at map-level distance *after* it has already sent the per-object-visibility destroy for that unit. The proxy relayed that stray movement, so the modern client re-created and re-ghosted the just-destroyed unit "running in place" at its last position until the player re-approached and a real `CreateObject` arrived.

This tracks guids that were just destroyed or went out-of-range (`SMSG_DESTROY_OBJECT` and the FarObjects/out-of-range path) in a short-TTL set, and drops any trailing movement for them in `HandleMovementMessages` / `HandleMonsterMove` until a `CreateObject` re-creates the unit and clears the mark. The set is TTL-bounded (10s) with an opportunistic sweep so a long spawn/despawn session can't grow it unbounded. The local player is exempt.

The `movement.dropped_stray_after_destroy` diagnostic is gated behind `DebugOutput`.

> Note: lightly overlaps #370 in the `UpdateHandler` out-of-range block — trivial textual conflict if both merge.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -330,6 +330,41 @@ public sealed class GameSessionData
     // Forwarding the first one is sufficient — the client cancels the cast bar / visual and
     // subsequent same-cast failures add no new state.
     public Dictionary<(WowGuid128 Caster, uint SpellId), long> RecentlyForwardedSpellFailedOther = new();
+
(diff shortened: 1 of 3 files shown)


jameopotato:
Recreated on origin and landed in beta as b97e338 for v5.1.7-beta.1 — cherry-picked to resolve the UpdateHandler overlap with #370, your authorship preserved. Thanks @Mongrul! Closing since the commit is now on beta.

PR #374, opened 2026-06-15
Retract observed shooters' latched ranged-aim once they stop firing

Mongrul:
## Problem

Watching another unit auto-repeat (Auto Shot 75 / wand Shoot 5019) on a modern 1.14 client draws and holds the bow/wand and **never lowers it** after they stop. Forwarding the observed `SPELL_START` latches the 1.14 client's intrinsic auto-repeat aim (spell 75 is `SPELL_ATTR2_AUTO_REPEAT` with no client `SpellVisual`), and vanilla never broadcasts a "stopped" for *other* units, so nothing retracts it. Your own character is fine — `SMSG_SPELL_FAILURE` lowers the local aim.

The prior hold-and-replay (#352) only guarded an *aborted* shot (an unpaired START with no GO). A *completed* START+GO latches the aim just the same, so an observed shooter stayed stuck after the last shot of every series.

## Fix

A **hybrid retract** that synthesizes `SMSG_CANCEL_AUTO_REPEAT` carrying **that unit's GUID** (the modern `CancelAutoRepeat` packet is per-GUID; the legacy handler only ever cancels the local player, so a per-shooter cancel had never actually been tried). Each shot latches the shooter against the unit it's shooting, then two mechanisms lower the weapon:

**Deterministic stop edges — instant retract** from signals the wire already gives us:
- **Target dies** — `PARTY_KILL_LOG`, or the target's health hits 0 (covers DoT / non-party / environment deaths). Terminal-proof: a corpse can't be shot, so the retract can't be contradicted.
- **Shooter dies.**
- **Shooter moves** — translational-start opcodes only (`START_FORWARD/BACKWARD/STRAFE/JUMP/SWIM`); turn/facing/heartbeat/stop are excluded so a shooter that only pivots to track its target is never falsely retracted.
- **Shooter retargets or clears its target** (`UNIT_FIELD_TARGET` change).

**Quiescence sweep — catch-all** for the case no edge covers: a shooter that stops *with its target still alive and stands still* (out of ammo, `/stopattack`, LoS break, **target dummy**). A 500 ms sweep lowers the weapon once it goes quiet past 4 s.

Properties:
- **No mid-series flicker** — the 4 s threshold sits above the slowest bow cadence + jitter; every shot refreshes it.
- **Self-healing** — a premature retract is harmless: the next shot's START re-raises the aim.
- **No splash to you** — tracking excludes the local player and pet, so your own auto-shot is never retracted.
- Sweep callback runs off the lock; the timer self-disarms when the table empties and is cleared on reconnect (`ResetInFlightCastState`). Per-shooter target state lives in one struct (`ObservedShooterState`).
- The `start_held` / `start_replayed` / `remote_cancel` diagnostics are gated behind `DebugOutput`.

> Why both? An edge-driven retract alone can't cover a shooter that stops with its target alive and stationary — there's no packet edge for it, so the sweep is load-bearing. The edges just make the common cases (kill / move / retarget) lower instantly instead of waiting out the 4 s window.

## Testing

**In-game (Twinstar/Kronos 1.12 ↔ 1.14 client):**
- **Quiescence (target dummy):** an observer watching a hunter stop firing at a target dummy — target stays alive, shooter stationary — lowers the bow ~4 s after the last shot.
- **Edges:** killing the mob, or the shooter moving / retargeting, lowers the bow **instantly**.
- Original sweep validation still holds: 6 stop/start Auto Shot cycles each sheathed cleanly 4.0–4.5 s after the last shot, no flicker mid-series (including across a haste change from ~2.55 s to ~1.95 s cadence).
- **No splash:** confirmed on a hunter observer that firing your *own* Auto Shot while watching another shooter stop does not interrupt your shots when their cancel fires.

**Automated:** full suite green, including **14** `ObservedAutoRepeatTests` (9 sweep + 5 edge) covering target-death fan-out, retarget vs. same-target, per-shooter isolation, the quiet window and refresh-keeps-alive across slow cadence, idempotency, and teardown.

### DIFF
--- a/HermesProxy.Tests/World/ObservedAutoRepeatTests.cs
+++ b/HermesProxy.Tests/World/ObservedAutoRepeatTests.cs
@@ -8,9 +8,10 @@ using Xunit;
 
 namespace HermesProxy.Tests.World;
 
-// JimsProxy (observed-bow retract): quiescence-sweep that lowers an observed shooter's latched
-// ranged-aim (Auto Shot 75 / wand Shoot 5019) by firing SMSG_CANCEL_AUTO_REPEAT once they stop.
-// The sweep clock is injected so the 3s quiet window is exercised deterministically with no waits.
+// JimsProxy (observed-bow retract): the retract that lowers an observed shooter's latched ranged-aim
+// (Auto Shot 75 / wand Shoot 5019) is driven by DETERMINISTIC stop edges — target death, shooter
+// death/move, retarget — instead of a time-based quiescence sweep. These cover the GameState tracking
+// the WorldClient packet handlers key off; no clock, so every case is synchronous.
 public class ObservedAutoRepeatTests
 {
     static ObservedAutoRepeatTests()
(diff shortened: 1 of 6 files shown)


jameopotato (approved):
**Approving.** Reviewed at full rigor (spell-system files). The "hold by default" bar for spell-system changes is cleared here because the whole feature is gated to non-local units (`caster != CurrentPlayerGuid && caster != CurrentPetGuid`), so it cannot affect your own casting.

**Verified against the code:**
- `CancelAutoRepeat` is genuinely per-GUID (`SpellPackets.cs` → `WritePackedGuid128(Guid)`), so the synthesized cancel targets the specific shooter — the mechanism this fix rests on.
- `ClearAllObservedAutoRepeat()` lands in `ResetInFlightCastState()`, which runs on unplanned reconnect (`GlobalSessionData.cs`), so tracking is cleared, the sweep timer disposed, and `OnObservedAutoRepeatExpire` nulled — the next observed shot rebinds the callback to the live WorldClient via `??=`.
- Note fires on every observed auto-repeat `SPELL_GO` (including bare tick GOs), so the 4s quiet window refreshes mid-series; local player/pet are excluded; the timer self-disarms when the table empties; all three diagnostics are `DebugOutput`-gated.
- 9 deterministic unit tests (injected clock) plus the in-game 6-cycle verification cover the behavior well.

**Non-blocking suggestions:**
1. `OnObservedAutoRepeatExpire ??= SendObservedAutoRepeatCancel` (in `HandleSpellGo`) is assigned off-lock from the ReceiveLoop while `ClearAllObservedAutoRepeat` nulls it under `_observedAutoRepeatLock`. The reference write is atomic so it's safe, but assigning it under the lock would keep the bind/clear fully ordered.
2. Worth one in-game check before heavy play: a **slowed** shooter firing >4s apart (heavy ranged slow) will trip a mid-series retract that the next shot re-raises — self-healing, but confirm the brief flicker is acceptable. `ObservedAutoRepeatQuietMs` (4000) has ~1s margin over the slowest base bow cadence.

Not merging — approval only, per request.

jameopotato (changes requested):
Thanks for this — the per-shooter tracking, the lazy self-disarming sweep timer, the off-lock cancel callback, and the 9 tests with an injectable clock are all solid, and the root-cause writeup (forwarding an observed `SPELL_START` latches the 1.14 client's intrinsic auto-repeat aim, and vanilla never broadcasts a stop for *other* units) matches what #352 established.

The reason we're holding it: the retract is driven by a **time-based quiescence sweep** (`ObservedAutoRepeatQuietMs=4000` + a 500ms `Timer`). We've been holding a hard line against time-based cleanup for cast/visual state — it tends to false-positive under latency — so we'd like to avoid introducing a timeout sweep here even though observed units genuinely give no stop signal.

Could you rework it (or weigh in on the approach) around a **deterministic target-death preempt** instead? The idea:

- When an observed shooter's **target dies**, that's a real, packet-observable stop edge — the shooter stops auto-repeat server-side, and a dead unit can't be shot, so the retract can never be contradicted (the same terminal-proof reasoning as #389's melee ghost-swing fix).
- You already track observed shooters by GUID; pair that with the shooter's cached `UNIT_FIELD_TARGET`, and on a death event (`SMSG_PARTY_KILL_LOG` victim, or a unit's health→0 in the update path) synth the per-GUID `SMSG_CANCEL_AUTO_REPEAT` for any shooter whose target just died.

Honest tradeoff: target-death only covers the **common** case (the mob they're shooting dies — which is the scenario we keep seeing in groups). It won't catch a shooter who stops while the target's still alive (moved, `/stopattack`, OoR/LoS, out of ammo) — but that leaves the bow latched no worse than today, and it lets us drop the timeout entirely. We think that's the right trade.

Heads-up: this would likely **share the `SMSG_PARTY_KILL_LOG` preempt hook with #389** (ghost-swing on melee target death), so it may be worth coordinating with that PR. #389 is still in in-game testing, so no rush.

Open to your thoughts if you see a better deterministic signal.

jameopotato:
Follow-up on the rework — beyond target-death, a few more **deterministic triggers** worth considering for the per-GUID `SMSG_CANCEL_AUTO_REPEAT` (which your PR already proves is the effector that works; the open question is just what fires it):

- **Target-death** — the primary one above; terminal-proof.
- **LoS break** — promising *if* the 1.12 server emits something observable for the **remote** unit when its auto-shot fails LoS (e.g. `SPELL_FAILED_OTHER`). Worth a capture to confirm what's actually on the wire for other units — note #352 found `FAILED_OTHER` non-working as an *effector* (client ignores it), but it might still work as a *trigger* (we receive it, then fire the per-GUID cancel ourselves).
- **`/stopattack`** — deterministic and user-authoritative, but only for the **local** player's own auto-repeat, which already retracts via `SPELL_FAILURE`. For an observed **remote** shooter we never see their `/stopattack` (1.12 doesn't broadcast one for other units), so it doesn't extend this fix — flagging so it's not assumed to.
- **Shooter death / shooter movement** (`MSG_MOVE_*` for the observed unit) — other real stop edges if you want broader coverage.

None of these need to be exhaustive — target-death alone (dropping the timeout) is a solid first cut. The rest are optional coverage to layer on if the signal turns out to be cleanly available. Curious what you find on the LoS wire signal.

Mongrul:
Pushed a follow-up commit (`ffdfac3`) that turns this into a **hybrid retract**.

The original quiescence sweep covers the one case nothing else can — a shooter that stops with its target still alive and stands still (out of ammo, `/stopattack`, LoS, a target dummy) — but it's the only signal, so even an obvious stop waits out the full 4 s window.

This adds per-shooter target tracking and lowers the bow **instantly** on the deterministic stop edges the wire already gives us, while keeping the sweep as the catch-all:

- target dies (`PARTY_KILL_LOG` / health → 0) — terminal-proof
- shooter dies
- shooter moves (translational-start opcodes only; turn/facing/heartbeat excluded)
- shooter retargets or clears its target (`UNIT_FIELD_TARGET`)

A premature retract self-heals on the next shot's START, so the edges can't get it wrong. Validated in-game (2-boxing): the target-dummy case lowers ~4 s after the last shot via the sweep, and kill/move/retarget lower instantly via the edges. Tests: 14 `ObservedAutoRepeatTests` (was 9), full suite green.

Description updated to match.

jameopotato (dismissed):
Full re-review of the hybrid (`ffdfac3`) — accepting.

**In-game effect:** an observed shooter's frozen drawn-bow pose now lowers instantly when their target dies (PARTY_KILL_LOG + health→0, terminal-proof), when the shooter dies, moves (translational starts only — tracking a target by turning never false-retracts), or retargets; the 4s quiescence sweep remains the catch-all for stop-in-place (ammo out, /stopattack, target dummy). A premature retract self-heals on the next shot's replayed START — worst case is a one-beat bow-blink on another player's model, never local cast/button/GCD state.

**Policy note:** we're explicitly accepting the retained time-based sweep here as an exception to the no-timeout-cleanup line, on the grounds that it acts only on remote units' cosmetic aim with a self-healing failure mode, and that without it the stationary-stop slice keeps the original frozen-bow bug forever.

**Verified at full rigor:** one-remove-one-cancel invariant across racing edges; all state behind `_observedAutoRepeatLock`; sweep callback invoked outside the lock with teardown guards; timer lazily armed, self-disarming, test-gated (no real timers in the suite); reconnect reset unbinds/re-binds the callback. 14 dedicated tests cover every helper and both sweep behaviors.

**Merge (`origin/beta` merged in by maintainer):** conflict was in `HandleMovementMessages` vs the beta ghost-fix — resolved with the ghost-drop ordered BEFORE the move edge (a destroyed unit's stray move must neither reach the client nor count as a stop edge). 595 tests green on the merged result.

Non-blocking follow-ups: reconcile the SpellHandler.cs:1234 comment still listing CANCEL_AUTO_REPEAT as a non-working effector (true only for the local-GUID variant), and optionally add MSG_MOVE_KNOCK_BACK to the stopping-move allowlist.

Merges in this wave's batch.

jameopotato (approved):
Re-stamping approval (previous one was auto-dismissed by the stale-review rule racing the maintainer merge push 8f2b0bf — full review in the dismissed entry above). Verdict unchanged: hybrid accepted, 595 tests green on the merged head, merges in this wave's batch.

PR #377, opened 2026-06-16
LowLatency forward-all: skip on-START sweep under LL (C2)

jameopotato:
## What
Skip the on-START `ClearNonStartedNormalCasts` sweep under **LowLatencyMode**. The sweep fails OTHER non-started normal casts when one starts — right for hold-and-fire, wrong for forward-all, where concurrent non-started casts are legitimate mashed presses the server resolves individually. Sweeping them client-side while the server is still processing them is the **root** of the dup-rejection mis-dismiss class (the cast "interrupted but fired").

## Why it's the root
With the sweep off, a duplicate's rejection finds its own queue entry (caller-aware dequeue consumes it) instead of falling onto the live cast. So the transient-no-dismiss-started fix (#376) becomes a rarely-needed **backstop** — this removes the source. Field signal: `cast.non_started_swept` → 0 under LowLatency.

## Risks
- **Queue growth under mash:** non-started casts accumulate instead of being swept — bounded by the watchdog (2.5s eviction) + #372's dup-collapse.
- **Orphan-clear latency:** a cast the server never answers clears via the watchdog (≤2.5s) instead of the next SPELL_START's sweep. Slower but correct.
- Normal mode byte-unchanged (gated).

## Verification
- Build clean; 544 tests pass.
- Field: frantic-mash Frostbolt/AE under LowLatency → `cast.non_started_swept` should be 0, no stuck buttons / dismissed-but-fired.

## Stacking
Stacked on #376, `--base fix/transient-failure-no-dismiss-started` so this PR shows only the C2 diff; it auto-retargets to beta when #376 merges. First of the T2 conflict-scan items (C1/C3/C4 to follow). T3 deprioritized (handling > avoidance).

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -1372,23 +1372,34 @@ public partial class WorldClient
                 pendingCast.HasSentPrepare = true;
             }
 
-            // Clear non-started casts and send failures for them
-            // (keeps the started cast so SPELL_GO can dequeue it)
-            var failedCasts = GetSession().GameState.ClearNonStartedNormalCasts();
-            // DIAGNOSTIC (stuck-spell investigation): hoist toggle read; remove with diagnostics
-            bool debugEventsNormal = Framework.Settings.DebugOutput;
-            foreach (var failed in failedCasts)
+            // JimsProxy (C2 / forward-all): the on-START sweep fails OTHER non-started normal casts
+            // when one starts — correct under hold-and-fire (one cast at a time), but WRONG under
+            // LowLatencyMode forward-all, where concurrent non-started casts are legitimate mashed
+            // presses the server accepts/rejects individually. Sweeping them client-side while the
+            // server is still processing them is the ROOT of the dup-rejection-mis-dismiss class
+            // (the cast "interrupted but fired" desync). Under LowLatency skip it entirely; each cast
+            // resolves via its own SPELL_GO / CAST_FAILED (caller-aware dequeue) or the watchdog.
+            // Normal mode unchanged. (HasStarted / IsOffGcd casts are already exempt in the sweep.)
+            if (!Settings.LowLatencyMode)
             {
-                GetSession().InstanceSocket.SendCastRequestFailed(failed, false);
-                // DIAGNOSTIC (stuck-spell investigation): remove when closed
-                if (debugEventsNormal)
-                    Log.Event("cast.non_started_swept", new
-                    {
-                        queue = "normal",
-                        spell_id = failed.SpellId,
-                        triggering_spell_id = (uint)spell.Cast.SpellID,
-                        client_cast_id = failed.ClientGUID.ToString(),
-                    });
+                // Clear non-started casts and send failures for them
+                // (keeps the started cast so SPELL_GO can dequeue it)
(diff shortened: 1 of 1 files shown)


jameopotato:
**Parked (deliberately), rebased clean.** This draft is now exactly one commit on current beta: the C2 on-START sweep skip under LowLatencyMode (`be954d5`, field-run in the C2 session). The unfielded watchdog-1s tune was moved to the preserved side branch `fix/ll-watchdog-1s` — it lowers a verdict deadline to below realistic Kronos RTT spikes and needs its own scrutiny before it rides along.

**Un-park condition:** LL-cohort telemetry from v5.1.7 stable. The direct signal is `cast.transient_stale_dropped` frequency under LowLatencyMode — if it fires regularly in the field, the on-START sweep is still mis-failing concurrent forwards and C2 should come off the shelf (it composes with #397's hardened stale-drop by design: with dups unswept, rejections consume their own entries and the stale-drop becomes the rarely-needed backstop it was meant to be).

jameopotato:
**Closing unmerged — the un-park condition is not met.**

The condition set when this was parked was: `cast.transient_stale_dropped` firing *regularly* under LowLatencyMode in field telemetry. Swept the JSONL corpus (378 files; 80 LL-enabled runs, identified by `cast.low_latency_forward`, which only exists inside the `Settings.LowLatencyMode` branch).

Both emitters are `DebugOutput`-gated, so a zero only counts as evidence if logging was on — used the adjacent `cast.forwarded` under the same gate as a per-run witness. On in every LL run but one (`20260614-205007`, excluded).

Split at #397's merge (2026-07-03), the hardened stale-drop this was designed to compose with:

| window | LL forwards | `transient_stale_dropped` | rate |
|---|---|---|---|
| pre-#397 (06-14 → 06-26) | 2,442 | 74 | 3.0% |
| post-#397 (07-05 → 07-27) | 3,852 | **2** | 0.05% |

~58x drop across a *larger* post-#397 sample. Worst pre-#397 session: 23 stale-drops in one run (`B2/20260626-203236`). Worst post-#397 session: 1.

**The two survivors are one recurring interaction, not the target class.** Both are spell 6552, same character (`target_low 27032`), 12 days apart, identical adjacent signature:

```
SPELL_GO 7381
cast.unmatched_failure_forwarded  spell_id=6552
cast.transient_stale_dropped      spell_id=6552
aura.slot.set                     spell_id=2458   (Berserker Stance)
```

A stance-swap/Pummel edge, not the frantic-mash concurrency this PR was written for.

**Stated against the close, for the record:**
- The mechanism C2 disables is still live — the on-START sweep fired 31 times under LL in the post-#397 runs. The *premise* holds; only the *justification* has evaporated.
- 6552 is **not** in `SpellOffGcd1.csv`, so it takes the on-GCD LL path — those two events are technically in-class and C2 might have prevented them.
- The corpus is predominantly one player's sessions, not a broad LL cohort. This is "no evidence for", not "proof of absence".

**Rationale.** #397 removed the source at the root, which is what C2 was for. Keeping it means paying its real costs (queue growth under mash; orphan clears falling back to the 2.5s watchdog) to buy a defect firing ~twice a month on one spell, with no accompanying stuck-cast report.

Branches are preserved on origin and this is resurrectable verbatim if an LL stuck-cast report lands:
- `fix/low-latency-skip-sweep` @ `be954d5` (this PR)
- `fix/ll-watchdog-1s` @ `2658a1b` (the split-off watchdog tune)

Issue #383, opened 2026-06-23
Warlock summon (Ritual of Summoning) not visible to others — summoners appear standing still


jameopotato:
Reported via Discord. Affected build: **5.1.7-beta.1**.

## Symptom
Players performing a Warlock summon are not seen doing so — to other players nearby, the summoners/participants just appear to be **standing still**. The Ritual of Summoning channel/animation (and/or the Summoning Portal) isn't visible to observers.

## Notes (hypothesis, unverified)
Likely an observed-caster gap — the Ritual of Summoning channel isn't forwarded for other players, and/or the Summoning Portal GameObject isn't spawned/visible to observers.

## Repro (from report, unverified)
1. A Warlock + helpers start a Ritual of Summoning nearby.
2. Observe from another character — the participants appear idle, with no channel/portal visible.

## Related
- #211 Warlock Pet Control/Use Bug (open)

jameopotato:
Fixed by #393. Kronos sends Ritual of Summoning participants UNIT_CHANNEL_SPELL but no channel object, so the 1.14 client had the channel spell but no portal to face and never struck the channel pose for observers. #393 restores the channel object from the GameObject that cast the spell, so the summoning pose renders correctly.

In-game confirmed by the original reporter (who supplied the before logs). Ships in v5.1.7-beta.2.

PR #385, opened 2026-06-23
Render the Kronos Chronoboon Displacer correctly on the 1.14 client

Mongrul:
Kronos backports the Chronoboon Displacer as item 25007 — an id the retail 1.14 client maps to a different item — and its tooltip is dynamic: the server rewrites the Name/Description (the stored world-buff list) on every store/restore, but the modern client caches item templates by id and never re-renders a cached one. The boon rendered with a frozen tooltip, no bag sound, and a buggy on-use cooldown.

**Live tooltip — per-use entry aliasing.** On each store/restore (caught at the on-use cast's `SMSG_SPELL_GO`), the proxy re-queries the boon, mints a throwaway alias entry id the client has never seen, and re-creates the bag item under it so the client fetches the live tooltip clean. The guid→alias map is static so it survives a relogin; alias records are pre-built on the WC thread and drained in the client's query window so the WS handler never races the record stores.

**Bag sound — new per-emulator Item-table overlay.** The pickup/putdown clink is driven by `ItemGroupSoundsId`, which lives in the **Item** DB2 table (ItemSparse already had a Kronos overlay; Item did not). Adds `CSV/Hotfix/Item1.kronos.csv` + `LoadItemEmulatorHotfixes` (mirror of the ItemSparse emulator loader, distinct id range). A fresh hotfix id in `SMSG_AVAILABLE_HOTFIXES` makes even a cached client re-fetch 25007 at login, so the sound lands with no item use and no client cache clear. Kronos uses sound group 24 (empty-flask / soul-shard clink).

**On-use cooldown.** The server's one-shot `SMSG_COOLDOWN_EVENT` is dropped (it would blink on the stale item just before the destroy+recreate) and re-asserted on the recreated item. On a relogin with the boon **banked**, the client learns the cooldown at login (`SMSG_SEND_KNOWN_SPELLS`) but never binds it to a bank item shown only on bank-open, so the proxy captures the remaining time and repaints a GUID-bound `ItemCooldown` in `HandleShowBank`.

Server-agnostic: the Item overlay is gated to `ServerType=Kronos` and all behavior keys off the item name / on-use cast, so other servers are unaffected. Validated in-game on Kronos (live tooltip across store/restore/bank/relog, bag clink on a cold login, cooldown sweep surviving a banked relog).


jameopotato (changes requested):
Thanks for this — the alias-on-refresh architecture is well thought out, the per-emulator `Item` overlay (`LoadItemEmulatorHotfixes`) is a clean mirror of the existing ItemSparse overlay, and the WC/WS threading care (pre-building the alias packets at mint so `HandleDbQueryBulk` never mutates the record stores off the WS thread) is exactly right. The live path traces cleanly and it's in-game validated.

One thing to address before merge: the PR carries a good chunk of **dead/vestigial code** from the proactive-login-alias approach that was dropped on 2026-06-22 (per your own comment in `StoreObjectUpdateInternal` — bag sound moved to the static overlay). I confirmed via `git grep` against the PR head that these have no live callers/readers:

- **`RefreshDynamicItemRecords`** (`GameData.cs`) — defined, never called.
- **`RefreshDynamicHotfixRecord`** — only called by `RefreshDynamicItemRecords` (so dead with it), along with **`HotfixDynamicItemRefreshBegin`** + **`_dynamicItemRefreshHotfixCursor`**.
- **`_knownChronoboonEntries`** / **`MarkChronoboonEntry`** / **`IsKnownChronoboonEntry`** — `MarkChronoboonEntry` is still called in `HandleItemQueryResponse`, but `IsKnownChronoboonEntry` (the only reader) is never called, so it's a write-only set.
- **`ChronoboonBaseEntry`** const — never referenced.
- **`LoginRefreshEntries`** — only ever `.Remove()`'d in `HandleItemQueryResponse`, never added → permanently empty, so **`isLoginRefresh` is always false**. That makes two branches in `HandleItemQueryResponse` dead: the `isLoginRefresh && !Chronoboon → return` guard never fires, and the cooldown re-assert's `!isLoginRefresh` is always true.

Could you strip those (and collapse the now-constant `isLoginRefresh` conditionals)? In a change this concurrency-sensitive, removing the dead paths makes the live thread-safety story actually auditable, and it's a low-risk delete since none of it executes.

Non-blocking follow-ups (fine to leave for later):
- `EvictItemEntryAlias` does full O(`ItemEffectStore`) + O(`Hotfixes`) scans on every re-mint. Not hot-path (rare item), but an index would avoid the linear sweep.
- The static record stores (`ItemTemplates`/`ItemRecordsStore`/`ItemSparseRecordsStore`/`ItemEffectStore`/`Hotfixes`) are plain dictionaries mutated on the WC thread here and read on the WS thread in `HandleDbQueryBulk`. The loot path already has this exposure so it's pre-existing, but this adds more WC-thread writes — worth a separate lock/concurrent-store pass at some point.

Once the dead code is out I'll re-review — should be a quick turnaround to approve.

Mongrul:
Thanks for the thorough review. Stripped all of it in 061d757 — confirmed each had no live caller/reader before removing:

- `RefreshDynamicItemRecords` + `RefreshDynamicHotfixRecord` (and the `HotfixDynamicItemRefreshBegin` / `_dynamicItemRefreshHotfixCursor` pair) — the live tooltip refresh goes through the alias-recreate path, so the DBReply path was never reached.
- `_knownChronoboonEntries` / `MarkChronoboonEntry` / `IsKnownChronoboonEntry` — write-only set, dropped along with the now-unreferenced `ChronoboonBaseEntry` const.
- `LoginRefreshEntries` — only ever `.Remove()`'d, never added, so `isLoginRefresh` was always false. Removed it and collapsed the two branches it gated: the `isLoginRefresh && !Chronoboon -> return` guard and the `!isLoginRefresh` on the cooldown re-assert.

Net: 104 deletions / 1 insertion (the collapsed `if (onUseSpell > 0 && onUseCooldown > 0)`), dead paths only so no behavior change, builds clean. The live concurrency story is now just the alias-recreate flow you already traced.

Both non-blocking follow-ups are fair — leaving them out of scope here, but noted: `EvictItemEntryAlias` could take an index instead of the linear `ItemEffectStore`/`Hotfixes` scans, and the static record stores being WC-written / WS-read deserve a dedicated lock/concurrent-store pass (pre-existing on the loot path too). Ready for another look.

jameopotato (approved):
Re-review of `061d757` + conflict resolution:

**Change-request verified complete.** All listed dead symbols are gone from the PR head (grep-confirmed 0 references): `RefreshDynamicItemRecords`, `RefreshDynamicHotfixRecord`, `HotfixDynamicItemRefreshBegin`/`_dynamicItemRefreshHotfixCursor`, `_knownChronoboonEntries`/`MarkChronoboonEntry`/`IsKnownChronoboonEntry`, `ChronoboonBaseEntry`, `LoginRefreshEntries`/`isLoginRefresh`. The always-false `isLoginRefresh` collapse is semantically exact in both directions: the never-firing name-gate guard deleted outright, the cooldown re-assert made unconditional. Net −104/+1, live alias-recreate path untouched. 581 tests green on the PR head.

**Conflict resolution (`bcb37a2`, pushed by maintainer):** the branch was 7 commits behind beta and conflicted in `HandleSpellGo` — both this PR (Chronoboon re-query at GO) and #383's ritual channel-source memo added blocks at the same location. Resolved by keeping both (beta's #383 block first, matching beta layout); 581 tests green on the merged result.

Non-blocking follow-ups from the original review stand as noted-for-later: `EvictItemEntryAlias` linear scans, and the shared record-store WC/WS thread exposure (pre-existing pattern).

Approving — merges in this wave's batch.

PR #389, opened 2026-06-25
Preempt auto-attack stop when our target dies (ghost-swing fix)

jameopotato:
## Summary
- When a *settled* auto-attack target dies, the modern client keeps swinging the corpse for ~1 RTT (~200ms on a remote realm). It treats auto-attack as **server-authoritative**, so it only tears the swing down once the legacy server echoes our `CMSG_ATTACK_STOP` — a full round trip after the kill. The result is a damage-less "ghost swing" with no combat-log line, most noticeable at high latency.
- On `SMSG_PARTY_KILL_LOG`, if the killed unit is our current auto-attack target and we're in a settled attack state, synthesize the player's `SMSG_ATTACK_STOP` (`NowDead=1`) to the client immediately and clear `CurrentAttackTarget`. The real server stop ~200ms later is then a no-op.

## Why it's safe to preempt
`SMSG_PARTY_KILL_LOG` is **terminal proof**: the victim is already dead server-side, and a dead unit can never be auto-attacked, so the server can never contradict "you stopped attacking it." This is the same class of guaranteed outcome as the existing movement-cancel cast preemption (movement always interrupts a cast-time spell). The preempt only ever asserts a state the server is forced to agree with — never a guess.

## Compatibility with #321 (auto-attack stuck ~30s on a rejected swing)
Disjoint state by construction. `TryClearSettledAttackTargetOnDeath` fires **only** when `!WaitingForAttackStart && !DeferredAttackStop`:
- #321 owns the swing-start handshake (`WaitingForAttackStart == true`, swing rejected because the target died before it started). This guard skips that case — and there's no started swing to "ghost" there anyway.
- This owns the settled case (`WaitingForAttackStart == false`). By the time the real server stop echoes, `CurrentAttackTarget` is already cleared, so it lands on the `!WaitingForAttackStart` branch, never #321's.
- Both clear `CurrentAttackTarget = default` as the safe remedy, so neither can cause or undo the other.

## Edge cases
- Different unit dies (party member's kill / our DoT on another mob): no fire, target untouched.
- Pure caster (never meleed): no settled target, no fire.
- Stale `CurrentAttackTarget` (an earlier swing was rejected): worst case a no-op stop on the client; "not attacking the dead unit" is still true.
- **Known gap:** gray mobs that award no XP send no `SMSG_PARTY_KILL_LOG`, so they aren't preempted. By design — only terminal-proof triggers.

## Test plan
- [x] `AttackStopPreemptOnDeathTests` — full gating truth table (settled-fire, different-victim, WaitingForAttackStart-defer, DeferredAttackStop-defer, not-attacking)
- [x] Full suite green (586/586)
- [ ] PTR: grind XP mobs, confirm no post-death swing/sound and `combat.attack_stop_preempted` fires
- [ ] PTR: target-switch and rapid re-attack still feel correct (no #321 regression)

### DIFF
--- a/HermesProxy.Tests/World/AttackStopPreemptOnDeathTests.cs
+++ b/HermesProxy.Tests/World/AttackStopPreemptOnDeathTests.cs
@@ -1,0 +1,90 @@
+using HermesProxy.World;
+using HermesProxy.World.Enums;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (ghost-swing fix): gating for the preemptive auto-attack SMSG_ATTACK_STOP the proxy
+// sends to the modern client when the unit we're meleeing dies (SMSG_PARTY_KILL_LOG), instead of
+// waiting ~1 RTT for the legacy server to echo our stop. These tests lock in the disjoint-state
+// contract with PR #321 ("auto-attack stuck for ~30s when server rejects CMSG_ATTACK_SWING"):
+// the preempt fires ONLY in a settled auto-attack, and must defer (leave state untouched)
+// whenever a swing-start handshake is in flight — that state is owned by the SMSG_ATTACK_STOP
+// handler's #321 logic.
+public class AttackStopPreemptOnDeathTests
+{
+    private static GameSessionData NewState() => WowGuidTestHelper.CreateMockGameSessionData();
+    private static WowGuid64 Creature(uint counter) => new WowGuid64(HighGuidTypeLegacy.Creature, 1234, counter);
+
+    [Fact]
+    public void TryClearSettledAttackTargetOnDeath_SettledAttackOnDeadVictim_ClearsAndReturnsTrue()
+    {
+        var state = NewState();
+        var victim = Creature(1);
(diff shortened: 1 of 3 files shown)


jameopotato:
Design note / future link: the `SMSG_PARTY_KILL_LOG` "preempt on target death" pattern here could also drive the **observed-shooter ranged-aim retract** (#374) — when a remote shooter's target dies, synth a per-GUID `SMSG_CANCEL_AUTO_REPEAT` for that shooter, the same terminal-proof reasoning (a dead unit can't be shot, so the server can't contradict the retract).

Bodies stay separate — this PR retracts the **local player's melee** (`SMSG_ATTACK_STOP`, keyed on `CurrentAttackTarget`); the observed case retracts a **remote unit's ranged aim** (`SMSG_CANCEL_AUTO_REPEAT`, keyed on observed shooters whose cached `UNIT_FIELD_TARGET` just died) — but they could share this hook. Noting it so when #374 is reworked toward a deterministic target-death trigger (replacing its current time-based sweep) we coordinate the two. No action here; this PR's own in-game gate comes first.

PR #392, opened 2026-06-26
Fix Eggscilloscope on Rookery Egg (#387): present TRAP GO as GOOBER

### LINKED ISSUE #387: Eggscilloscope (12144) fails on UBRS rookery eggs — "Not in range"
## Summary
The **Eggscilloscope** (item ID **12144**), used for the **"Egg Freezing"** quest in the **Upper Blackrock Spire** rookery, does not work through the proxy. Using it on a Rookery Egg returns **"Not in range"** — or the use plays but has no effect. The same item/quest works on a **native 1.12 client**, isolating this to 1.14→1.12 proxy translation. **Three independent players** reproduced it, all on **latest beta.1**.

The item's on-use spell is **Freeze Rookery Egg (15748)** (quest "Prototype" variant **Freeze Rookery Egg - Prototype (16028)**), cast on the **Rookery Egg GameObject**.

> The provided Eggscilloscope Prototype for the quest Egg Freezing does not work on the eggs at the rookery within Upper Blackrock Spire

> i can use it on an egg the eggscillosope / but it dont work

## Environment
- **Proxy:** latest **beta.1**
- **Client → Server:** 1.14.2 Classic Era → 1.12 Kronos
- **Native 1.12 client: works** -> bug is in the proxy

## Steps to reproduce
1. On "Egg Freezing" in UBRS, obtain the Eggscilloscope / Prototype (item 12144).
2. Approach a **Rookery Egg** (GameObject) in the rookery.
3. Use the item on the egg.
4. **Expected:** egg detected/frozen (15748/16028 resolves; quest progresses).
5. **Actual:** **"Not in range"**, or the use animation plays but nothing happens.

## Suspected area (unconfirmed — for triage)
Item-use is handled in \`HandleUseItem\` — \`HermesProxy/World/Server/PacketHandlers/SpellHandler.cs:800\`. The egg is a **GameObject target**, and the symptom has two distinct possible roots that current captures can't yet separate:

**A. Server-side rejection** (use reaches the 1.12 server; its range check fails):
- *GameObject target translation* — \`ConvertSpellTargetFlags\` + \`WriteSpellTargets(use.Cast.Target, …)\` (\`:876-877\`): verify the \`TARGET_FLAG_OBJECT\`/GameObject GUID for the egg is written to the 1.12 \`CMSG_USE_ITEM\`. Note \`:809\` sets \`TargetGuid = use.Cast.Target.Unit\`, which is empty for a GameObject-targeted use — confirm the GO target rides through \`Cast.Target.Object\`.
- *Player position* — \`SyncLegacyServerPositionBeforeCast(use.Cast, source:"use_item")\` (\`:863\`) is documented for **reticle-targeted** uses (bombs, sapper, grenades); a **GameObject-targeted** item may not be covered, leaving a stale server-side player position → range fail.

**B. Client-side rejection** (the 1.14 client never sends the use; its own range check fails):
- Points at the **Rookery Egg GameObject's position translation** (the client renders the egg out of range), not the use-item path at all. In this case no \`CMSG_USE_ITEM\` is emitted.

## Logs
A reporter JSONL exists (\`jimsproxy-20260622-180131.jsonl\`, Kronos V) but that session ran with **DebugOutput off**, so the failed use leaves no trace (no \`cast.forwarded {source:"item"}\`; \`15748/16028\` never appear since the cast doesn't succeed). **A recapture with DebugOutput (and PacketsLog) enabled is needed to confirm A vs B** — it will show whether a \`CMSG_USE_ITEM\` even leaves the client.

## Notes
3 independent reports, consistent symptom → reproducible. Likely a broader **item-use / interaction on a GameObject** translation gap; other GameObject-targeted on-use quest items may share it.

jameopotato:
Fixes #387

## Summary
- The Rookery Egg (GameObject **175124**) is type **TRAP** in 1.12 data — it fires Summon Rookery Whelp on proximity. The 1.14 client won't select or spell-target a TRAP GameObject, so the Eggscilloscope (Freeze Rookery Egg **15748/16028**) found **no valid target**: no mouseover highlight, no interaction at any distance, and no `CMSG_USE_ITEM` ever left the client.
- Presents the egg as **GOOBER** (a selectable use/activate quest object) in both the GameObject query response (the value cached in `gameobjectcache.wdb`) and the object-update type field. Entry-scoped to 175124.
- The trap's proximity whelp-spawn is server-side and unaffected by the client-facing type change.

## Deployment note
The stale `TRAP` type is cached in the client's `gameobjectcache.wdb` and persists within the same client build, so this takes effect only after a one-time GO-cache clear. The launcher's auto-clear-cache option covers existing players.

## Test plan
- [x] In-game on Kronos (after cache clear): egg highlights on mouseover; Eggscilloscope targets the egg; freeze fires and quest progresses. Packet-confirmed: `gameobject.egg.typeoverride` (175124, 6→10) → `cast.forwarded` spell 16028 `source:"item"` → `spell.cast` start→go.
- [ ] Sanity: other Blackrock Spire GOs (doors / chests / runes) still interact normally.

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/QueryHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/QueryHandler.cs
@@ -542,6 +542,16 @@ public partial class WorldClient
         GameObjectStats gameObject = response.Stats;
 
         gameObject.Type = packet.ReadUInt32();
+
+        // Rookery Egg (175124): present vanilla TRAP (6) as GOOBER (10) so the 1.14 client lets
+        // the Eggscilloscope target it (see UpdateHandler / #387). This is the value cached in
+        // gameobjectcache.wdb, so players must clear that cache once to pick up the change.
+        if (response.GameObjectID == 175124 && gameObject.Type == 6)
+        {
+            gameObject.Type = 10;
+            Log.Event("gameobject.egg.typeoverride", new { entry = response.GameObjectID, from = 6, to = 10 });
+        }
+
         gameObject.DisplayID = packet.ReadUInt32();
 
         for (int i = 0; i < 4; i++)
--- a/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/UpdateHandler.cs
@@ -4171,7 +4171,17 @@ public partial class WorldClient
             int GAMEOBJECT_TYPE_ID = LegacyVersion.GetUpdateField(GameObjectField.GAMEOBJECT_TYPE_ID);
             if (GAMEOBJECT_TYPE_ID >= 0 && updateMaskArray[GAMEOBJECT_TYPE_ID])
             {
-                updateData.GameObjectData.TypeID = (sbyte)updates[GAMEOBJECT_TYPE_ID].Int32Value;
+                sbyte goType = (sbyte)updates[GAMEOBJECT_TYPE_ID].Int32Value;
+
+                // Rookery Egg (175124): vanilla GO type is TRAP (6), which fires Summon Rookery
+                // Whelp on proximity. The 1.14 client won't let you select/spell-target a TRAP, so
+                // the Eggscilloscope (Freeze Rookery Egg 15748) finds no valid target. Present it
+                // as GOOBER (10) — a selectable "use/activate" quest object — so the client can
+                // target it. The whelp-spawn trap is server-side and unaffected. See #387.
+                if (updateData.ObjectData.EntryID == 175124 && goType == 6)
+                    goType = 10;
+
+                updateData.GameObjectData.TypeID = goType;
             }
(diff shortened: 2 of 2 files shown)


jameopotato:
### Client cache — note for patch notes

This fix changes the GameObject **type** the client receives for the Rookery Egg (175124: `TRAP`→`GOOBER`). The client caches that value in `Cache\WDB\enUS\gameobjectcache.wdb`, and it persists across sessions within the same client build — the client only re-queries a GameObject it doesn't already have cached.

**Consequence:** a player who has already visited the rookery keeps the stale `TRAP` type and will see **no change** from the proxy update until their GO cache is cleared.

- The launcher's auto-clear-cache option handles this for launcher users.
- **When writing the patch notes for this release, include a user-facing recommendation to clear cache (or confirm the launcher auto-clear ran) to receive the fix** — otherwise existing players won't notice any difference despite updating.

Issue #382, opened 2026-06-23
Client locks up in BGs when an enemy uses a Gnomish Mind Control Cap on a nearby ally


jameopotato:
Reported via Discord. Affected build: **5.1.7-beta.1**.

## Symptom
The client freezes / locks up in battlegrounds when an **enemy** player uses a Gnomish Mind Control Cap on a **nearby allied** player. Still reproducing on the latest beta.

## Notes (hypothesis, unverified)
Likely a proxy translation issue with the charm/possess control state for an *observed* unit being mind-controlled (the reporter is just nearby — neither the MC user nor the target).

## Repro (from report, unverified)
1. Be in a BG near an allied player.
2. An enemy player uses a Gnomish Mind Control Cap on that ally.
3. The client locks up.

## Related
- #173 Gnomish MC Cap - No pet bar in PVP (closed)
- #112 SpellCastResultVanilla: Possessed enum typo (closed)

jameopotato:
Working on this one.

Was unable to reproduce the FPS drop on PTR with 3 clients (caster, target, observer) however that may just be from not having a full BG worth of noise. 
Regardless, 1.14 does not expect a CHARMEDBY packet for players and struggles to render raid frames with an enemy faction player in group. Will start with that fix in next beta and look wait for feedback.

jameopotato:
The presentation shim from #399 is on beta as of **v5.1.8-beta.2**: players observed being charmed by *another player* are now presented to bystanders as possessed, the way priest Mind Control already renders. NPC raid mind-controls (Lucifron et al.) are deliberately unaffected.

To be clear about status: this is an evidence-backed change, not a verified fix — the lockup was never reproduced locally (render-side symptoms don't show in packet streams, and local rigs can't reach BG crowd scale). **Verification ask for the next BG session on the beta proxy:** capture and send the session JSONL — we're looking for `charm.observed_player.possess_shim` firing when an ally gets capped, and whether the lockup still occurs. One known tradeoff while charmed: the ally is unclickable to observers for the ~20s charm duration.

Issue stays open until that field verification.

jameopotato:
**Investigation reopened (2026-07-16).** Field reports confirm the BG FPS drop persists on 5.1.8, which includes the #399 presentation shim — so the shim is either not firing in real BG conditions or (more likely) firing and insufficient.

[...]

Issue #173, opened 2026-05-08
Gnomish MC Cap - No pet bar in PVP


Mongrul:
Fix coming over for the next beta.

 ## Problem

  User report: Gnomish Mind Control Cap (item 4036, spell 13181) used on an
  enemy player in a battleground shows an entirely empty pet bar — no
  abilities, no actions. The cap visibly applies the charm aura but the
  modern client receives nothing to populate the temporary pet action bar.

  PVE creature MC works (partially) because it hits a different parser code
  path that's not gated by this lookup.

  ## Root cause

  `PetHandler.HandlePetSpellsMessage` calls
  `GetSession().GameState.GetItemId(spells.PetGUID)` to find the controlled
  unit's creature_entry — needed for the vanilla `CreatureFamily` fallback
  chain (cached → template → Wolf default).

  For **player** targets the cached `OBJECT_FIELD_ENTRY` field (vanilla
  index 3, creature-only) is never set. `GetItemId` used a raw dictionary
  indexer:

  ```csharp
  uint itemId = updates[OBJECT_FIELD_ENTRY].UInt32Value;

  …which threw KeyNotFoundException: "The given key '3' was not present in the dictionary." The PetHandler's outer
  try/catch then silently dropped
  the entire SMSG_PET_SPELLS_MESSAGE payload — empty bar.

  Confirmed via wire capture in bundle jimsproxy-20260507-223446:
  SMSG_PET_SPELLS_MESSAGE 60 bytes, full hex preserved by the existing
  pet.spells_message.parse_failed Log.Event:

  7901D2E31C00000000000000000001010000020000070100000700000007000000
  810000008100000081000000810200000601000006000000060000

  Fix

  Switch the indexer to TryGetValue, return 0 on miss. The existing
  PetHandler fallback chain (line 87-110) already expects
  creatureEntry == 0 for charm-style pets and falls through to
  creature_family = 1 (Wolf, generic vanilla beast family — guaranteed to
  resolve to a non-nil name in PetPaperDollFrame_SetStats). The rest of
  the packet now parses normally and the spell list reaches the modern
  client.

  Risk

  Tiny. One internal helper method, behavior change only on a code path that
  previously threw. All callers either ignore the return or treat 0 as a
  miss already (the PetHandler fallback chain explicitly handles this).

  Out of scope

  The PVE creature-MC path has a separate bug (handler returns 0 spells
  parsed from a 60-byte payload that visibly contains them) — different
  code path, tracked separately.

[...]

PR #399, opened 2026-07-03
Present observed charmed players as possessed (#382 bystander lockup)

jameopotato:
Addresses #382 (kept open pending reporter verification in a live BG).

## Symptom
In a BG, when an **enemy** uses a Gnomish Mind Control Cap on a nearby **ally**, a friendly **bystander's** client FPS-drops / locks up. Neither the charmer nor the target is affected, and priest Mind Control never triggers it (reporter-confirmed).

## Evidence trail
- **Flood refuted:** observer/participant captures show the charm arrives as a tiny, clean packet burst (~9 s2c), far below rates the client handles fine elsewhere in the same sessions.
- **Participants unaffected:** confirmed by 2-box capture and by the reporter ("used on me / I used it — no lockup").
- **The discriminator:** priest MC (605) = `Apply Aura: Possess`; the cap's control aura (13181) = `Apply Aura: Charm` (13180 is a dummy trigger). Verified via classic spell data.
- **Field-level difference (vmangos):** possess sets `UNIT_FLAG_POSSESSED` on the target; charm sets `UNIT_FLAG_PLAYER_CONTROLLED` + `CHARMEDBY` + faction flip + pet CharmInfo — **no POSSESSED bit**.
- **Client gap:** modern/1.14 has no concept of a *charmed player* (players are never charm targets in retail; MC there is possess). A friendly bystander rendering "my groupmate is an enemy's charmed pet" hits an unrepresentable state; the cost scales with raid/crowd size, which is why it's BG-only and why a 3-player local rig can't surface it.
- **Precedent:** #173 was the same root class on the proxy side — the pet path expected creature attributes a charmed *player* lacks (`OBJECT_FIELD_ENTRY` KeyNotFound → dropped pet bar).

## Fix
Track players observed being charmed by someone **other than the local player** (`GameSessionData.ObservedCharmedPlayers`) and OR `UNIT_FLAG_POSSESSED` into their `UNIT_FIELD_FLAGS` translation, so observing clients render their known-good possess path.

Scoping (preserves charm-vs-possess gameplay for participants):
- unit must be a **player** → creature charm (Enslave Demon, Eyes of the Beast, mob MC) untouched
- `CHARMEDBY != CurrentPlayerGuid` → the charmer keeps its real charm/pet bar
- `unit != CurrentPlayerGuid` → the target's own experience untouched
- possess targets already carry POSSESSED → no-op for priest MC
- per-session state → multibox-safe (each client session judges independently)

Lifecycle: entry added on observed charm apply, removed when `CHARMEDBY` empties (charm end) or on a fresh create block (re-appearance after out-of-range charm end). `charm.observed_player.possess_shim` / `..._cleared` events mark both edges.

## Known tradeoff
`UNIT_FLAG_POSSESSED` makes the charmed ally unclickable to observers for the ~20s charm (same as possessed units). Accepted vs. the lockup.

## Verification
- Build clean; full suite green (615/615).
- **Cannot be verified locally** — the render cost needs BG-scale raid/crowd. Reporter test steps:
  1. Run a BG on this build with logging on; be near an ally hit by an enemy's Gnomish Mind Control Cap.
  2. Expect: no FPS drop/lockup while the ally is charmed (should no longer occur).
  3. Send the JSONL — the `charm.observed_player.possess_shim` event proves the shim fired on the charmed ally.

### DIFF
--- a/HermesProxy/GlobalSessionData.cs
+++ b/HermesProxy/GlobalSessionData.cs
@@ -724,6 +724,18 @@ public sealed class GameSessionData
     // synthesize it. Same shape as KnownHoveringMobs above.
     public HashSet<WowGuid128> KnownSwimmingMobs = [];
 
+    // JimsProxy (#382 observer lockup): players we observe being CHARMED by someone other
+    // than the local player (Gnomish Mind Control Cap 13181 = vanilla MOD_CHARM on a player).
+    // Vanilla charm reaches observers as PLAYER_CONTROLLED + CHARMEDBY with no POSSESSED bit,
+    // and the 1.14 client has no representation for a charmed PLAYER (modern can't charm
+    // players; possess — priest MC 605 — renders fine, which is why priest MC never locks
+    // observers). While a guid is in here, the UNIT_FIELD_FLAGS translation forces
+    // UNIT_FLAG_POSSESSED so observing clients render their known possess path instead.
+    // Scoped to third-party observers only: the charmer keeps its real charm/pet bar and
+    // the target is untouched. Self-clears when CHARMEDBY empties (charm end) or on a
+    // fresh create block (re-appearance after an out-of-range charm end).
+    public HashSet<WowGuid128> ObservedCharmedPlayers = [];
+
     // JimsProxy (Tallstrider-Fix): per-GUID last-known facing orientation, populated from
     // any MovementInfo we observe (spawn, heartbeat, ObjectUpdate movement block). Used by
     // MovementHandler.HandleMonsterMove to compare the creature's current facing against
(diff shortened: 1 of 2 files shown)


jameopotato:
Review tightening (ae2f0fa): the shim now also requires the **charmer** to be a player (`charmedBy.IsPlayer()`, which subsumes the empty-guid check). NPC charms on players — Lucifron's Dominate Mind et al. — are deliberately excluded: years of raid MCs with no observer-lockup reports, so that long-stable content stays untouched. Widening to NPC charmers is a one-line follow-up if reporter verification implicates them.

PR #413, opened 2026-07-11
Fix PallyPower Lua error storm in cross-version raids (COOLDOWNS normalize + addon whisper-echo drop)

Mongrul:
## Symptom

Raiding with a mix of 1.14 (proxy) and 1.12-native paladins, the 1.14 client gets hammered with two PallyPower Lua errors — observed 624x + 943x in one raid night:

- `PallyPower.lua:1796: attempt to perform arithmetic on local 'duration1' (a nil value)` in `ParseMessage`
- `PallyPower.lua:635: attempt to perform arithmetic on field 'start' (a nil value)` in `PallyPowerBlessingsGrid_Update` (every frame while the Blessings window is open)

## Root cause

The 1.14 PallyPower parses any addon message containing `COOLDOWNS` by `strsplit`-ing the **entire message** on `:` and reading 4 positional tokens. A cooldown pair is skipped only when its *remaining* token is literally `"n"`; anything else is `tonumber()`'d and used in arithmetic unconditionally. A message with a malformed tail crashes ParseMessage **after** it has created an empty `CooldownInfo[1] = {}` — and that half-built table then errors the Blessings grid on every OnUpdate. Since every `SELF` from that paladin wipes and rebuilds their record, the cycle repeats all night.

Neither addon can be patched (players run stock CurseForge / vanilla-archive builds), so this is fixed at the proxy wire boundary, same as the original PLPWR interop.

## Fix (2 commits)

**1. Normalize inbound COOLDOWNS tails** (`AddonInteropTranslator`): inbound legacy→modern PLPWR bodies containing `COOLDOWNS` get the tail rewritten to exactly 4 receiver-safe tokens. Well-formed tails pass through byte-identical; salvageable numeric tokens are kept; a pair survives only if remaining is `"n"` or both halves are strictly plain-decimal (narrower than Lua `tonumber` so NaN/Infinity/hex can never reach the receiver's arithmetic). The substring gate deliberately mirrors the receiver's own unanchored `strfind` + whole-message `strsplit`. Outbound is untouched — the legacy wire stays canonical. Every rewrite logs `addon.interop.normalized` with the original body.

**2. Drop addon whisper-inform echoes** (`ChatHandler`, both vanilla and TBC/WotLK inbound handlers): legacy servers echo every whisper back as WHISPER_INFORM carrying the **target's** guid. The proxy forwarded addon-language echoes to the modern client as regular addon messages, so addons parsed our own outbound whisper as if the target sent it — in a normal raid PallyPower answers each REQ by whisper, so every 1.12 paladin's REQ made our own status (free-assign flag, symbol count, cooldowns) land on *their* grid row. A native modern client never receives its own addon whispers back; the drop restores parity. The drop runs before the HealComm inbound translate so no bridge can act on a mis-attributed echo. Normal chat whisper-informs ("To X: ...") are unaffected. Drops log `addon.whisper_inform.dropped` (body only for PLPWR, keeping binary/bursty addon sync out of the JSONL).

## Testing

- 14 new xUnit cases: malformed tail variants (bare, short, space-delimited, garbage, NaN/Infinity/exponent, n-duration+numeric-remaining), well-formed byte-identical passthrough, outbound untouched. Full suite 629/629 passing.
- Existing PLPWR ASSIGN/MASSIGN/SELF/PASSIGN translation and round-trip tests unchanged and passing.
- Verified against the 1.14 addon source (v1.4.2-classic): the pair rule matches the receiver's guard exactly, including the `":n:300"` shape that crashes stock parsing.
- In-game validation scheduled for the next cross-version raid; both diagnostic events will pinpoint the exact sending fork/format in the field.

### DIFF
--- a/HermesProxy.Tests/World/AddonInteropTranslatorTests.cs
+++ b/HermesProxy.Tests/World/AddonInteropTranslatorTests.cs
@@ -178,6 +178,52 @@ public class AddonInteropTranslatorTests
     }
 
     // -------------------------------------------------------------------
+    // COOLDOWNS normalization (inbound only) — malformed tails half-build
+    // the receiver's CooldownInfo and error every grid OnUpdate.
+    // -------------------------------------------------------------------
+
+    [Theory]
+    // Bare COOLDOWNS with no tail at all (empty cooldown list variants).
+    [InlineData("COOLDOWNS", "COOLDOWNS:n:n:n:n")]
+    [InlineData("FREEASSIGN NO | SYMCOUNT 184 | COOLDOWNS", "FREEASSIGN NO | SYMCOUNT 184 | COOLDOWNS:n:n:n:n")]
+    // Short tail: only one pair sent — pad the second pair.
+    [InlineData("COOLDOWNS:300:180", "COOLDOWNS:300:180:n:n")]
+    // Duration "n" but remaining numeric crashes the receiver — force the pair to n:n.
+    [InlineData("COOLDOWNS:n:300:n:n", "COOLDOWNS:n:n:n:n")]
+    // Garbage tokens become n:n pairs.
+    [InlineData("COOLDOWNS:abc:def:n:n", "COOLDOWNS:n:n:n:n")]
+    // Parseable-by-.NET but not receiver-safe tokens are rejected too.
+    [InlineData("COOLDOWNS:NaN:NaN:n:n", "COOLDOWNS:n:n:n:n")]
+    [InlineData("COOLDOWNS:Infinity:300:n:n", "COOLDOWNS:n:n:n:n")]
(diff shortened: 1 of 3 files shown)


jameopotato (approved):
Reviewed in full: COOLDOWNS normalize verified against the 1.14 addon source (unanchored strfind + whole-message strsplit + remaining~='n' guard — the fix mirrors it exactly, including keeping (numeric, n) pairs the modern sender genuinely emits), and the whisper-inform echo drop verified as native-parity (vmangos deliberately suppresses WHISPER_INFORM for LANG_ADDON; a native modern client never sees its own addon whispers). Full suite ran green on this exact head via throwaway branch ci/pr-413. Post-raid follow-ups tracked separately.

Issue #345, opened 2026-05-29
Item-use off-GCD casts not exempted from non-started-cast sweep (potential stuck-lit)


jameopotato:
Follow-up from #344 (off-GCD spell sweep exemption).

`HandleUseItem` (`World/Server/PacketHandlers/SpellHandler.cs`, ~L797) enqueues item-use casts into `PendingNormalCasts` but does **not** set `ClientCastRequest.IsOffGcd`. So the off-GCD exemption added in #344 to `ClearNonStartedNormalCasts` does **not** cover item-use casts.

On-use items (trinkets, potions, bombs, engineering gadgets) are off-GCD in 1.12. If such a cast is still pending when a concurrent spell's `SMSG_SPELL_START` arrives, `ClearNonStartedNormalCasts` can sweep it and emit a premature `CAST_FAILED` while the server casts it anyway — the same client "current spell" stuck-lit class that #344 fixed for Bloodrage and other off-GCD abilities.

**Not yet observed in any log** — raising as a paper trail for later investigation.

Possible fix: set `IsOffGcd` on the item path (mirror the spell path in `HandleCastSpell`), or otherwise exempt item-use casts from the sweep. Needs a repro + a decision on whether item-use casts should be queue-swept at all.

jameopotato:
Addressed in v5.1.6-beta.1 — off-GCD item-use casts (trinkets, bombs, gadgets) are now exempt from the non-started-cast sweep, so the action button should no longer stick lit. Reopen if it recurs. https://github.com/jameopotato/jimsproxy/releases/tag/v5.1.6-beta.1

PR #443, opened 2026-07-28
Don't let an in-flight off-GCD cast park every on-GCD spell (#442)

### LINKED ISSUE #442: Item-use cast leaks into PendingNormalCasts and silently parks all on-GCD casting until relog
**Version:** 5.1.9-beta.2 · **Server:** Kronos · **Mode:** Spell Queue (Low Latency off) · **Class:** Mage
**Occurrences:** twice, both during Ragnaros

## Symptom

All spell casting stops. No error text, no red message. Every keypress plays the cast animation and nothing happens; the character sits in the casting pose. Item use and movement still work. Survives `/reload` — only a full relog or proxy restart clears it.

## Confirmed in a raid JSONL

Molten Core session, ending at the restart that cleared it:

```
12:41:02.660  CAST 10181 go        Frostbolt completes, queue drains clean
12:41:02.858  CAST 10058 start/go  Mana Ruby              (+198 ms)
12:41:03.357  CAST 18385 start/go  Robe of the Archmage   (+499 ms)
12:41:03.378  spell.held_pending   spell_id 10181, displaced 0
12:41:04.539  spell.held_pending   spell_id 10181, displaced 10181
12:41:04.698  spell.held_pending   spell_id 10181, displaced 10181
              ... unbroken for 96 s, cycling through 10181 / 10151 / 10161 /
                  10202 / 122 / 10187 / 10216 ...
12:42:39      log ends - proxy restarted
```

Player damage casts per minute across the collapse: **15, 23, 2** — while raid-wide cast volume held at **1100–1682**, so the encounter was still in full swing. The player did not die.

Every Frostbolt cycle before the item pair closes cleanly (`cast.held_fire_on_cast_complete` on each), and the queue is demonstrably empty at 12:41:02.660. The only casts forwarded between that point and the first park are the two item uses — so the leaked `!HasStarted` entry is one of them, despite both logging `start` **and** `go`.

## Mechanism

1. Two off-GCD item-use casts are forwarded back to back. One leaks into `PendingNormalCasts` with `HasStarted == false` despite receiving its `SPELL_GO`.
2. `HasForwardedPendingCast()` (`GlobalSessionData.cs:2136`) returns true for **any** `!HasStarted` entry — it is not keyed by spell.
3. Every subsequent `CMSG_CAST_SPELL` hits the guard at `Server/SpellHandler.cs:619` and is parked by `ForceHoldCast`, **which arms no timer**. `:633` deliberately withholds `SpellPrepare` ("hide the queue from the client"), and each new press displaces the previous one with `CastFailed(DontReport)` — hence no error text, no lit button, and a client left holding its cast.
4. Nothing can reap it:
   - `ClearNonStartedNormalCasts` skips `IsOffGcd` entries (`:2291`) — which, since #345, is **every item-use cast**
   - `WatchdogDeadlineMs` is never armed for a forwarded-but-unstarted cast
   - the keepalive eviction pump is gated on `IdentityPinnedCastIdsActive`, i.e. Low Latency mode only

The state clears only via `ResetInFlightCastState` on relog / zone / reconnect.

## Why this evaded reproduction

Five solo sessions and 30+ item presses — same-frame pairs, ~450 ms pairs, items 108–160 ms after a cast completes, mid-cast-bar, mid-channel, both forwarding modes — never leaked an entry. Every item use resolved. The leak appears to require live raid conditions.

## Proposed fix

Exempt `IsOffGcd` entries from `HasForwardedPendingCast()`, mirroring the sweep exemption already granted by #344/#345. An in-flight off-GCD or item cast has no ordering claim over an on-GCD spell — packet captures confirm the server accepts both concurrently — so it should never gate one. Deterministic, no timer, consistent with the no-timeout-cleanup rule.

**This treats the lockout, not the leak.** An item entry surviving its own `SPELL_GO` is still unexplained and needs a separate issue; `DebugOutput` was off in the capture, so the enqueue side wasn't recorded.

jameopotato:
Fixes #442.

## Problem

`HasForwardedPendingCast()` returned true for **any** `!HasStarted` entry in `PendingNormalCasts`, including off-GCD ones — which since #345 means **every item-use cast**.

That gate drives the `ForceHoldCast` park at `Server/SpellHandler.cs:619`. While it holds, every on-GCD `CMSG_CAST_SPELL` is parked instead of forwarded, and the park is deliberately invisible to the client: no `SpellPrepare` for the new hold (`:633`), `CastFailed(DontReport)` for the displaced one. So the player sees the cast animation start, no cast, no error text, no lit button.

Nothing could reap a leaked off-GCD entry:

- `ClearNonStartedNormalCasts` skips `IsOffGcd` entries (`GlobalSessionData.cs:2291`)
- `WatchdogDeadlineMs` is never armed for a forwarded-but-unstarted cast
- the keepalive eviction pump is gated on `IdentityPinnedCastIdsActive` (Low Latency only)

The gate is not keyed by spell, so one leaked item entry killed *all* casting until relog.

## Evidence

From a live Ragnaros raid JSONL (5.1.9-beta.2, Spell Queue mode):

```
12:41:02.660  CAST 10181 go        Frostbolt completes, queue drains clean
12:41:02.858  CAST 10058 start/go  Mana Ruby              (+198 ms)
12:41:03.357  CAST 18385 start/go  Robe of the Archmage   (+499 ms)
12:41:03.378  spell.held_pending   spell_id 10181, displaced 0
12:41:04.539  spell.held_pending   spell_id 10181, displaced 10181
              ... unbroken for 96 s across 10181/10151/10161/10202/122/10187/10216 ...
12:42:39      log ends - proxy restarted
```

Player damage casts per minute across the collapse: **15, 23, 2**, while raid-wide cast volume held at **1100-1682** — the encounter was still in full swing and the player did not die.

## Change

One condition: off-GCD entries no longer satisfy the gate.

```csharp
if (!item.HasStarted && !item.IsOffGcd)
    return true;
```

Off-GCD casts coexist with a normal cast server-side — the legacy server casts them independently, confirmed on the wire across five capture sessions where item and spell casts were accepted concurrently every time. An in-flight off-GCD cast therefore has no ordering claim over an on-GCD spell and must not gate one. This is the same reasoning as the sweep exemption granted in #344/#345, applied to the gate that was left behind.

Deterministic, no timer, consistent with the packet-pairing-not-clocks rule.

## Tests

`HermesProxy.Tests/World/ForwardedPendingCastGateTests.cs` — 8 cases.

Verified red before the fix: `ForwardedUnstartedOffGcdCast_IsNotCounted`, `ForwardedUnstartedItemUseCast_IsNotCounted`, `LeakedItemUseCasts_DoNotJamOnGcdCasting` all fail on the unpatched tree; the 5 regression guards pass both ways.

Full suite: **767/767 passing.**

## Regression risk

The gate exists to preserve ordering while a cast is in flight. Narrowing it to on-GCD casts means an off-GCD press no longer delays a following on-GCD press by up to one round trip — which is the correct vanilla behaviour and what the sweep exemption already assumes. `LeakedOffGcdCast_DoesNotMaskARealForwardedOnGcdCast` pins that a genuine on-GCD cast alongside a leaked item entry still parks, so the exemption can't blind the gate.

Not affected: Low Latency mode (returns at `:454`, never consults this gate), pet casts, next-melee/auto-repeat slots.

## Scope

**This treats the lockout, not the leak.** An item-use entry surviving its own `SPELL_GO` is still unexplained — `DebugOutput` was off in the capture, so the enqueue side wasn't recorded. Tracked separately; the fix here makes that leak survivable rather than session-ending.

### DIFF
--- a/HermesProxy.Tests/World/ForwardedPendingCastGateTests.cs
+++ b/HermesProxy.Tests/World/ForwardedPendingCastGateTests.cs
@@ -1,0 +1,124 @@
+using HermesProxy;
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (#442): HasForwardedPendingCast() gates the ForceHoldCast park at
+// Server/SpellHandler.cs:619, which silently swallows EVERY on-GCD press (no SpellPrepare,
+// DontReport on the displaced one) for as long as it returns true. It is not keyed by spell,
+// arms no timer, and off-GCD entries — every item-use cast since #345 — are exempt from the
+// ClearNonStartedNormalCasts sweep and never receive a WatchdogDeadlineMs. So a leaked
+// off-GCD entry used to mean: all casting dead until relog. Observed live on Ragnaros
+// (Mana Ruby + Robe of the Archmage, then 96s of spell.held_pending).
+//
+// These pin the fix: off-GCD/item entries no longer satisfy the gate, while real on-GCD
+// forwarded casts still do — including when a leaked off-GCD entry sits alongside one.
+public class ForwardedPendingCastGateTests
+{
+    private static GameSessionData NewSession() => GameSessionData.CreateForTesting();
+
+    private static ClientCastRequest OnGcdCast(uint spellId, bool started = false) => new()
+    {
+        SpellId = spellId,
+        HasStarted = started,
+        IsOffGcd = false,
+    };
+
+    private static ClientCastRequest OffGcdCast(uint spellId, bool started = false) => new()
+    {
+        SpellId = spellId,
+        HasStarted = started,
+        IsOffGcd = true,
+    };
+
+    private static ClientCastRequest ItemUseCast(uint spellId, bool started = false) => new()
+    {
+        SpellId = spellId,
+        HasStarted = started,
+        IsOffGcd = true,                  // HandleUseItem tags item-use casts off-GCD (#345)
+        ItemGUID = new WowGuid128(1, 2),  // non-empty => item-use cast
+    };
+
+    [Fact]
+    public void EmptyQueue_NoForwardedCast()
+    {
+        var s = NewSession();
+        Assert.False(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void ForwardedUnstartedOnGcdCast_IsCounted()
+    {
+        // The behaviour the gate exists for: a normal cast is in flight to the legacy server
+        // and hasn't been confirmed yet, so the next press is parked to preserve ordering.
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(OnGcdCast(133)); // Fireball, forwarded-unstarted
+
+        Assert.True(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void StartedOnGcdCast_IsNotCounted()
+    {
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(OnGcdCast(133, started: true));
+
+        Assert.False(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void ForwardedUnstartedOffGcdCast_IsNotCounted()
+    {
+        // #442: an off-GCD cast coexists with a normal cast server-side, so it must not park
+        // on-GCD presses. Before the fix this returned true and jammed casting until relog.
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(OffGcdCast(11305)); // Sprint, forwarded-unstarted
+
+        Assert.False(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void ForwardedUnstartedItemUseCast_IsNotCounted()
+    {
+        // The reported case: Mana Ruby / Robe of the Archmage leak into the queue and
+        // previously parked every subsequent spell.
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(ItemUseCast(10058)); // Mana Ruby
+
+        Assert.False(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void LeakedItemUseCasts_DoNotJamOnGcdCasting()
+    {
+        // Exact shape of the Ragnaros log: two item-use entries left forwarded-unstarted.
+        // The gate must be clear so the next Frostbolt press forwards instead of parking.
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(ItemUseCast(10058)); // Mana Ruby
+        s.PendingNormalCasts.Enqueue(ItemUseCast(18385)); // Robe of the Archmage
+
+        Assert.False(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void LeakedOffGcdCast_DoesNotMaskARealForwardedOnGcdCast()
+    {
+        // The exemption must not blind the gate: a genuine in-flight on-GCD cast alongside a
+        // leaked item entry still parks the next press.
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(ItemUseCast(10058));  // leaked, ignored
+        s.PendingNormalCasts.Enqueue(OnGcdCast(133));      // real forwarded cast
+
+        Assert.True(s.HasForwardedPendingCast());
+    }
+
+    [Fact]
+    public void StartedOffGcdCast_IsNotCounted()
+    {
+        var s = NewSession();
+        s.PendingNormalCasts.Enqueue(OffGcdCast(11305, started: true));
+
+        Assert.False(s.HasForwardedPendingCast());
+    }
+}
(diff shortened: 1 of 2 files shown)


jameopotato:
## Further evidence from the full raid log — press timing is not the trigger

Scanning the whole raid session (05:54–12:42) turns up **two** `spell.held_pending` clusters, and the player's item uses rule out the "quick succession" framing:

| Item uses | Gap | Result |
|---|---|---|
| 11:35:15.475 ROBE + 11:35:15.721 Mana Ruby | **246 ms** | nothing |
| 12:25:28.900 ROBE + 12:25:53.954 Mana Ruby | 25 s | nothing |
| 12:14:28.341 ROBE **alone** | — | 2 parks, self-cleared |
| 12:41:02.859 Mana Ruby + 12:41:03.357 ROBE | **498 ms** | **terminal, 51 parks / 69 s** |

The 11:35 pair was **tighter** than the one that caused the lockout and did nothing; a **single** Robe use produced the near-miss. So neither the pair nor the interval is the cause — the leak is not about concurrency between two item casts.

## Why one recovered and the other didn't

Near-miss at 12:14 — the park is released almost immediately:

```
12:14:32.246  spell.held_pending  12051
12:14:32.394  spell.held_fire     12051   held_for_ms 141   <- released
12:14:32.539  spell.held_pending  10193
12:14:32.543  spell.held_fire     10193   held_for_ms 0     <- released
12:14:33.140  gcd.begin           10193                     <- a cast completed, timer armed
```

Terminal at 12:41 — **51 parks, not one `spell.held_fire`, and no `gcd.begin` ever again.**

The park is released by the GCD-expiry timer, and that timer is armed **only** by `BeginGcd` at a completed on-GCD local `SPELL_GO` (`Client/SpellHandler.cs:2069`). Once the gate is stuck true, nothing reaches the server, nothing completes, no `BeginGcd`, no timer, no release — **the recovery path requires a completed cast, and the state prevents any cast from completing.** At 12:14 the leak happened to be transient, so a cast completed and the timer bailed the player out.

That is why the fix targets *entry* into the state rather than escape from it: escape is only possible by luck.

## Robe slot-index lead — dry

Item 14152's on-use sits at `LegacySlotIndex 2` behind two equip-trigger spells, which made a slot-resolution divergence plausible. It appears in **no** `hotfix.itemeffect` event in the entire session — no `dbreply`, no `preserved_at_login`, no relocation — while mana gems, healthstones and potions all required corrections. The proxy found the Robe's `ItemEffect` already consistent with the server's template. No positive evidence for that theory.

## Still open

What leaks the entry. Both leak events involve the Robe (18385), one of them without any second item, and neither correlates with press timing. `DebugOutput` was off, so the enqueue side (`cast.forwarded` with `queue_depth_before`) wasn't recorded — that's the capture needed to identify which entry survives its own `SPELL_GO`.

jameopotato:
## Correction to the previous comment, plus the proof that was missing

### 1. The 12:14 cluster was not a near-miss — retracting that

I described `12:14:32.246 → 12:14:32.539` as "the same leak, transient". That is almost certainly wrong. Two parks released after **141 ms** and **0 ms** is `HasForwardedPendingCast()` behaving exactly as designed — a forwarded cast simply hadn't been confirmed yet inside a normal round-trip window. It is routine traffic, not a near-miss.

Consequence: the "ROBE alone → 2 parks, self-cleared" row in my table implied that a single Robe use nearly reproduced the bug. It did not. The conclusion that **press timing is not the trigger still holds**, but it rests solely on the 11:35:15 counterexample — a 246 ms Robe+Ruby pair, tighter than the 498 ms pair that caused the lockout, with no lockout — not on two independent observations as implied.

### 2. Proof the leaked entry is off-GCD (this was assumed, not shown)

The fix in #443 only helps if the leaked entry has `IsOffGcd == true`. That was inferred from "the only casts forwarded in that window were item uses", but `DebugOutput` was off so the enqueue side was never captured. The log does settle it, via a path-exclusion argument:

If the stuck entry were **Frostbolt (10181)**, every subsequent Frostbolt press would match `HasNonStartedPendingCastForSpell(10181)` and be dropped at `Server/SpellHandler.cs:458` with `cast.dropped.duplicate reason=in_flight_same_spell` — and that guard is evaluated **before** the `ForceHoldCast` park at `:619`.

Observed in the 69-second window: **51 × `spell.held_pending`, 0 × `cast.dropped.duplicate`**, across spells 10181, 10151, 10161, 10202, 122, 10187, 10216.

So the stuck entry's spell id matches **none** of the spells he pressed. The only other casts forwarded between the clean queue drain at 12:41:02.660 and the first park at 12:41:03.378 are the two item uses (10058, 18385) — both `IsOffGcd` since #345. The exemption therefore applies to it.

### 3. Residual after the fix: the item stays unusable for the session

#443 restores spell casting but does **not** fully clear the state. A leaked item entry still satisfies `HasInFlightNormalCastForSpell()`, which gates `HandleUseItem` at `:962`. That item — the Robe, the gem — remains **silently unusable for the rest of the session** (`cast.dropped.duplicate reason=in_flight_same_spell_use_item`), until relog.

So post-fix the symptom degrades from "no spell casting at all" to "one item permanently dead", which is a large improvement but not a cure. The underlying leak remains the thing to fix.

jameopotato:
Superseded by the duplicate-collapse fix — that removes the leaked queue entry instead of teaching one gate to ignore it, so the item no longer stays unusable for the session either. Root cause is now understood: the client's double-send leaves an unstarted twin that SPELL_GO deliberately spares (H7), and for an item use the server-side half that should consume it never arrives.

PR #444, opened 2026-07-29
Pair Kronos's empty-GUID inventory rejection with the pending item cast (#442)

### LINKED ISSUE #442: Item-use cast leaks into PendingNormalCasts and silently parks all on-GCD casting until relog

jameopotato:
Fixes #442. Root cause sealed live on 2026-07-29 (DebugOutput + PacketsLog on, deliberate reproduction).

## The complete chain, observed

```
14:34:37.089  CMSG_USE_ITEM press A forwarded (Mana Ruby)       qd_before=0
14:34:37.263  SPELL_START + SPELL_GO — server executes A, GO drains the queue
14:34:37.289  press B — a double-press straddling the GO by 26 ms — passes the
              now-empty dup guard and is forwarded                qd_before=0
14:34:37.473  Kronos rejects it: SMSG_INVENTORY_CHANGE_FAILURE
              result 23 (modern ItemNotFound), BOTH item GUIDs EMPTY
              raw wire: 17 | 00 00 | 00 00 | 00
14:34:38.089  spell.held_pending — every subsequent on-GCD press parked
...           cross-map teleport cures it: ResetInFlightCastState logs
              normal_casts_cleared = 1  (the leaked entry, counted by the cure)
```

Live-jam behavioural tests all matched prediction: off-GCD item uses still fire, ESC clears the button but not the state, no timer heals it, `/reload` doesn't fix it, same-MAP teleport (Searing Gorge → Ironforge, both map 0) doesn't fix it, cross-map (→ Darnassus, map 1) does.

## Why the entry was unreapable

`HandleInventoryChangeFailure` dequeues the pending cast **by item GUID** — an empty-GUID rejection matches nothing. The orphan then defeats every cleanup mechanism: unstarted (jams `HasForwardedPendingCast()` → `ForceHoldCast` silently parks every on-GCD press, no error text, until relog/map change), `IsOffGcd` since #345 (spared by `ClearNonStartedNormalCasts`), never `SPELL_FAILURE`-peeked (no `WatchdogDeadlineMs`), and a permanent `HasInFlightNormalCastForSpell` match (that item dead for the session).

Two rejection layers exist in the wild: the spell layer (`SPELL_FAILURE` + trailing `CAST_FAILED` — observed in the raid log at 12:25, self-healing) and the inventory layer (this one — leaked). The raid lockout at 12:41 and both of the original reporter's Ragnaros incidents fit the inventory layer.

## Change

When the GUID dequeue misses **and** the failure's GUIDs are empty, pair the rejection with the **oldest forwarded-unstarted item-use entry** (FIFO) and release its button with `DontReport`. Both wire variants of the handler. Deterministic packet pairing — the rejection is the server's answer to an outstanding use; when the GUID doesn't identify which, FIFO does. No timers.

**Known edge (documented on the method):** two different items in flight plus an *unrelated* empty-GUID inventory failure inside the one-RTT window can evict the wrong (healthy) entry. Cost: its later GO finds no queue match and forwards unmatched — cosmetic, self-limiting. The alternative is the permanent lockout.

**Backfill ordering (review Issue B, `4e502e13`):** the vanilla handler's pre-existing move-backfill rewrites empty wire GUIDs from the last forwarded bag move for 2 s — so the fallback gates on wire-GUID emptiness captured **before** that rewrite. Gating on the rewritten value would have silently disabled the fallback (and resurrected the lockout) whenever the rejection landed within 2 s of a bag shuffle.

**Empty-GUID guard (review Issue A, `a2510c55`):** `TryDequeueItemCast` now refuses an empty GUID outright. Spell entries leave `ItemGUID` at `default` (struct), so the plain `==` match paired `empty == empty` with the oldest unstarted **spell** entry — a spurious visible CastFailed for a healthy cast, and the real orphan surviving behind the `else if` whenever a spell entry was queued alongside it (exactly the raid shape). Anonymous rejections now always fall through to the FIFO fallback.

**Instrumentation, ungated** — the two blind spots that hid this for a week:
- `cast.item_use_rejected_by_inventory` (result code, `dequeued_via: guid|fifo_fallback`, evicted spell)
- `cast.result.discarded` (the status≠2 `CAST_FAILED` swallow — previously logged nowhere)

## History on this branch

The first attempt (collapse-at-GO, `b51692eb`) is reverted in `e1e1d295`: the CMSG-side dup guard makes pre-GO twins unreachable — confirmed live, one entry at the GO, press B arrived 26 ms after. Kept in history rather than force-pushed away.

## Tests

`ItemUseInventoryRejectionTests` — 9 cases: orphan dequeued; **both gates clear after eviction** (lockout + dead item); started entries never taken; spell entries never taken; FIFO order between two items; surrounding queue order preserved; empty queue safe; empty-GUID guid-dequeue matches nothing (Issue A pin — red without the guard); composed raid shape routes the rejection to the item, spell survives. Suite: **768/768.**

## Verification plan

The repro is now minutes — confirmed pre-fix on 2026-07-29: `/console maxfps 10`, then spam the gem across its own GO. The low framerate widens the straddle window from ~16 ms at 60 FPS to ~100 ms per frame (which is also why this hit in 40-man raids and resisted five solo repro sessions); that recipe produced the 14:34:37 lockout capture above. On this build the expected log is `cast.item_use_rejected_by_inventory` with `dequeued_via: fifo_fallback`, **no** `spell.held_pending`, casting uninterrupted, item usable after its cooldown.


jameopotato:
## Review — Request changes

Root-cause analysis and the FIFO pairing design are sound; tests pin the new primitive well; the release path reuses the field-proven malfunction-eviction pattern (`SpellPrepare` + `CastFailed(DontReport)`). Two correctness gaps found in the full enqueue/dequeue trace, both with precise fixes:

### Issue A — empty GUIDs never reach the fallback when a spell entry is queued
`WowGuid128` is a struct and a normal `CMSG_CAST_SPELL` enqueue never sets `ItemGUID`, so it stays `default`. `TryDequeueItemCast` matches by plain `==` — an **empty failure GUID matches the oldest unstarted spell entry** (`empty == empty`). Since the new fallback is an `else if` behind that call:

1. Queue = `[Frostbolt unstarted, gem-orphan]` (reachable: weave a cast, press the gem ~30ms later — both gates pass; or any off-GCD spell forwarded inside the one-RTT window).
2. Empty-GUID rejection → `TryDequeueItemCast(empty)` evicts the **Frostbolt** (wrong entry) with a visible spurious `CastFailed(SpellInProgress)`; its real `SPELL_START` later finds no queue match.
3. The `else if` never runs → **the orphan survives → the lockout still happens.**

The false-match is pre-existing, but 40-man raids — where both Ragnaros incidents occurred — are exactly where a concurrent unstarted entry is most likely: the fix works in the solo repro shape and silently degrades in the target shape.

**Required:** early `if (itemGuid.IsEmpty()) return false;` in `TryDequeueItemCast` (honest per its own doc comment; also fixes the pre-existing spurious spell-fail), plus a pinning test.

### Issue B — the keyring backfill masks the fallback (vanilla handler)
`HandleInventoryChangeFailureVanilla` backfills `failure.Item[0]` from `LastMoveItem0` when both wire GUIDs are empty and a bag move happened within 2s — the new `else if (failure.Item[0].IsEmpty() && ...)` then reads the **backfilled** value and skips the fallback. A #442 rejection landing within 2s of an unrelated bag move (mid-raid consumable shuffling) leaves the orphan jammed.

**Required:** capture `bool wireEmpty = failure.Item[0].IsEmpty() && failure.Item[1].IsEmpty();` **before** the backfill; gate the fallback on `wireEmpty`. Worst case becomes a benign double-action (client unlock for the move + silent orphan eviction), each self-limiting — same class as the PR's accepted two-items edge.

### Accepted as-is
- The two-items FIFO edge (documented, cosmetic, self-limiting) — right trade vs. permanent lockout.
- `cast.item_use_rejected_by_inventory` ungated (rejection-rate). `cast.result.discarded` ungated is acceptable but note it is proc-rate-bounded (07-17 incident: ~2s cadence under stuck-stun), not strictly press-rate — the chattiest of the new events if JSONL size ever matters.

After rework, the verification plan stands as written (low-FPS gem spam across own GO; expect `dequeued_via: fifo_fallback`, no `spell.held_pending`, item usable after cooldown).

https://claude.ai/code/session_01YKWSpntqMUYVpkoNUq8xs3

Issue #450, opened 2026-07-31
Killing white blow presents seconds late (floating text + swing sound after loot window) — trailing ATTACKER_STATE_UPDATE lands after #389's preempt stop


jameopotato:
## Report

v5.2.0-beta.2, 2026-07-31 ~09:39 session, warrior, no fps cap. Second-to-last mob of the session (a basilisk): the killing **white hit** — floating damage text **and** the swing impact sound — presented **after the mob died, after the player right-clicked the corpse, and after the loot window was open**. Felt severely out of order. This is the ghost-swing family reopened: PR #389 fixed the damage-less post-death ghost swing; this is a *real* hit presenting late in a case that fix didn't cover.

## Wire evidence (proven)

Session artifacts (local, reporter-held): `Logs/jimsproxy-20260731-092645.jsonl` + `PacketsLog/modern_42597_1785515213.pkt`.

Kronos sends the melee killing blow **after** its own kill packets. Legacy burst, all in the same millisecond (ts 1785515959885):

```
SMSG_ATTACK_STOP          (mob -> player)
SMSG_PARTY_KILL_LOG
SMSG_LOG_XP_GAIN
SMSG_ATTACKER_STATE_UPDATE (player -> basilisk, 215 dmg — the killing blow)
```

This inverted order is Kronos's norm for melee killing blows: **6 of 12 kills** this session have the killing ASU trailing PARTY_KILL_LOG by 0ms (the other 6 look like spell killing blows).

The proxy forwarded everything in order within the same tick, but the #389 preempt (`combat.attack_stop_preempted` — fired on all 12 kills) inserts itself between the kill and the hit. Modern wire order (from the .pkt, one tick, 1411609093):

```
SMSG_ATTACK_STOP    (mob -> player)                      [forwarded]
SMSG_PARTY_KILL_LOG                                       [forwarded]
SMSG_ATTACK_STOP    (player -> mob, NowDead=1)            [synth, #389 preempt]
SMSG_THREAT_CLEAR                                         [synth]
SMSG_LOG_XP_GAIN                                          [forwarded]
SMSG_ATTACKER_STATE_UPDATE (player -> mob, 215)           [forwarded — the late-presenting hit]
SMSG_THREAT_UPDATE / SMSG_HIGHEST_THREAT_UPDATE           [synth — see finding 3]
```

[...]

PR #452, opened 2026-07-31
Hold the ghost-swing preempt stop until after the killing-blow ASU (#450)

jameopotato:
Refs #450 (left open for field verification per process — close at the beta it ships in, once verified).

## Problem

Kronos sends the melee killing blow's `SMSG_ATTACKER_STATE_UPDATE` **after** `SMSG_PARTY_KILL_LOG` in the same burst — 6 of 12 kills in the #450 capture (`modern_42597_1785515213.pkt` + `jimsproxy-20260731-092645.jsonl`, kill ts 1785515959885). PR #389's preemptive `SMSG_ATTACK_STOP` (NowDead=1), emitted inline from the kill-log handler, therefore landed **between the kill and the hit** on the modern wire:

```
SMSG_ATTACK_STOP (mob→player)      forwarded
SMSG_PARTY_KILL_LOG                 forwarded
SMSG_ATTACK_STOP (player, NowDead)  ← #389 synth, inline
SMSG_LOG_XP_GAIN                    forwarded
SMSG_ATTACKER_STATE_UPDATE (215)    ← the killing blow, too late
```

The modern client, told its attack already stopped, re-plays the late hit as a fresh swing on the corpse — floating text + swing sound seconds later, after the loot window is already open (reporter observed both, no fps cap; wire delivery was ~200ms *before* the loot click, so the lateness is client presentation triggered by the ordering).

Secondary finding fixed here too: the trailing killing-blow damage re-created the dead mob's ThreatTracker list (`threat.emit` top_threat=172 *after* the kill-log `threat.clear`) and emitted `THREAT_UPDATE` + `HIGHEST_THREAT_UPDATE` for a corpse until the death values-update cleared it again ~20ms later.

## Fix

The preempt stop is now **armed** by `HandlePartyKillLog` (same `TryClearSettledAttackTargetOnDeath` gate — the #321 disjoint-state contract is untouched) and flushed by whichever trigger fires first:

| flush | trigger |
|---|---|
| `asu` | `HandleAttackerStateUpdate` forwarded the local player's hit on the armed victim — stop emitted right behind it (hit → stop, the natural order) |
| `drain` | receive loop: legacy socket has no more buffered packets (spell killing blows — no trailing hit); same read pass, sub-ms later than the old inline emit |
| `rearm` | a second kill armed before the first flushed (multi-kill burst) — older stop flushed immediately |
| (silent) | the server's own player `SMSG_ATTACK_STOP` echo consumes the armed stop — no duplicate |

`combat.attack_stop_preempted` keeps its name and gains the `flush` field. A mid-burst TCP segment boundary can only drain early, i.e. reproduce the old inline ordering — never anything worse.

ThreatTracker: `OnMobKilled` (kill-log path only) tombstones the guid and clears; `AddThreat`/`SetThreat`/`SetToTop` no-op while tombstoned (DebugOutput-gated `threat.dead_mob_add_ignored`). Any lifecycle signal for the guid — combat-state observation (either edge), destroy, leave-combat wipe, reset — disarms, so a respawned mob reusing the same creature guid is never blocked. Plain `ClearMob` (evade/leash) deliberately does not tombstone.

## Tests

- 14 new: `AttackStopPreemptOrderingTests` (7 — arm/consume/take state machine, multi-kill rearm, wrong-victim/wrong-attacker guards) + `ThreatDeadMobTombstoneTests` (7 — gate, evade-clear non-tombstone, all four disarm paths).
- Tombstone gate proven red-first: with the `AddThreat` gate disabled, the trailing-damage test fails exactly as the capture predicted.
- Full suite: **795/795**.

## Known limits / out of scope

- A pending stop surviving a world reconnect could emit one stale (harmless, NowDead) stop on the new connection.
- #450 finding 1 — the client no longer sends `CMSG_ATTACK_STOP` after kills, so Kronos keeps the server-side swing running (`ATTACKSWING_DEADTARGET` 0.6–3.6s after 8/12 kills) — is deliberately **not** here; it touches the #321 swing state machine and gets its own PR after this ships and is observed.

## Field verification

Melee killing blow + fast loot: the modern .pkt must show the ASU **before** the synthesized stop, JSONL shows `combat.attack_stop_preempted` with `flush:"asu"`, and the hit presents as the mob drops. Spell kills: `flush:"drain"`. `threat.dead_mob_add_ignored` replaces the post-kill threat resurrection.

https://claude.ai/code/session_01AzFxGSbHw1ECaaq31kdbNc


jameopotato:
## In-game verification 2026-08-12 — PASS (hold lifted, merged)

Fix build (`5.2.0-fix-450-preempt-stop-order.1+14`, session `jimsproxy-20260812-141139.jsonl`, 9 melee kills): `combat.attack_stop_preempted` **`flush:"asu"` 9/9** — the stop released behind the killing blow on every kill; kill→blow gap 0–6ms; **tombstone 9/9 with zero corpse-threat resurrections**. A 10th PARTY_KILL_LOG without a settled melee target correctly armed nothing. No drain/rearm fallbacks needed.

Same-hour stock control (`5.2.0-beta.2`, session `jimsproxy-20260812-142924.jsonl`, same mob camp): inline preempt (no `flush` field) on all 9 kills, and the tester observed the inverted presentation the fix removes — mob visibly in its dying animation before the final hit and damage render. Fix build: presentation clean; the first kill showed damage text a beat after death, which is Kronos's true announcement order (kill log before the blow packet, present on both builds) rather than the replay bug.

Residuals on both builds, out of scope per the PR body: `ATTACKSWING_DEADTARGET` forwarded 0.2–1.6s after **every** kill (9/9 this session — the post-#389 client never tells the server to stop swinging; follow-up PR), and the inherent death-before-damage announcement order.

#450 stays open for field observation at the beta this ships in.

PR #454, opened 2026-08-03
Res sickness keeps its debuff timer — preserve fresh duration pushes across same-tick slot swaps

jameopotato:
Reported 2026-08-02: res at a spirit healer applies res sickness with no debuff countdown. Root cause wire-proven the same day in `jimsproxy-20260802-182258.jsonl`.

## The race, observed (ts 1785725264529–538)

```
...529  SMSG_SPELL_GO 15007 (res sickness self-cast)
...529  SMSG_UPDATE_AURA_DURATION        ← the timer, stored for slot 32
...537  SMSG_UPDATE_OBJECT               ← slot 32: Ghost (8326) → 15007, direct swap
...538  aura.slot.set  duration_full:0, duration_left:-1   ← timer already wiped
```

Vanilla cores send `SMSG_UPDATE_AURA_DURATION` **immediately** at aura apply but batch the field update to end of tick — so on a direct slot swap the NEW occupant's duration always precedes the swap by a few ms. Ghost always occupies slot 32 (first debuff slot) while dead and is replaced by sickness in the same server tick, so this is deterministic at every spirit-healer res, not a rare race.

The swap-wipe guard in the UpdateHandler aura loop (the Rupture Y→X stale-duration fix) saw prev-emitted Ghost ≠ 15007 and called `ClearAuraDuration` — destroying the fresh server-authoritative timer stored 8 ms earlier. Every fallback then missed: `GetAuraDuration` sentinel `-1`, no AuraDurations CSV row for 15007 (level-scaled spell, correctly absent), expiry-map restore is create-only. Confirmed the guard specifically: no `aura.slot.cleared` for slot 32 between the Ghost emit and the sickness emit, so `prevEmitted` was still armed. The re-emit 16 s later (mask_levels) shows the same 0/-1 — never recovers in-session.

## Change

The guard can't distinguish "previous occupant's leftover" from "new occupant's duration that raced ahead" by spell ID — the duration packet carries only a slot. It CAN by time:

- `GameSessionData.UnitAuraDurationPushTime` — per (unit, slot) TickCount written **only** by the two server duration-push handlers (`HandleUpdateAuraDuration`, `HandleSetExtraAuraInfo`), never by emit-path stores (finisher snapshot, expiry restore). That exclusivity is load-bearing: stale-Y durations are stored at Y's cast/emit and are always older than the window at swap time, so the Rupture wipe still fires for them.
- `HasFreshAuraDurationPush` — 1 s window, wraparound-safe (`unchecked` age, negative = not fresh).
- Swap guard keeps the duration when a push for that exact slot is fresh and emits ungated `aura.duration.preserved_on_swap` (rare: once per actual swap race — field-verify signal); wipes exactly as before otherwise.
- `ClearAuraDuration` / `EvictUnitAuraState` drop the push entry with the rest of the per-slot state.

`GetAuraDuration` already decays `left` by elapsed-since-store, so the preserved value is served correctly.

**Known accepted edge:** a self-aura duration refresh followed within 1 s by a direct same-slot replacement briefly shows the old spell's duration under the new one — requires refresh+replace across adjacent server ticks; self-corrects on the next update. The pre-fix alternative was a permanently timerless debuff.

**Noted, deliberately not fixed here:** `HandleUpdateAuraDuration` reads the *cached* slot occupant, so during the race it emits the fresh duration attached to the stale spell (Ghost) and records its expiry under 8326. Cosmetic 8 ms blip, overwritten by the swap emit; fixing it needs a spell ID the packet doesn't carry.

## Tests

`AuraDurationSwapPreserveTests` — 9 cases: fresh push true at the observed 8 ms gap; window edge inclusive; stale push false; **emit-path stores don't arm the window** (the Rupture-regression pin); wrong slot; clock-backwards; `ClearAuraDuration` and `EvictUnitAuraState` drop push state; the logged sequence replayed at state level end-to-end (push → fresh check → decayed duration served). Red-proof pass: sabotaged wrong-dict + both missing cleanups → 4 red, incl. the Rupture pin against the tempting reuse of the shared `UnitAuraDurationUpdateTime`. Suite **790/790**.

## Verification plan

Field gate before this reaches a tag: die, res at a spirit healer. Expected log: `aura.duration.preserved_on_swap` (prev_spell_id 8326 → spell_id 15007) and a ticking countdown on the debuff; `aura.slot.set` for 15007 with real duration_full/left instead of 0/-1. Also covers any other same-tick swap into an occupied slot on the player.

https://claude.ai/code/session_01NWnZQiuCasQZJBGMAQ58io

## Regression provenance

This is a **regression introduced 2026-07-27** in the squash of #435 (`8cdcb544`, the party-pet/stale-member fix, whose aura-state hardening added the swap-wipe guard). Before that commit the aura loop read `GetAuraDuration` with no swap wipe, so the duration stored 8 ms earlier survived and res sickness showed a normal countdown. First shipped in **v5.2.0-beta.1**; present in beta.2. 5.1.9 stable is unaffected. This PR restores the pre-beta.1 outcome for the race while keeping #435's guard for genuinely stale durations.


jameopotato:
## Repro / fix-confirmation steps

**Reproduce the bug (any v5.2.0-beta.1 or beta.2 build):**
1. Character level 11+ (below 11 the spirit healer applies no sickness).
2. Die, release spirit, and take the **spirit healer resurrection** at the graveyard (not a corpse run, not a player res).
3. Resurrection Sickness appears on the debuff bar with **no countdown** — no timer ring, tooltip shows no remaining time, and it stays that way for the whole session. Log signature: `aura.slot.set` for `spell_id:15007` with `duration_full:0, duration_left:-1`.

**Confirm the fix (this branch):**
1. Same steps — die, release, spirit-healer res.
2. Expected on screen: the debuff shows a **ticking countdown** immediately (1 min at levels 11–19, 10 min at 20+).
3. Expected in the JSONL log:
   - `aura.duration.preserved_on_swap` with `prev_spell_id:8326` (Ghost) → `spell_id:15007`
   - the following `aura.slot.set` for 15007 with real `duration_full`/`duration_left` values instead of `0/-1`
4. Regression check (guard still works): normal play — recasting DoTs/debuffs that land in occupied slots must not inherit a previous aura's leftover timer (the #435 Rupture behavior). Any `aura.duration.preserved_on_swap` lines outside a res are worth a second look; they should be rare to nonexistent.

Repro is deterministic — one death per build is enough to verify.

PR #455, opened 2026-08-03
World-entry root-ceremony breadcrumb (always-on) + dev delivery harness (stage 0.5)

jameopotato:
## Problem

The 2026-08-02 stage-0 captures (12 BG transitions on the Kronos dev's local core + 4 PTR transfers, #453 tripwire build) established the arrival **root ceremony**: every world-entry delivers SMSG_MOVE_ROOT ×2 at the loading boundary and SMSG_MOVE_UNROOT ~1s later — every leg client-acked in every healthy capture, every MoveCounter **0** (the 1.12 server never increments; 39/39 ops). A root whose unroot leg fails is the leading mechanism for the post-BG-exit movement lockup (translation + jump dead, turning/casting alive, `/reload` cures — the symptom **is** root semantics; creates are clean 18/18 so the root exists only as force-ops, which per the reload-asymmetry model is exactly the reload-curable kind).

Two operational gaps blocked progress:
1. Everything recording the ceremony was DebugOutput-gated — a reporter's normal Export Diagnostics could never carry the discriminator, and reporter contact is off the table.
2. Forcing the suspected race needed server-side changes (vetoed — server dev is a last resort) or waiting for a slow machine to lose it naturally.

## Change

Logic lives in `World/Client/WorldEntryCeremony.cs` (extracted pure state/predicates, TransportClearGate precedent).

**(a) Always-on breadcrumb** — `GameSessionData.WorldEntryCeremony` counts the player's ROOT/UNROOT forwards (`Client/MovementHandler.HandleMoveForceFlagChange`) and their acks (`Server/MovementHandler.HandleMoveForceAck2`) between arrival anchors (NEW_WORLD forward / login verify), stamps `CMSG_MOVE_INIT_ACTIVE_MOVER_COMPLETE`, and at the next boundary emits **one** `worldentry.ceremony.unclosed` JSONL line when a ceremony opened a root without observably closing it:
- **no unroot forwarded** → server-side skip (theory T1)
- **unroot forwarded, never acked** → client swallow (theories T2/T3′ — the stuck-stun golden capture proved a discarded force-op is visible precisely as its missing ack)
- **root acks short** → a root leg was discarded

Healthy ceremonies log only under DebugOutput (`worldentry.ceremony.closed`). Packet-ordered, **no timers**, ≤1 line per transition, ~zero volume. Known benign false positive: a genuine gameplay root held across a hearth logs one line — it's a breadcrumb for correlation, not an alarm (documented in-code).

**(b) Dev delivery harness** — config-gated, **default off**, launcher never writes these keys, no patch-notes line per house policy:
- `WorldEntryHarnessMode=drop_unroot` (+`WorldEntryHarnessCount`) — swallow the player's next N arrival unroots. Experiment E1: does a stranded force-root reproduce the exact reported symptom table including the `/reload` cure? (Converts the two central hypotheses into facts in one Deadmines hop.) Note: the legacy server's awaited unroot-ack never comes — if Kronos enforces force-op acks, the kick is itself a result.
- `WorldEntryHarnessMode=hold_until_init` (+`WorldEntryHarnessDelayMs`) — queue the player's arrival ROOT/UNROOT while the client is loading; release at init-mover-complete + offset (E2: walk deliveries through the never-captured quadrant). Fail-open: leftovers flush at the next boundary so ops can never leak across a transition. The release delay is a wall-clock timer — deliberately allowed here because this path is unreachable outside the dev-only config (called out in-code; the no-timers house rule governs shipped fix logic).
- `WorldEntryMintMoveCounters` — mint session-monotonic force-op counters toward the client (recognizable ≥1001 in captures), restore the server's original on the ack path (E3: A/B for the counter-collision theory).

## Verification

- Build clean; **suite 790/790** (8 new tests: unclosed-predicate truth table including the golden-capture fingerprint shapes, tracker begin/reset, mint monotonic / resolve-once / unknown-counter passthrough).
- Red-proof note: these tests pin new pure logic (instrumentation), not a bug fix — there is no pre-existing failing symptom to revert to red against. The predicate's truth table is anchored to real captures: the healthy 18/18 shape asserts `false`, the stuck-stun golden capture's missing-root-ack shape asserts `true`.
- Zero-behavior-change with defaults: the only new runtime effects are counter increments and ≤1 JSONL line per transition.
- Field gates (after merge, before any conclusions): (1) healthy transitions on a dev box produce zero `unclosed` lines; (2) E1 `drop_unroot` run reproduces the reported symptom table incl. `/reload` — that run is the mechanism confirmation, deliberately **not** claimed by this PR.

## Risk

- Breadcrumb: counting + one conditional log line; no packet mutated, no timing changed. Worst case = a spurious breadcrumb line.
- Harness: entirely behind default-off config strings that ship nowhere; the fail-open flush bounds the worst case of a misconfigured hold to "ceremony delivered late at the next boundary."
- Counter minting mutates outbound counters ONLY under its toggle; ack path restores originals, unknown counters pass through untouched.

https://claude.ai/code/session_01U1YoZNxHnkZZT9tTRNzAHN


jameopotato:
Withdrawn per process: diagnostics/harness work deploys as a worktree test build, not a PR. PR slot reserved for the actual fix; branch stays for the eventual fix to build on.

Issue #328, opened 2026-05-27
BG-exit lockout: black action bars, can't move, requires relog


jameopotato:
After leaving a BG, player is fully locked out: action bars go black, can't move, must relog to recover.

## Evidence available
- JSONL log (602k lines, 2026-05-26, Kronos V) from CHICKENSOUL: 4/5 BG exits had paired `CONTROL_UPDATE false/true`; the 5th exit had `false` with no `true` for the rest of the session.
- Additional context from reporter: was **underwater** when teleporting both into and out of the BG.
- Trigger chain: mage portal to UC → flight → BG → exit → stuck.

## Hypothesized mechanism
Server omits the `CONTROL_UPDATE(HasControl=true)` that's normally paired with the `(false)` scoreboard-freeze packet.

## Related work
PR #327 (`fix/bg-exit-control-restore`) — synthesizes the missing `ControlUpdate(true)` on `MOVE_UNROOT` within 5s of a world transfer. This is a defensive synth, **not a root-cause fix** — the underlying server-side or accumulated-state trigger is still unknown. Worth keeping for safety but the actual cause needs separate investigation.

## Evidence needed
- More BG-exit logs from anyone hitting this, especially not-underwater at exit time (to confirm/refute the underwater correlation).
- Confirmation whether the bug recurs with PR #327 merged.

jameopotato:
**Dev investigation update (2026-07-01) — likely a distinct mechanism from the /reload-fixable movement wedges (#329/#331/#332)**

Cross-referencing against the movement-input wedge cluster: this is probably not the same bug. Those are `/reload`-fixable, bars-not-darkened, rotation-works → client-side movement-input corruption (leading theory: the transport-clear synth firing-for-nothing and racing at low RTT). This issue is black-bars + can't-rotate + relog-only, the signature of genuine `HasControl=false` (server-side control not restored) — matching the hypothesis already in the body.

From capture analysis the world transfer itself always completes (`CMSG_WORLD_PORT_RESPONSE` always arrives), so this isn't a stuck handshake — it's specifically the missing `CONTROL_UPDATE(true)` restore. #340 (merged) now logs the parsed `HasControl` value, so a fresh capture confirms/refutes the "server omits the restore" theory directly (the earlier 602k-line log predated that logging). Recommend keeping this separate from #329/#331/#332 until a capture says otherwise.

PR #456, opened 2026-08-04
Cure the BG-exit movement lockup: synthesize the unroot a departing player never received (#328)

jameopotato:
## Problem

After leaving a battleground — reported specifically by players who **spam-click "Leave Battleground" the instant it appears** (some macro it) — the player arrives able to turn, mount and cast but **unable to move or jump**, permanently. `/reload` or relog cures it. The #328 world-transition family; recurring for months, patched around ~9 times without a root cause.

**The established chain** (rounds of client/server RE with Mirasu + live harness runs on PTR):
1. Every Kronos arrival delivers a root ceremony to the player: `SMSG_MOVE_ROOT` ×2 at the loading boundary, `SMSG_MOVE_UNROOT` ~1s later — wire-verified on 18/18 arrivals across two machines, every leg client-acked when healthy.
2. An instant leave departs while the BG-end root is being removed. The server's unroot fires in the between-maps window (`oldmap->Remove` → `map->Add`) and is **silently discarded** — cmangos `Unit::SendMoveRoot`, `Unit.cpp:751` `if (!IsInWorld()) return;`. Deterministic, not a random drop.
3. The client arrives still force-rooted; the server believes the player mobile, so nothing root-related is ever sent again. Rooted = can turn/cast/mount, cannot translate or jump — the exact symptom table.

**Harness-proven equivalence:** swallowing exactly one arrival unroot reproduces the reported bug precisely, including the `/reload` cure (the client rebuilds derived state; live-verified that reload does NOT re-run mover init, killing the alternative explanations).

## Change

Pure logic + evidence in `World/Client/WorldEntryCeremony.cs` (TransportClearGate extraction precedent).

**(a) The cure — carried-root reconcile.** `GameSessionData.ClientBelievesRooted` tracks the proxy's record of what the client was last told about its root state: set when a self root of either family is forwarded, cleared on either-family unroot (live-verified: the 42597 client honors spline-family unroots against force-roots too). If the player crosses a loading boundary still believing itself rooted, the proxy synthesizes the missing force-unroot:
- **Cross-map** (`NEW_WORLD`): armed at NEW_WORLD, decided at the player's first destination update, delivered **after the full UPDATE_OBJECT forwards** (the stuck-stun delivery pattern — the unroot can never race ahead of the create it cures).
- **Same-map** (hearth / mage teleport / portal within a continent — these produce `MSG_MOVE_TELEPORT_ACK`, no NEW_WORLD): armed at the self teleport, delivered at the client's `CMSG_MOVE_TELEPORT_ACK` (proof the teleport was processed).

Sentinel counter `0xFFFFFF02`; the client's ack is **swallowed** (logged as `worldentry.carried_root.cure_acked` — its presence is the proof of application) so the legacy server never sees an ack for an op it didn't send (spurious force-acks can feed Kronos's malformed-input kick counters).

**Design note — belief-ONLY gate, deliberately.** The first implementation double-checked the destination's movement flags and could **never fire**: a player's server-side `m_movementInfo` is an echo of the client's own reported state (including the root-ack it sent when it got rooted), so the stranded client poisons exactly the evidence that was consulted (verified live; the stuck-stun golden capture's `Flags=0x400` create is the same echo). Safety without the check: a legitimately rooted cross-map arrival is re-rooted immediately by the server's own arrival ceremony (fires on every arrival, 18/18); the same-map residue (hearthing away during the final seconds of a live root aura) is cosmetic. The gate fires **zero** times across every healthy capture we hold — nobody crosses a boundary rooted in normal play.

**(b) Always-on unclosed-ceremony breadcrumb.** Per-arrival accounting of ceremony legs (force root/unroot forwards + acks, spline-family legs, mover-init), flushed at the next boundary (next login-verify / transfer / legacy disconnect — the disconnect anchor added in review so a quit-and-relog doesn't eat the one breadcrumb we want). An opened-but-not-observably-closed ceremony logs ONE `worldentry.ceremony.unclosed` line — so any future residual lockup arrives in a normal Export Diagnostics already labeled with its failure mode (never-sent vs client-swallowed vs wrong-family), DebugOutput not required. Healthy ceremonies log only under DebugOutput. No timers anywhere; ≤1 line per transition.

One config key added in review: kill switch `WorldEntryCarriedRootCure` (default true; gates the cure synth only — the breadcrumb stays active regardless). No launcher changes, no behavior under healthy traffic beyond counters and the conditional breadcrumb line. The investigation harness (delivery-manipulation modes used to prove all of the above) lives only on the `worldentry-ceremony-tools` branch and does not ship.

## Verification

- **Suite 824/824** after the beta merge-in (10 new tests: unclosed/anomalous truth tables anchored to the healthy 18/18 wire shape and the stuck-stun golden capture's missing-ack fingerprint; belief-only gate; sentinel disjointness from the legacy constant 0 and the 0xFFFFFFFF transport-clear sentinel). Originally 791/791 pre-merge with the same 10 tests.
- **Red case, live:** harness swallowed one login unroot on PTR → player fully movement-locked (move/jump dead, turn/mount/cast fine), `/reload` cured — the reported bug, manufactured on demand.
- **Green case, live (this code):** same manufactured strand → **hearth and mage teleport each arrived auto-cured, zero player action** — wire: `worldentry.carried_root.armed` → `worldentry.carried_root_cured` (+7ms after the client's teleport-ack) → client ACKED the synth unroot.
- **Field gate (ship-and-observe):** BG spam-click lockup reports should cease. Any residual occurrence now self-documents via the breadcrumb; an `unclosed` line with zero unroots of either family would indicate the in-arrival variant (server-side, would become a precise upstream report to the Kronos dev).

## Risk

- The synth unroot fires only when the tracked belief says the client is rooted at a boundary crossing — zero firings across all healthy captures; unroot-on-unrooted is a proven no-op; real roots are re-asserted by the server's own arrival ceremony.
- The ack swallow is keyed to the sentinel range only — legitimate acks (Kronos's constant counter 0) can never match.
- Stale belief after a `/reload` mid-root costs at most one harmless no-op unroot at the next boundary (fail-safe direction, documented).


jameopotato:
Review of fa34e304 (read-only pass; suite not independently re-run — main checkout is occupied by the #454 branch and this branch is locked in its worktree; test arithmetic verified instead, see nit at bottom). Verdict: **support merge** — the red/green live proof (manufactured strand → exact reported symptom incl. /reload cure → auto-cured on the wire with the client acking the synth unroot) plus the deterministic `!IsInWorld()` discard chain is the strongest evidence class available short of a natural capture, and the fail-safe direction holds everywhere I traced (belief-only gate, no-op unroot when wrong, server re-roots legitimate roots via its own 18/18 arrival ceremony, sentinel disjoint from the constant-0 legacy counters and the 0xFFFFFFFF transport-clear sentinel). Three points before/at merge, one call-out for the maintainer:

**1. Spline-family contradiction — needs a one-line authoritative answer (pre-merge).**
`WorldEntryCeremony.cs` (tracker comment) states Mirasu R40 branch (c) as fact: spline-unroot is *"applied to a different slot than a force-root — a force-root + spline-unroot pair strands the root."* But `Client/MovementHandler.cs` (`HandleMoveSplineSetFlag`) clears `ClientBelievesRooted` on a self spline-unroot citing *"R5-proven: the client accepts a spline-family unroot as clearing."* Both can't be true. If R40(c) is right, the belief-clear on spline-unroot silently disables the cure in exactly the BG-end `CLIENT_CONTROL_LOST` window this PR targets (force-root carried, spline-unroot forwarded → belief=false → cure never fires). The code follows the live R5 result, which I'd trust over static RE — but please state which is authoritative and fix the stale tracker comment so the falsified lore doesn't get re-derived later. (Mitigation either way: any self spline leg trips the anomaly breadcrumb, so the field would tell us.)

**2. First-update-object anchoring — known-fragile pattern, cheap hardening (fine as fast-follow).**
`UpdateHandler.cs`: `WorldEntryPendingCarriedRootCheck` is consumed at the **first movement block of any unit** after NEW_WORLD (same region as the deferred transport synth), and the cure unroot is delivered at that packet's end. If the player's create is in a *later* UpdateObject packet, the synth unroot lands pre-create and is discarded — with `ClientBelievesRooted` already cleared, so the cure is lost for good. This is the exact anchoring fragility documented on the zep transport synth (fires pre-create on re-attach logins, first update = some other unit). Safe in practice here because it arms only on transfers, where the destination streams the self-create first — but the established recommendation from that prior art is anchor-on-self-create-forwarded, and it's a small change.

**3. Relog eats the unclosed breadcrumb (fine as fast-follow).**
The ceremony is flushed only at the next `TRANSFER_PENDING` or login-verify *within the same session*. Disconnect teardown never flushes, and a fresh `GameSessionData` gets a fresh tracker — so the most common user reaction to the lockup (quit/relog) writes **no** `worldentry.ceremony.unclosed` line, which is precisely the capture we want from the field. A `FlushWorldEntryCeremony("disconnect")` in the teardown path closes the gap. (`/reload` and hearth-away paths are unaffected — those flush at the next boundary.)

**4. For the maintainer — explicit sign-off requested on the always-on exception.**
`worldentry.carried_root_cured`, `worldentry.carried_root.cure_acked`, and `worldentry.ceremony.unclosed` are deliberately **not** DebugOutput-gated — that's the ship-and-observe instrument (reporters run DebugOutput off), it's bounded (≤1 line per transition, fires only on cure/anomaly), and healthy traffic logs nothing. Justified in my view, but it deviates from the standing "DebugOutput-gate always" rule, so it should be an explicit call rather than an assumption.

Nit: the body says "8 new tests" — `WorldEntryCeremonyTests.cs` contains **10** (`[Fact]` count), which is what reconciles 781 (beta) + 10 = the claimed 791. Worth correcting the body so the arithmetic reads clean.

Scope note for the record: this cures the *carried-root* family (departure-side loss, R40(a)) and any variant where the proxy's forwarded state and the client's applied state agree. The arrival-side variant — client discards a genuinely-forwarded unroot mid-load — is deliberately *detected, not cured* (belief says unrooted; breadcrumb labels it `unroot forwarded, never acked`). If the hardware-dependent residual is that variant, reports won't fully cease, but every occurrence now arrives labeled in a normal Export Diagnostics.

PR #458, opened 2026-08-07
Refresh build-single/ distribution to 5.2.0-beta.2

jameopotato:
## What

Rebuild the tracked single-file distribution under `build-single/` from the current beta tip (`ade2b73f`, 5.2.0-beta.2) and copy the full publish output over the stale committed artifact.

## Why

`build-single/JimsProxy.exe` was **5.0.6-latency-fixes.1** (`ff668d11`, Apr 30) — many releases behind, and paired with April-era CSV data the current code no longer matches. Anyone grabbing the in-repo single-file build got a months-old proxy.

## How

`scripts/Build_Windows.bat` recipe — `dotnet publish HermesProxy -p:PublishSingleFile=true -r win-x64 -c Release --self-contained` (single-file, self-contained, win-x64) — then the publish output (`JimsProxy.exe` + `HermesProxy.config` + `CSV/` + `CSV/Hotfix/`) copied into `build-single/`.

## Diff (16 files, forward-only — nothing removed)

- **JimsProxy.exe** → 5.2.0-beta.2 @ `ade2b73f`
- **CSV**: +8 new files the newer exe reads (`CreatureDisplayInfoVanilla`, `CreatureFamily`, `SpellAuraEffects1`, `SpellChanneled1`, `SpellRankChain`, `TalentSpellRanks`, plus Kronos `Item1`/`ItemSparse1` hotfixes), 8 refreshed, 0 deleted
- **HermesProxy.config**: gains `ThreatEngine=false` and `ServerType=Kronos`; all other defaults unchanged

Distribution artifact only — no proxy source or behavior change. (GH large-file warning expected; matches the existing tracked-binary pattern.)


jameopotato:
Holding per review — beta.3 shipped 08-13 and v5.2.0-beta.4 has just been cut, so this snapshot lands two betas stale. Please refresh build-single/ against the v5.2.0-beta.4 tag and re-point this PR. Separate question worth a look: nothing in the repo, scripts, or launcher references build-single/ — we may want to move this distribution to GitHub Releases and .gitignore the folder rather than committing an ~80MB exe per refresh.

jameopotato:
Superseded by #494, which retires `build-single/` rather than refreshing it.

Recapping the reasoning from the review comment on this PR, now with numbers:

- `build-single/` is **100 files, ~115 MB on disk**; `build-single/JimsProxy.exe` alone is **80,146,765 bytes**. GitHub reports the whole public repo at ~85 MB, so the directory is essentially the repo's entire weight — paid by every clone and every `git submodule update` the launcher build performs.
- **Nothing references it.** `git grep build-single` is empty across this repo (scripts, workflows, README) and the launcher repo. Checked twice, two days apart.
- **GitHub cannot download a directory**, so using it meant cloning the whole repo with git — worse than the one-click archive it was meant to be.

Refreshing it, as this PR does, would add a *second* ~80 MB blob to history permanently and need repeating every release.

The replacement is a release asset. **`JimsProxy-v5.2.0-beta.4-win-x64.zip` is already live on the v5.2.0-beta.4 release** (~38 MB, with setup notes), and a separate branch repairs `.github/workflows/Release.yml` so every future tag gets one automatically — that workflow already builds and attaches assets, it had just failed all 10 of its runs.

Leaving this open for you to close rather than closing it myself.

jameopotato:
Closing in favour of #494, which retires `build-single/` instead of refreshing it.

This PR is correct maintenance of the wrong mechanism: it would add a second ~80 MB exe blob to permanent history, and need repeating every release. It has also gone stale — it targets 5.2.0-beta.2 (July), and we have since shipped v5.2.0 stable and v5.2.1-beta.1.

The prebuilt proxy now ships as a release asset (`JimsProxy-<tag>-win-x64.zip`, on v5.2.0 and v5.2.0-beta.4) rather than as a tracked directory.

Branch `chore/build-single-5.2.0-beta.2` is left in place; nothing is lost.

PR #459, opened 2026-08-09
Cure the capped instance-reset bad login: eviction merge + loading-notify map fix (+ login op hardening)

### LINKED ISSUE #457: Fast relog after in-instance /camp: server returns a still-logout-stunned player with a partial world; state persists across relogs
## Summary

Using the solo dungeon-reset trick (log out inside an instance, get kicked from the group while offline, log back in at the instance start), roughly 1 in 5 re-logins comes back broken: the world is missing most of its objects ("all mobs invisible, including the scepter portal"), or in a variant we have not yet captured, the loading screen never completes. **Once one login breaks, subsequent re-logins also break** — the state persists server-side for 10+ minutes and across a full proxy restart — until the player stays offline long enough for it to clear.

Reported from Maraudon Princess farming on Kronos V (reporter reproduces ~1/5 nightly). Two-account repro procedure confirmed working by us (the trick part; the broken state not yet self-reproduced — GM instant-logout and PTR caveats apply to our attempts).

## The broken-login fingerprint (uniform, 7/7 captured failures)

| dimension | healthy (23 logins) | broken (7/7) |
|---|---|---|
| player create `UNIT_FIELD_FLAGS` | `0x1008` | **`0x41008` — `UNIT_FLAG_STUNNED` (0x40000) still set** |
| stun aura present | — | **none — flag without aura = the artificial logout stun** (#431 detector: `artificial_logout_stun: true`) |
| self-create delay after `SMSG_LOGIN_VERIFY_WORLD` | 11–190 ms | **818–832 ms** (tight band, every wedge-entry login; echoes the +844 ms init drift in the 2026-07-19 stuck-stun golden capture) |
| NPC creates in first 12 s (Maraudon entrance) | 41–61 | **12–15 (~25 %)** — measured at proxy input, i.e. the server sent a partial world |
| wall-clock wait before re-login | 16–157 s all healthy | 13 / 29 / 40 / 53 / 88 s — **uncorrelated** |

The chain evidence: three consecutive broken logins spanning ~10 minutes and a **proxy restart** (fresh process, fresh sockets, fresh session — first login of the new process still broken).

## Why this is pinned server-side

1. `UNIT_FIELD_FLAGS` and logout-timer state are runtime-only — no DB column can deliver `STUNNED` into a fresh login's create. A stunned create **proves** the server built it from a still-live, still-mid-logout player object.
2. The create shortfall and the ~825 ms stall are both measured at the proxy's *input* from the server.
3. Persistence across a full proxy restart rules out proxy session state as the carrier.
4. In cmangos terms (interpretation, not a claim about Kronos internals): this matches `HandlePlayerReconnect` — reachable only while the session still holds a logging-out player — except that on Kronos the path evidently does **not** clear the logout state on completion (cmangos does, `CharacterHandler.cpp:1065`), which would make the wedge self-sustaining exactly as observed.

## Disclosed proxy-side residuals (adversarially reviewed, none affect the pinning above)

- Our relogin flow is not native-shaped (old legacy TCP closed + new connection authenticated with the cached session key ~450 ms before `PLAYER_LOGIN`, vs a native client's single persistent socket). It runs identically on all 23 healthy logins, so it is not sufficient to cause this — but a native-1.12-client control run has not been done, so "necessary co-factor" is not excluded.
- #431's cure synthesizes `CMSG_LOGOUT_CANCEL` on each wedged login. We cannot yet rule out that this interferes with the server's deferred logout finalization (i.e. participates in the wedge's *persistence*). Testable by disabling the #431 strip/cancel for a run.
- The reporter's procedure includes a group-kick while offline; all reporter captures ran with DebugOutput off, so whether the kick is required for wedge entry is unconfirmed.

## Proxy status

#431's layers correctly detect and strip the baked stun/root on every broken login (the movement half of the symptom is already cured). The partial world is **not proxy-curable**: the creates never arrive, and vanilla has no re-request protocol. Best available proxy-side actions: an always-on anomaly breadcrumb (fires only on the fingerprint) and reconcile-on-ready from the object cache (world-entry contract stage 4, speculative).

## Artifacts

Reporter (Kronos V, DebugOutput off): `jimsproxy20260804150729.jsonl` (L6 broken), `Mara Bad Logins/jimsproxy-20260804-214504.jsonl` (L10), `…-234452.jsonl` (L1+L2 chained), `Bad Login 3/jimsproxy-20260806-112244.jsonl` (L6), `…-122321.jsonl` (L1+L2 chained across restart). Ours (healthy baseline, DebugOutput+PacketsLog on): `jimsproxy-20260806-211548.jsonl` + `modern_42597_1786076161.pkt` (full wire of a healthy trick cycle), `jimsproxy-20260806-221155.jsonl`.

## Next steps

1. Hand the simple repro + fingerprint to the Kronos dev (three-client repro on a local core; watch what the login path does when the old player object is still mid-logout, and whether that path clears the logout state).
2. Native-1.12 control run (closes residual 1).
3. Optional A/B with #431 strip/cancel disabled (closes residual 2).

https://claude.ai/code/session_015RAux1Ane71zQ6U8jRwihd

jameopotato:
This PR now cures **both halves** of the capped instance-reset bad login (Maraudon/Deadmines reset trick at the 5/hour cap): the stuck-loading first login AND the "stunned" (no turn/no cast) + invisible-world state. Two commits, three mechanisms.

## Commit 1 — Login-eviction merge (the stuck-loading half)

Unchanged from the original PR (see the first revision of this body / CHANGES.md): when a login names an instanced map, hold the client-facing stream; if the over-cap eviction's `TRANSFER_PENDING`+`NEW_WORLD` follows, rewrite the held login-verify to the destination and synthesize the worldport-ack — the client does ONE clean load instead of the login-then-instant-transfer that permanently hangs its loading screen. **Field-verified 2026-08-09** (merged event on the wire, dismissal +568ms, no force-close; healthy logins unchanged).

## Commit 2 — The map-header fix (the stunned + invisible-world half, THE root cause)

`Server/PacketHandlers/CharacterHandler.HandleLoadScreen` trusted the client's `CMSG_LOADING_SCREEN_NOTIFY` MapID unconditionally. That notify echoes the map the loading screen **opened** with — the char-list map (36, inside the instance) — not where the server actually put the player (map 0 after the eviction). And `UpdateObject.Write()` stamps `CurrentMapId` into the **MapID header of every `SMSG_UPDATE_OBJECT`**, which the 1.14 client validates against its own map and **silently discards on mismatch**. Consequences from the loading screen's dismissal onward, on every mismatched login:

- The server's stun-clear values-update never applied → the client rationally kept believing it was stunned → **no turn, no cast, "You can't do that while you are stunned"**. Walk/strafe still worked because unroots ride force-ops that carry no map header — the corrupt "half-stun" explained.
- No creates applied → **invisible mobs / partial world** (the reporter's original symptom, the client-side component of issue 457).
- `/reload` can't fix proxy-side corruption; relog sends a fresh, correct notify → cured. The wild "~1 in 5" = whether the char-list map happened to be stale at that relogin.

The fix is a one-line guard: only accept the client's notify map when `CurrentMapId == null` (session seed); the server-derived sources (`LOGIN_VERIFY_WORLD` / `NEW_WORLD` / `INIT_WORLD_STATES`) are authoritative and always follow.

**How it was found:** the client-side hypotheses (construction latch, op-ordering — Mirasu R55/R56) were falsified in vivo first; then an interleaved same-evening build matrix on the live repro (two characters, slot-switch comparisons minutes apart) split clean/locked perfectly by build: **5/5 clean with the guard, 0/4 without** (18:43, 22:11, 23:14 locked; 21:05, 21:13, 21:14, 22:15×2 clean). The smoking-gun capture shows `showing=false, map_id=36` arriving while the client stands on map 0, with every subsequent proxy packet built from the corrupted value. Test-identity was verified three ways (lifecycle exe paths, binary hashes + string markers, runtime event fingerprints).

## Commit 2 also ships — PreCreateOpHold (login hardening)

Built as the suspected stun cure under the R56 ordering hypothesis, falsified in the field, **kept deliberately**: the wedge login's pre-create control ops went permanently unacked (R49 dead-zone behavior); with the hold, the client acks every force-op leg (previously 2 roots → 1 ack; now 4/4 at ~+377ms). Zero cost on healthy logins — their ops already arrive after the create (12/12 corpus), so nothing is ever held. Kill switch `LoginPreCreateOpHold` (default on), packet-driven releases only (self-create forward / login-failed discard / disconnect fail-open), 13 red-proven state-machine tests. This is the exact configuration that went 5/5 in the field; shipping it as tested.

## Verification

- Suite 814/814 (20 merge tests + 13 op-hold tests + baseline), both new state machines red-proven by mutation.
- Field: the interleaved matrix above, plus both merge-path and no-transfer wedge shapes covered.
- Field gate before tag: capped wedges show turn+cast immediately with **full mob visibility** (the invisible-world half should be cured by the same guard — explicit check requested), healthy logins byte-identical behavior.

## Review notes

- The `IsFirstEnterWorld`/transport-synth/speed-reassert divergences of the merge are documented in the first commit (unchanged).
- `login.eviction_merge.merged`, `login.precreate_op_hold.released` and their anomaly events are always-on (fire only on the bug); per-login events DebugOutput-gated.
- Not in this PR: the server-side wedge itself (Kronos re-attaches a still-logging-out player and stalls its create ~600–800ms) — that remains upstream-reportable; this PR makes the client experience of it fully clean.

https://claude.ai/code/session_01EN4xDoP47Ubg8nC8M4wgXG


jameopotato:
## Field verification 2026-08-09 — PASS (loading-screen half)

Capped Deadmines reset repro on this branch build (`5.2.0-login-eviction-merge.1+15(a880b5f)`, DebugOutput+PacketsLog ON, session `jimsproxy-20260809-002143.jsonl` + `modern_42597_1786260124.pkt`):

**The previously-stuck login (capped):** hold armed at LOGIN_VERIFY map 36 (stale inside pos -14.57,-385.48 — the exact wedge fingerprint) → `login.eviction_hold.transfer_pending` +19ms (map 0) → **`login.eviction_merge.merged`** +21ms (dest -11208.7, 1675.9 = Deadmines entrance, 18 held packets, hold_ms 32) → client single-loaded and **dismissed the loading screen at +568ms** (`loading_screen_notify(showing=false)` on the wire; INIT_ACTIVE_MOVER_COMPLETE +465ms). Player visually confirmed loading outside the instance. **No force-close needed — the stuck-loading symptom is gone.**

**Healthy control (uncapped trick login, same session):** `login.eviction_hold.started` → `released_healthy` (22 held packets, hold_ms 0) — single load, unchanged behavior. Cure relog afterward (continent map 0) correctly armed nothing. No `transfer_aborted` / `flushed_fail_open` anomalies.

**Known-limit half, as disclosed in the PR body:** the wedge's stun still rides the merged login (create raw `262152` (0x40008) + `artificial_logout_stun=true`; #431 stripped the create, server cleared the flag +1204ms, player still turn/cast-locked until relog). That is the separate construction-latch fix (step 2, client-RE pending) — this PR never claimed to cure it.

One observation for the record: the synthesized worldport-ack makes the server emit its transfer-arrival ceremony (ROOT+CONTROL_UPDATE+UNROOT at +187ms) in addition to the login ceremony, all pre-INIT_AMC — same dead-zone class as the old flow (which delivered the same ceremony pre-init too), no observed regression, noted as the composition point for the stage-3/step-2 work.

jameopotato:
Second commit pushed (187f3e6d): the stunned/invisible-world half is cured — root cause was OUR pre-existing loading-notify trust corrupting the UpdateObject MapID header (client silently discards mismatched updates post-dismissal, so it never saw the stun-clear or any creates). PR body rewritten with the full mechanism, the interleaved 5/5-vs-0/4 field matrix, and the op-hold's reframe from suspected-cure to kept-hardening. The earlier field-verify comment's 'stun half is a separate fix' framing is superseded.

PR #464, opened 2026-08-13
Clear the auto-attack handshake on an empty-victim ATTACK_STOP

### LINKED ISSUE #450: Killing white blow presents seconds late (floating text + swing sound after loot window) — trailing ATTACKER_STATE_UPDATE lands after #389's preempt stop

jameopotato:
## Symptom

Charge a mob that is **evading/leashing back from another fight** and your character stops being able to fight it — no melee swings at all, and abilities stop working too. Movement stays completely normal, so it reads as a broad "combat lockout". Reported 2026-08-12 (warrior, charged a wolf that had just killed a critter); logged out ~35s later still unable to act.

## Root cause

Kronos answers the Charge with `SMSG_ATTACKSTOP(player, 0)` instead of `SMSG_ATTACKSTART` when it refuses the engage. The modern client re-sends `CMSG_ATTACK_SWING` ~200ms later.

`HandleAttackStop` only cleared the swing handshake when the stop's victim **matched** `CurrentAttackTarget`. An **empty** victim matched nothing, so it fell into the `else` "target-switch — preserve the new target" branch and cleared nothing. `CurrentAttackTarget` stayed pinned to the mob and `WaitingForAttackStart` stayed true, so the de-dupe guard in the `CMSG_ATTACK_SWING` handler ate the retry **and every later one**. The swing never reached the legacy server again. Only `SMSG_CANCEL_COMBAT` could unwedge it, and it never arrived.

A stop naming **no** victim can't be a stale stop for the old target of a switch — it means "you have no attack target at all". This treats it as a rejection and clears the handshake. Over-clearing is the safe direction: worst case is one redundant `CMSG_ATTACK_SWING` forwarded, which is exactly what the client asked for.

This is the empty-victim sibling of `8c2c34e6` ("auto-attack stuck for ~30s when server rejects CMSG_ATTACK_SWING"), which fixed the named-victim form and explicitly assumed a non-matching victim implied a target switch. **That fix shipped without tests**, which is why the hole survived three months.

## Why the casts died too — same cause, not a second bug

Warrior rage income is ~6–9 per swing. With zero swings the only income was ~0.85 per incoming hit. Sunder Armor costs 15 rage (measured on the wire: 24.3 → 9.3 across the cast). Rage sat at **9.3** after the second Sunder and didn't cross 15 again for **8.5s** — every press in that window is refused *client-side* and emits no packet at all, which is why the capture looks silent while the player was pressing buttons.

## Wire evidence

`modern_42597_1786595484.pkt`, player low=27032, wolf low=4628, Charge at t=0:

```
+0.000  C2S CMSG_ATTACK_SWING(4628)          forwarded
+0.219  S2C SMSG_ATTACK_STOP(player, EMPTY)  legacy op 324; victim mask bytes are 00 00
+0.422  C2S CMSG_ATTACK_SWING(4628)          swallowed by the de-dupe guard
```

- **Zero** player `SMSG_ATTACKER_STATE_UPDATE` for the remaining 26s (~8 expected at his 3.20s median swing). Every ASU in the fight is the wolf hitting him. His last swing in the whole session was 18.9s *before* the Charge.
- Of the **8 Charges** in that session, the 6 answered with `ATTACK_START` were fine; the **2** answered with an empty-victim stop wedged — both against a mob that had just finished fighting another mob.
- The earlier one (t−386s, mob 5477) recovered only because a second Charge produced a real `ATTACK_START` 30.4s later. During those 30.4s: zero swings, zero casts, rage decaying 21 → 0. In the wolf fight he was in combat so Charge was unavailable — it never recovered.
- `combat.attack_stop_preempted` did **not** fire in the window, so #450/#452 are not involved. The stop is `packet.in s2c` — genuinely from Kronos, not synthesized.

## Change

Decision logic moves to `GameSessionData` as two pure operations — `TryBeginLocalPlayerAttackSwing` / `ApplyLocalPlayerAttackStop` — so it is testable without a socket, matching the existing `TryClearSettledAttackTargetOnDeath` precedent. The socket layer keeps only the "forward a deferred stop" decision.

Behaviour is otherwise unchanged: deferred-stop flush, settled-target clear, matching-victim rejection and target-switch preservation all keep their existing semantics, each now with a regression test.

## Testing

Red-green-red performed, per house rule:

1. With the fix: **854/854 pass** (847 baseline + 7 new).
2. Reverting **only** the `|| stopVictim == WowGuid64.Empty` clause: `ChargeAtEvadingMob_EmptyVictimStop_DoesNotWedgeAutoAttack` fails with `PreserveTargetSwitch` instead of `ClearRejectedHandshake` — the exact fall-through branch — and **no other test moves**.
3. Restoring it: **854/854**.

The regression test replays the captured sequence verbatim and asserts the second swing is forwarded.

**Suggested in-game check:** aggro a mob onto a critter (or walk it out of leash range), then Charge it during its return. Before: no swings for the rest of the fight. After: auto-attack engages normally. Field workaround for anyone who hits it on an older build — attack a *different* target once, which passes the de-dupe guard.

https://claude.ai/code/session_015BwjtWNMgHDNdk6DSjKPEz


jameopotato:
## Correction to this PR's framing (not the fix)

Two claims in the description and commit message overstate what the capture supports. The packet facts quoted are accurate; the interpretation built on them was not. Flagging it here so the record is straight.

**1. "2 occurrences" conflates two different things.** There were **two wedges at the packet level, but only one experienced lockout.**

The earlier one (t−386 s, mob 5477) was asymptomatic. Across that 30.4 s window rage decayed net monotonically 21.0 → 0.0, and there were **zero `ATTACKER_STATE_UPDATE` in either direction** — rage only decays out of combat, and nobody was hitting anybody. So the player charged, the engage was refused, he pressed attack once more (the swallowed swing), nothing happened, and he was **out of combat the whole time** until re-charging the same mob 30 s later. That is a charge that whiffed, not a lockout. The reporter confirms he only ever experienced one.

Contrast the wolf fight, where rage *climbed* 9.3 → 22.4 under an ASU every 2.0 s.

**2. "The mob had just finished fighting another mob" is correlation, not established mechanism.** Two samples. I don't actually know what makes Kronos emit `SMSG_ATTACKSTOP(player, 0)` here, and I presented that inference with more confidence than the evidence carries.

A field repro attempt since then sharpens it: charging a mob **early** in its leash (just turned to run back, deep in evade) got the Charge **rejected outright** — no animation, no `SPELL_GO` — so nothing wedged, and the player engaged normally after catching up. In the wolf case the Charge **executed** (`SPELL_START` → `SPELL_GO` → `SPELL_ENERGIZE_LOG`, rage 16.1 → 37.1) and the empty-victim stop came in that same 219 ms burst. Working hypothesis is therefore the evade→normal **transition** — the wolf was nearly done leashing and started swinging right after — rather than evade generally. That is a hypothesis, not a mechanism, and it is not reproducible on demand.

**3. Consequence for the notes.** This is rarer than the description implies, and the symptom only persists when the mob happens to re-aggro you (keeping you in combat so Charge is unavailable to self-heal it). It should sit low in any patch-note ordering, not read as a commonly-hit fix.

## What is unaffected

The defect itself. `ATTACK_STOP(player, EMPTY)` arriving while `WaitingForAttackStart` is true takes the target-switch fall-through, pins `CurrentAttackTarget`, and the de-dupe guard then eats every retry — proven at byte level (victim mask bytes `00 00`), with the red-green-red on the regression test unchanged. **The fix does not depend on knowing the server-side precondition**: whatever causes the empty stop, the proxy has to recover from it, and before this change it never did.

Follow-up: adding an always-on `Log.Event` for this condition so field frequency can be measured without needing a repro — the repro gate can't realistically close on a timing window this narrow.

⚠️ The source artifacts (`modern_42597_1786595484.pkt`, `jimsproxy-20260812-213115.jsonl`) were deleted when `Downloads\JimsProxy-beta3\` was redeployed, so the numbers above cannot be re-derived from the capture.

Issue #468, opened 2026-08-13
Observers see a player stuck in the melee attack pose after their kill — server-side attack is never stopped


jameopotato:
## Report (2026-08-13, two-box observation)

A hunter observing a warrior (separate proxy processes) saw the warrior remain in the drawn-weapon melee attack pose after the warrior's target (a wolf) died. The warrior's own client had left the pose normally; only observers kept seeing it. Longstanding — noticed before, reproduced during beta.3 gate testing.

## Mechanism

This is the third symptom of **#450 finding 1**, the deliberately-deferred half of the killing-blow work:

1. Since #389, the proxy synthesizes the local `SMSG_ATTACK_STOP` at kills, so the modern client — believing the server already stopped it — **never sends `CMSG_ATTACK_STOP`** (pre-#389 captures show the client sent it at every kill; post-#389 sessions show zero at kills).
2. Kronos therefore keeps the player's server-side auto-attack running against the corpse — proven by `ATTACKSWING_DEADTARGET` arriving 0.2–1.6s after **every** kill (9/9 in `jimsproxy-20260812-141139.jsonl`, 1/1 in `jimsproxy-20260813-134737.jsonl`).
3. A server that believes the attack is still running **never broadcasts `ATTACK_STOP` for that player to observers**. The local synth stop only reaches the killer's own client.
4. Observers' clients render the attack pose until some later attack-state edge for that player arrives (attacking a new target, logout, etc.).

Symptom ledger for the underlying defect: (1) red DEADTARGET error after every kill, (2) server-side corpse-swinging, (3) stuck attack pose for everyone observing a proxy user's kills. Native 1.12 clients are unaffected as attackers (they still send their own stop).

## Fix direction (one PR, both halves)

- **Root fix:** proxy-initiated legacy `CMSG_ATTACK_STOP` when the kill preempt fires — restores what the pre-#389 client did, stops the corpse swings, and produces the observer broadcast. Needs care around the #321 swing-start/target-switch state machine (why it was split out of #452).
- **Observer-side belt-and-suspenders:** melee equivalent of `RetractObservedShootersOnUnitDeath` (CombatHandler already lowers observed *shooters'* bows when their victim dies) — drop an observed attacker's melee pose on victim death, so our users see correct poses regardless of the attacker's client.

Refs #450, #452, #389.

PR #481, opened 2026-08-18
Post-kill upstream CMSG_ATTACK_STOP: close the per-kill attack-state desync

jameopotato:
> [!IMPORTANT]
> **⛔ HOLD — needs IN-FIELD verification before this merges (2026-08-18).** Stays draft until a live session on this build passes the field gate below: `combat.attack_stop_upstream_synth` on settled kills, `combat.attack_stop_upstream_echo_swallowed` ~200ms later, per-kill `ATTACKSWING_DEADTARGET` stragglers gone from the PacketsLog, and **no post-kill auto-attack re-engage regressions** (target-swap `ATTACK_STOP`+`SET_SELECTION`+`ATTACK_SWING` sequences must keep working). Scope note from the 2026-08-18 review: this is a **#389-regression hygiene fix** (the preempt suppressed the client's own kill-time stop, wire-proven by the pre-preempt 06-20 capture) — it makes **no claim on the stuck-button bug**, which predates #389.

## Symptom

After essentially every kill, the legacy server and the modern client spend 1.4–3.5 seconds disagreeing about whether the player is auto-attacking. Field shape (Jimothy, Silithus, 2026-08-18 capture): kill a mob with Execute → auto-attack stops and will not restart, ability presses feel dead, and in the fatal window the sunder-macro button sat stuck highlighted until death. This is the window where the post-kill stuck-highlight family (Sunder/Charge/Mount, "commonly after a kill") clusters.

## Evidence (`Downloads\JimsProxy-feared-sitting\`, `modern_42597_1787075164.pkt` + `jimsproxy-20260818-104556.jsonl`)

- Our kill-time preempt (`#450`/`#452`) sends the modern client `SMSG_ATTACK_STOP(NowDead)` — client-only. The client, told it stopped, never sends `CMSG_ATTACK_STOP` (modern auto-attack is server-authoritative).
- Kronos does not stop the killer's swing on victim death either: a **legacy `SMSG_ATTACKSWING_DEADTARGET` straggler follows ~every kill** (26 swing errors across ~26 kills), forwarded to the client as `AttackSwingError(DeadTarget)` 1.4–3.5s after the kill — proof the server was still swinging the corpse after we told the client it stopped.
- Fatal window (11:09:16→11:09:20): Execute kills the mob and zeroes rage in the same burst (`SMSG_POWER_UPDATE` decode: rage never exceeds 2.7 before death); target auto-clears; a live add is selected at +0.2s; then **zero c2s casts AND zero c2s attack commands for 3.7s** despite `/targetenemy` + `/startattack` + `/cast` macro presses. Rage explains the silent casts (Sunder 15/Cleave 20); nothing wire-side explains a silent `/startattack` with a live target — the client's attack state was wedged. The `DeadTarget` error landed mid-window (+459.2). The session's earlier structural replica death **without** a straggler produced no stick report.
- Scope honesty: the 08-03 **permanent** sticks (eat-context, `STUCK-MOUNT-BUTTON-20260803`) sit 20–28s from any kill or swing error — this change targets the kill-window desync class only, not that one.

## Fix

When the settled-attack preempt fires in `HandlePartyKillLog`, also send the legacy server the `CMSG_ATTACK_STOP` a real 1.12 client would have produced, so both sides agree the melee ended — the corpse-swing straggler (and its `DeadTarget` error mid-press) stops existing.

Kronos echoes that stop naming the corpse (wire-verified via this capture's retarget stops, which echo with the victim guid ~200ms later). `HandleAttackStop` consumes the echo via a new bounded pending set on `GameSessionData` (`RecordSyntheticUpstreamAttackStop` / `TryConsumeSyntheticUpstreamStopEcho`, pure) — **exact victim match only**, checked *before* the armed-preempt consume and `ApplyLocalPlayerAttackStop`:

- forwarding it would hand the client a duplicate stop;
- running the handshake bookkeeping could tear down a swing the player re-started within the echo RTT (the 2026-08-13 empty-victim wedge family);
- empty-victim stops always fall through unchanged (they can be real CC/death/refusal stops);
- ordering means a pathologically early echo can never eat the still-armed client stop;
- **review hardening (2026-09-05):** a pending entry is retired by the server's own `SMSG_ATTACK_START` naming the same victim (`InvalidateSyntheticUpstreamStopEcho` in `HandleAttackStart`). Static spawns keep their low guid across respawns on mangos-family cores and a player victim's guid never changes, so an echo that never came must not be left to swallow a genuine later stop on the re-engaged unit.

Diagnostics `combat.attack_stop_upstream_synth` / `combat.attack_stop_upstream_echo_swallowed` are DebugOutput-gated per the diagnostics policy.

## Kill switch

`PreemptAttackStopUpstream` (default **on**). `false` restores today's client-only preempt byte-for-byte.

## Tests

`AttackStopUpstreamEchoTests` (11, red-first — suite fails to compile without the primitives; stubbing the consume to always-false fails 5/8; a no-op invalidate fails the retirement test):
- exact-once consume; wrong-victim / empty-victim / nothing-pending negatives; record-empty ignored; FIFO bound eviction (8, oldest first);
- full kill → re-engage-inside-RTT → echo sequence leaves the new swing handshake untouched;
- disjointness: with nothing pending, a corpse-named stop mid re-engage still resolves `PreserveTargetSwitch` (existing #464/#321 semantics unchanged);
- ATTACK_START retirement: same victim retired, other victim left intact, nothing-pending / empty-victim no-ops.

Suite **972/972** (2026-09-05, on the v5.2.1-beta.1 tip, after the review hardening).

## Field gate (DebugOutput on)

Grind kills: expect `combat.attack_stop_upstream_synth` on each settled kill and `…_echo_swallowed` ~200ms later, and the per-kill `ATTACKSWING_DEADTARGET` stragglers to disappear from the PacketsLog. Watch specifically for post-kill auto-attack re-engage feeling unchanged-or-better (the swap sequences `ATTACK_STOP`+`SET_SELECTION`+`ATTACK_SWING` must keep working — covered by the exact-match + empty-fall-through design). Falsifier: any new failure to re-engage auto-attack after a kill → flip the kill switch and report.


jameopotato:
Field verification, 2026-08-30, build 5.2.1-fix-preempt-stop-upstream.1+6 (caf325d), Kronos V grind session (38 min, 50 kills; jimsproxy-20260830-084225.jsonl + modern_42597_1788104553.pkt):

- Zero ATTACKSWING_DEADTARGET packets in the whole session. Capture census: 33 swing errors, all NotInRange/BadFacing, reason DeadTarget zero times. On prior builds it followed roughly 70% of kills, at exactly lastWhiteHit + N x 3200ms (the weapon period).
- 34 of 50 kills had a settled auto-attack on the victim: 34/34 `combat.attack_stop_upstream_synth`, 34/34 `combat.attack_stop_upstream_echo_swallowed` about 200ms later. The capture shows exactly 34 player-attacker NowDead stops (the preempts) and no leaked echo duplicates reaching the client.
- The falsifier did not fire: 805 attacker-state updates and 64 organic player stop/starts through constant retargeting, no re-engage failures.
- Player-facing result: the blatant post-loot white damage pops (#450) are gone; one mild late swing remained, consistent with the client-internal animation deferral this PR does not touch.

Likely also the fix for #468 (the server-side corpse swing is what observers were seeing); that one needs an observer sighting to confirm and stays open.

PR #488, opened 2026-08-20
Extend RefireSpellGo: frame-clearing defer, observed casters, local cast-time, off-GCD excluded

Mongrul:
# What

Extends the opt-in `RefireSpellGo` mechanism (default OFF, unchanged) — the clean-frame duplicate-GO re-fire that closes the 1.14 client's coalesced-START+GO cast loop (stuck cast pose + looping cast sound + lit action button, survives `/reload`). Four changes, one bundle because they were soaked together:

1. **Frame-clearing defer — new `RefireSpellGoDeferMs`, default 50ms.** The old hardcoded 8ms in `SpellHandler` is *under* one 60fps frame (16.7ms), so the duplicate GO could land in the same client frame as the original — the exact frame it was supposed to clear. 50ms clears down to ~20fps crowds (where the bug is worst); the duplicate is inert, so a late land costs nothing.

2. **Observed-caster coverage.** Most stuck precast sounds are *other* players/NPCs — live-caught on an observed Flash of Light. A new `_observedGoStartPaired` flag is set in `HandleSpellStartOrGo` when an observed GO pairs with its seen START via `OtherCasterActiveCastIds` (the loop precondition); `HandleSpellGo` then schedules the same stripped duplicate GO (visual suppressed, empty targets/LogData). Auto-repeat excluded (its GO already has dedicated handling and a duplicate would double the shot visual).

3. **Local cast-time coverage.** The local gate changes from `pendingCast.StartedCastTimeMs == 0` (instants only) to `!GameData.IsChanneledSpell` — a spammed heal coalesces GO(N) with the *next* cast's START(N+1), the same end-event drop; the instants-only gate excluded the exact spell we caught looping live (Flash of Light). Channels stay excluded: re-firing a channel's GO restarts the channel visual.

4. **Off-GCD exclusion (both paths, `GameData.IsOffGcd`).** Field-caught regression during testing: off-GCD spells re-issue on GO, so a duplicate GO desyncs craft-all queues — heavy/runecloth bandages (3275/3276) stopped mid-queue. Every known loop target is on-GCD, so nothing is lost.

# Field validation

Six-day soak on Kronos (2026-08-13 → 08-19), Spell Queue mode (worst case for this bug), `RefireSpellGo=true`, DebugOutput on, including a full raid night on the 19th:

- **Zero looping cast sounds/poses across the whole soak** — first time the loop class has ever been beaten proxy-side (`RefireSpellGo` was inactive for every previously captured loop).
- Local refire fired 16/16 on spammed Flash of Light; observed refire 4/4, each paired exactly to one observed START (NPC and player casters). Zero parse failures, zero errors — the duplicate GOs perturb nothing.
- Crafting clean after the off-GCD exclusion (`cast.dropped.duplicate reason=off_gcd_double_send` gone).

Diagnostics: `cast.success_refire` / `cast.observed_success_refire` (DebugOutput-gated).

Default OFF, so zero behavior change for anyone who doesn't opt in. Build clean, suite 935/935 on beta.

# Interaction with #486

`_observedGoStartPaired` is set on the same pairing branch `fix/observed-castid-pairing` rewrites (`TryRemove` → `TryPairObservedGoCastId`). Merge either order — the resolution is one line: set the flag where the GO pairing succeeds. Happy to rebase this on top of #486 if you'd rather take that first.

# Not included

The observed-cast **eviction** (no-terminator class — deadline + supersede synthesis) is deliberately left out: it's soaking separately and hasn't produced a classified `evicted` fire yet, and its supersede edge would collide with #486's queue rework. It comes as its own PR once the soak proves it (per the two-fronts plan).

# Field evidence update (2026-09-04): the off-GCD exclusion is a bug fix to the shipped refire

A hunter on Kronos (~80 ms ping, beta 2026-08-23, `RefireSpellGo` on) hit **Feign Death stuck greyed for the rest of the session**, only with the option on. Log `jimsproxy-20260902-234809.jsonl`, three FD casts:

| Cast | Duplicate GO landed | COOLDOWN_EVENT landed | Result |
|---|---|---|---|
| 1 | GO +8 ms | GO +245 ms | fine |
| 2 | GO +21 ms | GO +55 ms | fine |
| 3 | GO +14 ms | same batch as GO (pressed mid-jump, aura broke in the cast's own server tick) | **stuck greyed** |

Feign Death is cooldown-on-event: the client greys the button on SPELL_GO and releases it only on SMSG_COOLDOWN_EVENT. When that event arrives before the duplicate GO, the duplicate re-arms the grey with nothing left to release it. Any defer loses this race. Beta's refire has no gate for it; the `GameData.IsOffGcd` exclusion in this PR is exactly the fix. Every player-castable cooldown-on-event spell in the 1.14 data (Feign Death, Stealth, Prowl, Vanish, Shadowmeld, Presence of Mind, Nature's Swiftness, Inner Focus, Cold Blood, Divine Favor, Amplify Curse, Combustion, Elemental Mastery, Lightning Shield) is off-GCD; every known loop target (Flash of Light, Shadow Bolt, Sunder, Battle Shout, Blade Flurry, Summon Warhorse) is on-GCD and keeps its coverage.

New commit `f1e53093` adds `HermesProxy.Tests/World/RefireOffGcdGuardTests.cs` (22 cases) pinning both halves of that data assumption. No code change. Suite 957/957. Test-rebased onto current beta: clean, no conflict with #486.

### DIFF
--- a/HermesProxy.Tests/World/RefireOffGcdGuardTests.cs
+++ b/HermesProxy.Tests/World/RefireOffGcdGuardTests.cs
@@ -1,0 +1,54 @@
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+/// <summary>
+/// RefireSpellGo off-GCD guard: the duplicate SPELL_GO must never be sent for off-GCD spells.
+/// Cooldown-on-event spells (Feign Death, Stealth, Vanish, Presence of Mind...) grey their button on
+/// GO and only release it on SMSG_COOLDOWN_EVENT; when the aura breaks in the cast's own server tick
+/// that event lands before the duplicate GO, which re-arms the grey for good (field-caught on Feign
+/// Death). These tests pin the data assumption the guard rests on: every player cooldown-on-event
+/// spell is in SpellOffGcd1.csv, and every known loop target the refire exists for is NOT.
+/// </summary>
+public class RefireOffGcdGuardTests
+{
+    static RefireOffGcdGuardTests()
+    {
+        GameData.LoadOffGcdSpells();
+    }
+
+    [Theory]
+    [InlineData(5384u)]  // Feign Death — the field-caught case
+    [InlineData(1784u)]  // Stealth
+    [InlineData(1787u)]  // Stealth (max rank)
+    [InlineData(5215u)]  // Prowl
+    [InlineData(9913u)]  // Prowl (max rank)
+    [InlineData(20580u)] // Shadowmeld
+    [InlineData(12043u)] // Presence of Mind
+    [InlineData(16188u)] // Nature's Swiftness (druid)
+    [InlineData(17116u)] // Nature's Swiftness (shaman)
+    [InlineData(14751u)] // Inner Focus
+    [InlineData(14177u)] // Cold Blood
+    [InlineData(20216u)] // Divine Favor
+    [InlineData(18288u)] // Amplify Curse
+    [InlineData(11129u)] // Combustion
+    [InlineData(16166u)] // Elemental Mastery
+    [InlineData(8788u)]  // Lightning Shield (rank 1)
+    public void CooldownOnEventSpells_AreOffGcd_SoRefireSkipsThem(uint spellId)
+    {
+        Assert.True(GameData.IsOffGcd(spellId), $"spell {spellId} must be off-GCD so RefireSpellGo skips it");
+    }
+
+    [Theory]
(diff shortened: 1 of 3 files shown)


jameopotato:
Refire SPELL_GO is not a path we want to keep progressing down. It's a symptom patch at best, and suppresses real evidence of the root cause at worst. 

Will remain as an opt-in option in the launcher's advanced settings for now, but could be fully removed later.

PR #490, opened 2026-08-20
Correct the shipped config defaults for standalone (non-launcher) use

jameopotato:
## What

Fix three drifted defaults in the tracked `HermesProxy/HermesProxy.config`, and document the two keys that had no documentation.

| Key | Before | After |
|---|---|---|
| `ClientBuild` | `40618` (1.14.0) | `42597` (1.14.2) |
| `PacketsLog` | `true` | `false` |
| `ThreatEngine` | *(no comment block)* | documented |
| `ServerType` | *(no comment block)* | documented, incl. `Kronos` \| `Generic` |

## Why

This file is what anyone **building from source or running the proxy without the launcher** actually gets. The launcher writes its own config from a template (`setup.rs`) and never reads this one, so the drift was invisible to every launcher user and hit only standalone users.

- **`ClientBuild 40618`** — development, testing, and the launcher's own client check all pin 1.14.2 build `42597` (`repair.rs:145`, `EXPECTED_CLIENT_VERSION`). A `ClientBuild` that doesn't match the running client fails login outright, so the shipped default was a guaranteed failure for the 1.14.2 client this fork targets. The comment block also listed 40618 first as "Default", which is where the value came from.
- **`PacketsLog true`** — every standalone session wrote a full packet capture to disk, unprompted and indefinitely. The launcher already forces this off; only standalone users were paying for it.
- **`ThreatEngine` / `ServerType`** — the only two keys in the file with no `Option/Description/Default` block. `ServerType` matters most: it silently selects the fork-specific item-data overlays loaded at startup (`GameData.cs:3170`), so a user who never learns it exists can't reason about their item data.

## How

Data-only. No code change — this file is read at startup by `ConfigurationManager`.

Deliberately **unchanged**: `ServerAddress` (`127.0.0.1`) and `ServerBuild` (`auto`) are correct generic defaults; the Kronos address belongs in the docs as a worked example, not baked into a file every fork-builder inherits. `ClientSeed` is the static-seed fallback (`RealmManager.cs:62`) and users should never edit it — real per-build seeds load automatically from `CSV/BuildAuthSeeds.csv`.

Per-fork wording on `ServerType`: `Generic` is described as scaffolding for future server support and explicitly **not** a tested configuration today, rather than presented as a supported option.

## Verification

- XML parses clean (`[xml]` load), BOM preserved, file remains **pure ASCII** (0 non-ASCII bytes — an em-dash was introduced during editing and removed).
- Values confirmed post-edit: `ClientBuild=42597`, `PacketsLog=false`, `ThreatEngine=false`, `ServerType=Kronos`.

**Field gate — not yet run.** A standalone run following only the documented manual steps (take the shipped config, set `ServerAddress`, start the proxy, log in) against Kronos. Test folder is staged; this PR should not be treated as field-verified until that passes.

## Context

First piece of a larger effort to make non-launcher use of the proxy actually work and be documented — README rewrite, `docs/` reference, and a setup page follow. Also supersedes the motivation behind #458 (refreshing the stale in-repo `build-single/` distribution), which is proposed for retirement in favour of a release asset.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,35 @@ A fork of [WowLegacyCore/HermesProxy](https://github.com/WowLegacyCore/HermesPro
 
 ---
 
+## 2026-08-20 — Correct the shipped config defaults for standalone (non-launcher) use
+
+**Issue:** `HermesProxy/HermesProxy.config` is what anyone building from source or running the
+proxy without the launcher actually gets, and it had drifted from the client this fork targets.
+`ClientBuild` shipped as `40618` (1.14.0) while development, testing, and the launcher's own
+client check all pin 1.14.2 build `42597` (`repair.rs:145`, `EXPECTED_CLIENT_VERSION`) — and a
+`ClientBuild` that does not match the running client fails login outright. `PacketsLog` shipped
+as `true`, so every standalone session wrote a full packet capture to disk unprompted and
+indefinitely; the launcher already forced this off, so only standalone users paid for it.
+Separately, `ThreatEngine` and `ServerType` were the only two keys in the file with no
+`Option/Description/Default` comment block, leaving a standalone user no way to learn what they
+do — `ServerType` in particular silently selects the fork-specific item-data overlays loaded at
+startup.
+
+**Change:** `HermesProxy/HermesProxy.config` — `ClientBuild` `40618` → `42597`; `PacketsLog`
+`true` → `false`; added full comment blocks for `ThreatEngine` and `ServerType`, documenting
+`ServerType` values as `Kronos` (default) and `Generic`, with `Generic` stated as scaffolding for
+future server support rather than a tested configuration. `ServerAddress` (`127.0.0.1`),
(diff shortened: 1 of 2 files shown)


jameopotato:
## Field gate: PASSED

The staged standalone test reached in-game with no problems.

**What this proves:** a user who downloads the proxy, edits **only** `ServerAddress`, and runs `JimsProxy.exe` themselves can log in — no launcher involved at any step. Specifically it clears the change under test: `ClientBuild=42597` is correct against the real 1.14.2 client, and no version-mismatch rejection occurs.

**Test composition:** 5.2.0-beta.4 exe (`f842db7d`) + `CSV/` + this PR's `HermesProxy.config`, with the single documented edit `ServerAddress` → `login.twinstar-wow.com`. Every other value left at the new shipped defaults, including `PacketsLog=false`.

This was the gate named in the PR body and in the CHANGES entry; both now read as verified rather than pending.

Issue #394, opened 2026-07-03
Cast sound and fx loop


AliziaKaline:
**Describe the bug**
Randomly (or at least it seems random) casting a spell, no matter which one, will make you stuck into infinite casting sound and fx. 

**Game Server**
Stonetavern, vMaNGOS, 1.12

**Hermes Proxy Version**
5.1.5-beta.5

**Hermes Config**
DebugOutput, PacketsLog, StructuredLog = true
VerboseLog = false
SpellCastEarlyFireOffsetMs = 0
RefireSpellGo = true

Video of the bug with logs : https://youtu.be/8KKttyw7lnU
Logs : https://drive.google.com/file/d/1pFBnC5U8jfk_jdcC9KILUUSUuXATayDV/view?usp=sharing
Spell triggering the bug (but could be any of them) : https://www.wowhead.com/classic/spell=1454/life-tap

jameopotato:
Fixed in **v5.1.8-beta.1** by #397. Your JSONL was the whole case: the GCD held-release timer (Windows ~15.6ms quantum) fired late, a fresh Life Tap press slipped every guard in the gap, and the server got two casts ~6ms apart — the loser's rejection then landed *between* the winner's START and GO, desyncing the pairing into the looping sound/fx you recorded. Two deterministic layers now close it: the stale held press is superseded instead of double-forwarded, and a transient rejection with no duplicate left to consume is dropped before it can touch the live cast.

It also explains your observations: `SpellCastEarlyFireOffsetMs` had no effect (the race is clock-quantization-driven), and `DebugOutput=true` "reducing" it was millisecond-scale timing perturbation.

@AliziaKaline — you have the best repro rate we've seen ("spam heal yourself"); if you can run **v5.1.8-beta.1** on Stonetavern, your session is the closing verification. If it recurs, please reopen with the JSONL.

AliziaKaline:
I just tested the 5.1.8-beta.1. The bug is still here, but seems to happen really less. I was able to play almost 3h without any problem.

Log : https://drive.google.com/file/d/1jrHfLPLf0pIMXyJ_IzvqbFiQIonkg-m5/view?usp=sharing
Spell : https://www.wowhead.com/classic/spell=905/lightning-shield

AliziaKaline:
Another one, on warlock.

Log : https://drive.google.com/file/d/15Uv52UGKRlAAuZ09WVrH6UE8E9Te_7aQ/view?usp=sharing
[...]

PR #491, opened 2026-08-21
Hold a dup press's CAST_FAILED until its started cast's terminal event (#394 frame-collision strand)

jameopotato:
# What

A new proxy-source strand of the #394 looping-cast class, found by per-cast autopsy of the reporter's 2026-08-14 log, and its fix: **a dup press's `CAST_FAILED` is no longer delivered while its same-spell STARTED cast is in flight — it's built exactly as before, held, and released immediately after that cast's terminal event forwards.**

# The strand

Spam a cast-time spell: press A starts the cast; impatient re-press B lands near cast-end. The legacy server batches B's rejection into the same tick that completes A, so the client receives, in one TCP flush — one client frame:

```
+0ms  SPELL_CAST_FAILED (dup B's rejection — CastID correctly B's, H7 pairing intact)
+0ms  SPELL_GO           (cast A completing)
```

The CastIDs are right — but the 1.14 client's kit-cancel sweep runs by **(unit, visualID), not CastID** (Mirasu R19). A same-frame FAILED+GO can tear the live cast's visual kit *ahead of* the GO's end-event-13 close. A torn kit whose close never lands is the orphaned loop sound: `SpellVisualKit` has no duration column, end-event-13 is the only terminator, and the handle sits at `obj+0x16c == 0` — no stop ever requested.

**Field evidence** (AliziaKaline's 08-14 JSONL, Stonetavern, 5.2h, DebugOutput on):
- 2/2 specimens with the exact signature (`cast.fail.spared_started` + same-ms `SPELL_GO`, Δ0–1ms), on spammed Lesser Heal 2053; the second sits **84s before a loop-shaped quiet logout**.
- Reporter named PW:Shield — **Shield and Lesser Heal share SpellVisualKit 270** (both on start-3 → end-event-13, same sound effect, verified in DB2 42597), so the perceived spell is structurally indistinguishable from the causal one. The report and the specimens agree.
- `RefireSpellGo` was **active** (600 refires that session, 167/167 on the named spell) and cannot reach this strand: no duplicate GO can close a sound handle the sweep already orphaned.
- Corpus sweep, delivered-to-client at |Δt|≤3ms: ours 13,055 across 70 sessions — all instant GCD-mash where the kit is never audibly live (the cofactor is a **live loop sound at sweep time**, i.e. cast-time spells); Mirasu's 15 instrumented sessions: **zero** delivered collisions, zero loops, consistent.

# The fix (SugarProxy's confirmed shape)

Re-decoded from the live Sugar binary this session (not the July doc): per-spell bucket; `SMSG_CAST_FAILED` → resolve first (consume queued press, else clear in-flight) → hold **only if the in-flight marker survived**; held list drained inside `SMSG_SPELL_GO` after the GO send *and* on the failure-terminal broadcast. Hold by data dependency, never a clock. Ours mirrors it with a strictly wider release set:

- **Hold**: dup resolved (`preferStarted:false`, unstarted, not MovementCancelled) while `HasStartedPendingCastForSpell` — packets built identically (same CastID pairing, same `SpellPrepare` rules, T1 pin untouched).
- **Release after the started cast's GO** (post-send — Sugar's replay position, empirically safe in its field record), **or its real CAST_FAILED** (FAILED+FAILED contradicts nothing), **or the stale sweep**: run at every cast-event handler top beside `RunWatchdogEviction`, releasing holds whose anchor left `PendingNormalCasts` by any eviction path — an unanswered dup press must still be delivered or the client's button strands lit.
- MovementCancelled acks exempt (the client blocks on them — the #161 bar-linger). `SuppressSpellCastErrors` DontReport acks ride the same hold (whether DontReport skips the kit sweep is unproven — Mirasu RE item, non-blocking).
- **Rider**: transient reasons (NotReady/SpellInProgress) no longer fire `HealCommBridge.OnLocalPlayerSpellStop` — a dup bounce is never an interrupt of the cast in progress (#397 axiom); the rez indicator no longer falsely clears mid-cast.

Worst-case UX cost: a dup press's red error text arrives one terminal later (~70–300ms in practice — the LL in-flight guard already drops early dups proxy-side, so survivors are late re-presses near cast-end).

# Scope

Local player casts only (pets follow if the field asks). Default ON, no config knob — deterministic packet-pairing correction, beta-channel. Diagnostics DebugOutput-gated: `cast.fail.dup_held` / `cast.fail.dup_flushed {release, held_ms}`.

# Testing

- 11 new tests (`DupFailureFrameHoldTests`): hold predicate incl. legacy-id (SoM) matching, FIFO hold/release, drained-once semantics, stale-sweep anchor lifecycle (started anchors hold; evicted/unstarted anchors release), and the Stonetavern specimen's wiring contract end-to-end at store level.
- Full suite **946/946** on the branch; build clean.
- Handler paths aren't unit-testable in this harness (no socket fixtures) — same limitation as every fix in this area.

# Field gate (ship-and-observe — no reporter interviews)

The verdict is Alizia's cohort on the next beta cut: spam-healing sessions that previously reproduced through an active refire. Success signature in their logs: `cast.fail.dup_held` → `dup_flushed release=go` pairs present, `cast.fail.spared_started` no longer co-timestamped with a same-ms `SPELL_GO` — and no loop report. This strand is refire-independent, so the result reads cleanly against Mirasu's refire-off control week.

# Interactions

- **#376/#397 transient guards**: unchanged and still required — they cover transient-with-no-dup (nothing to pair); this covers dup-present.
- **#488/RefireSpellGo**: independent; this strand is the one the refire demonstrably couldn't reach.
- **Mirasu's FIFO-wedge fix (incoming)**: different machinery (CastID pairing vs failure delivery); the hold defers only sends, never dequeues — textual adjacency in `HandleCastFailed` only.

https://claude.ai/code/session_016KDnDLRHcUyAuNgVLNB18w

### DIFF
--- a/HermesProxy.Tests/World/DupFailureFrameHoldTests.cs
+++ b/HermesProxy.Tests/World/DupFailureFrameHoldTests.cs
@@ -1,0 +1,218 @@
+using System;
+using HermesProxy;
+using HermesProxy.World.Server.Packets;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (dup-failure frame hold, the #394 collision strand): a dup press's CAST_FAILED
+// delivered while its same-spell STARTED cast is in flight can share a client frame with that
+// cast's SPELL_GO (the legacy server batches the dup rejection into the tick that completes
+// the cast — 2/2 specimens in the 2026-08-14 Stonetavern JSONL at Δ0-1ms). The client's
+// kit-cancel sweep runs by (unit, visualID), not CastID, so the same-frame FAILED can tear the
+// live kit ahead of the GO's end-event close and orphan the loop sound. The fix holds the
+// fully-built dup failure and releases it after the started cast's terminal event forwards
+// (SugarProxy's AddFailedPacket/GetFailedPacket shape — hold by data dependency, no clocks),
+// with a stale sweep tied to the pending-cast lifecycle so a silently-evicted anchor can't
+// strand the dup's button-release. These tests cover the store: the hold predicate
+// (HasStartedPendingCastForSpell), FIFO hold/release, and stale detection.
+public class DupFailureFrameHoldTests
+{
+    private static GameSessionData NewSession() => GameSessionData.CreateForTesting();
+
+    private static ClientCastRequest MakeCast(uint spellId, bool started = false, uint legacySpellId = 0)
+    {
+        return new ClientCastRequest
+        {
+            SpellId = spellId,
+            LegacySpellId = legacySpellId,
+            Timestamp = Environment.TickCount,
+            HasStarted = started,
+        };
+    }
+
+    private static GameSessionData.HeldDupFailure MakeHeld(
+        uint spellId,
+        uint reason = (uint)HermesProxy.World.Enums.SpellCastResultVanilla.SpellInProgress)
+    {
+        var held = new GameSessionData.HeldDupFailure
+        {
+            SpellId = spellId,
+            ReasonId = reason,
+            HeldAtMs = Environment.TickCount64,
+        };
+        held.Packets.Add(new CastFailed { SpellID = spellId, Reason = reason });
+        return held;
+    }
+
+    // ---- HasStartedPendingCastForSpell (the hold predicate) ----
+
+    [Fact]
+    public void HasStarted_EmptyQueue_ReturnsFalse()
+    {
+        Assert.False(NewSession().HasStartedPendingCastForSpell(2053));
+    }
+
+    [Fact]
+    public void HasStarted_OnlyUnstartedSameSpell_ReturnsFalse()
+    {
+        // A dup with no started twin (skip-START instant, or the cast already terminated):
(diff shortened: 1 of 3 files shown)


jameopotato:
Follow-up commit: moved the GO-handler's stale sweep from top-of-handler to after the GO forward. Reviewing #493's orphan-GO scenario surfaced the hole: a held dup whose STARTED anchor was destroy-evicted while the server still completes the cast would have its FAILED released at top and the GO forwarded right after — FAILED-then-GO in one flush with the kit live, the exact contradiction this PR prevents, re-created on a rarer path. Post-forward, every release lands after the GO in all geometries. The CAST_FAILED / SPELL_FAILURE handlers keep their top sweeps (failure-class emissions only; FAILED+FAILED contradicts nothing). Suite 946/946.

Note for whichever of #491/#493 merges second: both add an identical `HasStartedPendingCastForSpell` to `GlobalSessionData.cs` — one-hunk rebase, keep a single copy.

PR #493, opened 2026-08-21
Give every forwarded-START FIFO enqueue a consuming terminator, matched or not

Mongrul:
## Symptom

Mining (10248, 3.2s cast) bar kept running past the loot window on **every** node, session-wide, until a proxy restart. Wire was byte-clean throughout. Root-caused from `jimsproxy-20260820-191703.jsonl` (handoff: `MIRASU-MINING-CASTBAR-FIFO-WEDGE-20260820.md`, your §6 reply).

## Root cause

An orphan `SPELL_START` — the pending entry died in the `cast.destroy_eviction` race before its START arrived (~150ms) — still enqueues its forwarded CastID via `EnqueueForwardedStartCastId` (by design; the FIFO is the recovery store for exactly these no-pending STARTs). But its terminator took the **unmatched** paths, none of which touch the FIFO: every consumer (`TryPopForwardedStartCastId` in `HandleCastFailed` / `HandleSpellGo`, the watchdog remove-by-value, the parse-failure drain) sits behind a pending-entry match the orphan no longer has. The entry is immortal; every later same-spell cast pops the **previous** cast's CastID; GO arrives mispaired; the 1.14 client never closes the bar. `ClearForwardedStartCastIds` only runs on reconnect/transfer, so it never self-heals in-session — and it manufactures the exact stuck-cast / looping-visual class we're screening for in the A/B week.

## Fix — terminator symmetry on the three unmatched paths

*Every enqueue has exactly one consuming terminator, matched or not*, mirroring what the matched paths already do:

1. **Unmatched local CAST_FAILED** (`HandleCastFailed`, new trailing `else if`): pop the FIFO and forward the failure on the popped CastID — consumes the orphan entry and closes the bar the client actually opened. Safe by construction: transient dup-rejections early-return above (they never own FIFO entries), so only a real failure with zero pending entries reaches the pop. Marker: `cast.orphan_start_failure_resolved`.
2. **Unmatched local SPELL_FAILURE** (`HandleSpellFailure` else-branch): peek the FIFO for the forwarded CastID instead of re-minting the deterministic seed; the trailing CAST_FAILED still pops — same peek/pop pairing as the matched path. The `cast.unmatched_failure_forwarded` screening marker stays alive, now carrying `fifo_pinned`.
3. **GO-side stolen-dequeue guard** (`HandleSpellGo`, before the matched dequeue): a FIFO entry with an **unstarted** press pending and **no started** pending cast means the GO completes the orphan (one in-flight started cast per player; the press's own GO can't precede its START on the ordered stream) — stamp the GO from the FIFO and leave the press's entry for its own terminator, instead of the dequeue stealing it (which both mispairs this GO and re-arms the wedge). Uses a new `HasStartedPendingCastForSpell` twin of the existing non-started check. Marker: `cast.go.orphan_castid_recovered`.

## Correction to my handoff

The "sibling GO-side leak" as I described it was wrong — the unmatched-GO recovery pop already exists (the `cast.go.castid_recovered` else-branch), so the pure no-pending server-initiated START+GO pair was already balanced. The real residual GO edge is the stolen-dequeue case above (orphan completes while a fresh same-spell press sits queued — e.g. re-pressing the node after the eviction race when the cast completes instead of interrupting). Flagging it since you said you'd verify that pairing in review.

## Scope

Independent of #488, the refire question, and your dup-CAST_FAILED frame-hold (different machinery: CastID pairing vs failure delivery — your hold defers sends and never dequeues, so the only interaction is textual adjacency in `HandleCastFailed`).

Build 0 errors; suite **939/939** on the beta base (935 + 4 new in `ForwardedStartFifoSymmetryTests` pinning the orphan-GO precondition set and the pop-realignment sequence from the specimen).

https://claude.ai/code/session_01MQLjJBWwo8CAniARNUkBvm

### DIFF
--- a/HermesProxy.Tests/World/ForwardedStartFifoSymmetryTests.cs
+++ b/HermesProxy.Tests/World/ForwardedStartFifoSymmetryTests.cs
@@ -1,0 +1,101 @@
+using System;
+using HermesProxy;
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (fifo-terminator-symmetry): the forwarded-START CastID FIFO must obey
+// "every enqueue has exactly one consuming terminator, matched or not". An orphan
+// START (pending entry evicted before the START arrived) enqueues like any other,
+// but its terminator used to take the unmatched paths, which never consumed the
+// entry — leaving the FIFO off-by-one for every later same-spell cast for the rest
+// of the session (the mining cast-bar-overrun wedge, jimsproxy-20260820-191703).
+// These tests pin the session-level state the new unmatched-terminator paths key on.
+public class ForwardedStartFifoSymmetryTests
+{
+    private static GameSessionData NewSession() => GameSessionData.CreateForTesting();
+
+    private static WowGuid128 CastId(ulong low) => new WowGuid128(0x1234, low);
+
+    private static ClientCastRequest MakeCast(uint spellId, bool hasStarted = false, uint legacySpellId = 0)
+    {
+        return new ClientCastRequest
+        {
+            SpellId = spellId,
+            LegacySpellId = legacySpellId,
+            Timestamp = Environment.TickCount,
+            HasStarted = hasStarted,
+        };
+    }
+
+    [Fact]
+    public void HasStartedPendingCastForSpell_TrueOnlyForStartedSameSpell()
+    {
+        var session = NewSession();
+        session.EnqueuePendingNormalCast(MakeCast(10248, hasStarted: false));
+
+        Assert.False(session.HasStartedPendingCastForSpell(10248));
+
+        session.EnqueuePendingNormalCast(MakeCast(10248, hasStarted: true));
+
+        Assert.True(session.HasStartedPendingCastForSpell(10248));
+        Assert.False(session.HasStartedPendingCastForSpell(2053));
+    }
+
+    [Fact]
+    public void HasStartedPendingCastForSpell_MatchesLegacySpellId()
+    {
+        var session = NewSession();
+        session.EnqueuePendingNormalCast(MakeCast(25007, hasStarted: true, legacySpellId: 24997));
+
+        Assert.True(session.HasStartedPendingCastForSpell(24997));
+        Assert.True(session.HasStartedPendingCastForSpell(25007));
+    }
+
+    // The GO-side guard's exact precondition set: an orphan's forwarded-START entry in the
+    // FIFO, a fresh unstarted press pending, no started pending cast. The guard must resolve
+    // the GO from the FIFO and leave the press's pending entry untouched for its own
+    // terminator — dequeuing it here is the entry-steal that re-arms the wedge.
+    [Fact]
+    public void OrphanGoScenario_FifoResolvesWithoutStealingUnstartedPress()
+    {
+        var session = NewSession();
+        var orphanCastId = CastId(20733);   // orphan START forwarded after its pending entry died
+        session.EnqueueForwardedStartCastId(10248, orphanCastId);
+        var freshPress = MakeCast(10248, hasStarted: false);
+        session.EnqueuePendingNormalCast(freshPress);
+
+        Assert.True(session.HasNonStartedPendingCastForSpell(10248));
+        Assert.False(session.HasStartedPendingCastForSpell(10248));
+
+        Assert.True(session.TryPopForwardedStartCastId(10248, out var recovered));
+        Assert.Equal(orphanCastId, recovered);
+
+        // The press is still queued for its own START/GO/failure to consume.
+        Assert.True(session.HasNonStartedPendingCastForSpell(10248));
+        Assert.False(session.TryPopForwardedStartCastId(10248, out _));
+    }
+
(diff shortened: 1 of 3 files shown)


jameopotato:
Review with the MC-cap **(a)3 framing** in mind (the charm-victim AI-cast lane from the handoff bundle — your §8 pointer), verified statically against the `beta` base. Comment-only pass; findings below.

## Verified against beta

- **Transient safety of the new `HandleCastFailed` pop — confirmed, and stronger than the PR text states.** On beta a transient (NotReady/SpellInProgress) can never reach the new `else if`: with no unstarted dup pending, the `transient_stale_dropped` guard returns before the dequeue; with one pending, the dequeue (`preferStarted:false`) consumes it in the matched branch; the suppression branch and special slots also return/resolve upstream. So only a **real** failure with zero pending entries for that spell reaches the pop — the claim holds by the guard structure, not just by convention.
- **The pop sits at exactly the right place for the poisoning shape.** Beta's matched-GO T1 (`pendingCast.HasStarted` → `TryPopForwardedStartCastId`, ~`SpellHandler.cs:2136-2139`) *uses* the popped head — so a stranded orphan entry poisons the next real started cast's GO with the wrong CastID. Draining the orphan at its terminator (fix 1) protects that path at the source rather than papering over it downstream. This is the (a)3 defect closed for the terminator-delivered case.
- **GO-side stolen-dequeue guard — logic checks out.** The precondition set (unstarted press + no started pending + FIFO entry for that spell) can only be produced by an orphan of a cast-time spell, and a press's GO can't precede its START on the ordered stream, so the GO must be the orphan's. Instants never enqueue the FIFO, so they can't arm the guard. One negligible residual corner: a PoM-style instant-ified press of a spell that *also* has a live orphan in flight could get cross-stamped — requires the server to have accepted a second cast while the orphan is in flight, which the one-in-flight model mostly forbids, and it self-heals within one cast either way. Noted, not actionable.
- Fix 1 correctly skips `SpellPrepare` (no CMSG existed; the client tracks the bar by the START's CastID) — consistent with how the matched started-cast failure path stamps without a prepare.

## The (a)3 / charm framing — one residual to verify from your victim captures

The charm victim's pet-AI casts are exactly this FIFO's no-pending population (the START-side enqueue is unconditional for the local player), so:

- AI cast fails via **SMSG_CAST_FAILED** → fix 1 pops → symmetric. ✅
- AI cast completes via **SPELL_GO** → the pre-existing `cast.go.castid_recovered` pop → symmetric. ✅
- AI cast fails via **SMSG_SPELL_FAILURE with no trailing CAST_FAILED** → fix 2 peeks (right CastID on the wire, good) but nothing ever pops → the entry survives. Under a cap's failure storm that could strand up to `MaxForwardedStartCastIdsPerSpell` (8) same-spell entries, and each subsequent real started cast's matched-GO T1 pop then consumes one stale head and stamps the wrong id — the (a)3 client symptom, now bounded at ≤8 casts instead of session-permanent, but not zero.

Whether that third case exists on the wire is empirically answerable from your `jimsproxy-20260821-113918/-114838.jsonl` victim-lane captures: for the AI-cast failures during the caps, did SMSG_CAST_FAILED reach the victim session, or only SPELL_FAILURE? (cmangos `SendCastResult` returns early for triggered casts, so AI casts may plausibly be FAILURE-only.) If FAILURE-only shows up, the unmatched-FAILURE path likely wants the pop rather than the peek (there is no trailing terminator to defer to in that case) — happy to treat that as a follow-up rather than blocking this.

## Cross-PR coordination (needs a rebase whoever lands second)

**#491 and #493 both add `HasStartedPendingCastForSpell` to `GlobalSessionData.cs`** — the implementations are byte-identical (verified from both diffs), so semantics are safe, but the second PR to land must drop its copy or you get a duplicate member. Both PRs also touch `HandleCastFailed` *and* `HandleSpellGo` (#491's latest commit moves the GO-handler stale sweep), so expect textual conflicts beyond the one adjacency the Scope section mentions. Nothing here changes the logic of either PR — pure merge-order mechanics.

## Scope notes

- Tests pin the session-state contracts (helper semantics incl. legacy-id match, guard precondition set, pop realignment) — good coverage for what the fix keys on; a handler-level "transient can't reach the orphan pop" test would be nice-to-have but the beta guard structure makes that path near-unconstructible, so not requested.
- Static review only on my side — I did not independently run the suite (PR reports 939/939 on the beta base).

Net: the mechanism is right, the safety argument survives adversarial reading against beta, and it closes the (a)3 defect for every terminator that actually arrives. The FAILURE-only question and the #491 helper collision are the two things to resolve around it.

jameopotato:
Landed on beta as 7f5bb762 via cherry-pick (fork branch had a textual conflict with #486's HandleSpellFailure changes — resolved keep-both; your authorship preserved on the commit). Full suite on the merged tree: 950/950.

Review verdict was approve — mechanism verified twice (the earlier static pass on this thread plus a re-read against the post-#486 base). The FAILURE-only termination question from the review stands as the open follow-up: if your victim-lane captures show SPELL_FAILURE with no trailing CAST_FAILED for the AI-cast failures, the unmatched-FAILURE peek likely wants to become a pop. Rides v5.2.1-beta.1. Thanks Mirasu — textbook root-cause writeup, and the handoff self-correction on the GO-side was appreciated.

PR #510, opened 2026-09-01
Fishing: drop the previous bobber's stale channel zero-update on recast

Mongrul:
## Problem
Recasting fishing right after the cast bar empties desyncs fishing: the character stops the fishing animation while the new bobber floats out its full lifetime (fish even bite it), and every recast after that repeats the failure until the player waits out a whole bobber untouched.

## Cause (from a captured session, mechanism confirmed in cmangos-classic)
A fishing bobber outlives its channel by ~2s. On mangos-family cores the bobber's timeout (`GameObject::Update`, `GAMEOBJECT_TYPE_FISHINGNODE`) calls `Unit::FinishSpell(CURRENT_CHANNELED_SPELL)`, which sends a channel update with no time remaining for **whatever channel the player currently has** and finishes it, without checking that the bobber belongs to that spell. So when the player recasts inside the old bobber's remaining life, the old bobber's timeout ends the **new** channel ~100ms after it opened. The new bobber survives because a game object's lifetime is independent of the spell, and the `GO_READY` loot path has no channel requirement, which is why fish still bite it and the catch still works. Vanilla's `MSG_CHANNEL_UPDATE` carries no spell id, so the modern client faithfully ends the channel it just opened. Since the player naturally recasts the moment the char looks idle, every subsequent cast lands in the same race.

## Fix
The proxy keeps the client's channel open so the bobber can be waited out. `GameSessionData` tracks the local player's newest bobber (own `GAMEOBJECT_TYPE_FISHINGNODE` create block) and anchors the guard on packets, not clocks:

- **Arm** at `MSG_CHANNEL_START` of a fishing spell (`GameData.IsFishingChannelSpell`; ranks 1-4 + TBC Master) iff the previous bobber is still in the object cache. A session's first cast, a recast after the old bobber's destroy, and any non-fishing channel never arm.
- **Hold** the local player's zero-time channel update while armed, unless the client sent a channel-breaking action since the channel opened (`RecordLocalChannelBreakAction` in `HandleCastSpell`, `HandleCancelCast`, `HandleCancelChannelling`, `HandleGameObjUse`), in which case it forwards immediately.
- **Drop** the held update only when that bobber's `SMSG_DESTROY_OBJECT` (or `SMSG_FISH_NOT_HOOKED`) lands in the same read pass; if the anchor arrives first in the pass, the zero-update that follows is dropped directly.
- **Release** at legacy-socket drain (same rule as the #450 preempt attack stop): a zero-update with no anchor in its pass was genuine and goes through, closing the emote-guard channel window as before. An anchor with no zero-update in its pass disarms the guard.

The zero-update is only half of what the server sends when it ends the channel. The same tick also clears the player's `UNIT_CHANNEL_SPELL` and `UNIT_FIELD_CHANNEL_OBJECT`, and that values block is forwarded ahead of the held update. With those fields cleared the client drops the fishing pose, refuses the bobber click and refuses a recast until the bobber expires (this is what the 09-01 v1 test actually left behind: its one live drop was followed by a bite that never got a click). So after a drop the proxy also:

- **Re-asserts the fields** with a values update pointing the channel spell and channel object at the new bobber, so the pose, the bobber click and recasting all work.
- **Ends the client's channel itself** when that bobber's `SMSG_DESTROY_OBJECT` arrives (catch looted, fish escaped, timed out), with a synthesized zero-update, because the server no longer has a channel to end. A recast before that just supersedes it.

Everything else, including the old bobber's `SMSG_FISH_NOT_HOOKED` and despawn, still forwards. `spell.channel.stale_zero_update_dropped` and `spell.channel.orphan_end_synthesized` are behind `Settings.DebugOutput`; the release-at-drain edge logs `spell.channel.zero_update_released_at_drain` unconditionally.

Movement interrupts: the 1.14 client sends `CMSG_CANCEL_CHANNELLING` before the movement packet (70 of 70 episodes across 25 local sessions, one of them fishing), so a movement break always disarms the guard before the server's genuine zero-update arrives.

## Testing
- In-game on Kronos V, 2026-09-06, four sessions: spam recasts and recast-then-move never trip the guard (76 channels, zero guard events); the wedge shape (bite in the last second, delayed click, recast inside the bobber's remaining life) reproduced twice. With packet anchoring alone the drop landed and the wedge still showed (second wire above). With the field re-assert, the character kept fishing, the new bobber was clicked and looted 29 s later, and an immediate recast on top of it worked.
- 25 unit tests in `ChannelStaleZeroUpdateTests`: captured batch order (hold, drop on destroy, channel stays open, one-shot), fish-not-hooked anchor, anchor-before-update, anchor-then-drain disarm, release at drain, old bobber already destroyed, first cast, mid-channel recast, break-action disarm and reset, non-fishing channel, new START superseding a hold, orphaned client channel ended by the new bobber's destroy and cleared by a new START or drain release, rank coverage. Full suite 986/986.

https://claude.ai/code/session_01QW6ofT2m5aQQmD2ZnpagHj

### DIFF
--- a/HermesProxy.Tests/World/ChannelStaleZeroUpdateTests.cs
+++ b/HermesProxy.Tests/World/ChannelStaleZeroUpdateTests.cs
@@ -1,51 +1,141 @@
-using System;
 using HermesProxy;
 using HermesProxy.World;
+using HermesProxy.World.Objects;
+using HermesProxy.World.Server.Packets;
 using Xunit;
+using Disposition = HermesProxy.GameSessionData.LocalChannelZeroUpdateDisposition;
 
 namespace HermesProxy.Tests.World;
 
-// JimsProxy (fishing recast wedge 2026-09-01): state tests for the stale channel
-// zero-update guard in GameSessionData. Recasting fishing while the previous bobber
-// still exists server-side (its teardown lags the channel close by ~2s) makes
-// mangos-family servers tear the old bobber down on the tick AFTER the new channel
-// starts; the teardown's trailing MSG_CHANNEL_UPDATE(0) lands ~100ms behind the new
-// MSG_CHANNEL_START and would end the channel just opened on the modern client. The
-// guard arms only when a fishing channel starts moments after the previous fishing
-// channel closed, drops at most one zero-update, only within the post-start window,
-// and stands down on any client-side channel-breaking action.
+// JimsProxy (fishing recast wedge 2026-09-01): state tests for the held channel
+// zero-update guard in GameSessionData. A fishing bobber outlives its channel by ~2s,
+// and a mangos-family bobber timeout finishes whatever channel the player has at that
+// moment — so a recast inside that window has its NEW channel ended by the server
+// ~100ms after it opened while the new bobber lives on. The guard keeps the client's
+// channel open so the bobber can be waited out: it arms only when the previous bobber
+// is still in the object cache at the new CHANNEL_START, holds the zero-update, and
+// drops it only when that bobber's destroy (or SMSG_FISH_NOT_HOOKED) lands in the same
+// read pass; a socket drain releases anything still held as genuine.
 public class ChannelStaleZeroUpdateTests
 {
+    static ChannelStaleZeroUpdateTests()
+    {
+        // ServerPacket construction resolves opcodes through ModernVersion, whose
+        // static ctor needs a real build — same guard every packet-constructing
+        // test class uses so the class also runs green in isolation.
+        if (global::Framework.Settings.ClientBuild == HermesProxy.Enums.ClientVersionBuild.Zero)
+            global::Framework.Settings.ClientBuild = HermesProxy.Enums.ClientVersionBuild.V1_14_2_42597;
+    }
+
     private const uint FishingArtisan = 18248;
     private const uint MindFlay = 15407;
     private const uint ChannelDurationMs = 30000;
-    private const long T0 = 1_000_000;
+    private static readonly WowGuid128 OldBobber = new(0x1F00_0000_0000_0001, 0x0000_0000_0000_0001);
+    private static readonly WowGuid128 NewBobber = new(0x1F00_0000_0000_0002, 0x0000_0000_0000_0002);
+    private static readonly WowGuid128 SomeMob = new(0xF130_0000_0000_0003, 0x0000_0000_0000_0003);
 
     private static GameSessionData NewSession() => GameSessionData.CreateForTesting();
 
-    /// <summary>First channel runs its full course and closes; player recasts inside
-    /// the bobber-teardown lag. Returns the recast's start tick.</summary>
-    private static long RecastIntoTeardownWindow(GameSessionData session, uint spellId = FishingArtisan)
+    private static SpellChannelUpdate ZeroUpdate() => new() { TimeRemaining = 0 };
+
+    /// <summary>The server created a bobber for us (UPDATE_OBJECT create block).</summary>
+    private static void BobberCreated(GameSessionData session, WowGuid128 guid)
     {
-        session.OnLocalChannelStartAtTick(spellId, ChannelDurationMs, T0);
-        Assert.False(session.ConsumeLocalChannelZeroUpdateAtTick(T0 + ChannelDurationMs)); // natural end forwards
-        long recastTick = T0 + ChannelDurationMs + 1500; // old bobber still exists ~2s more
-        session.OnLocalChannelStartAtTick(spellId, ChannelDurationMs, recastTick);
-        return recastTick;
+        session.ObjectCacheLegacy[guid] = [];
+        session.LocalFishingBobberGuid = guid;
+    }
+
+    /// <summary>SMSG_DESTROY_OBJECT for a guid, as HandleDestroyObject drives the guard.</summary>
+    private static bool Destroyed(GameSessionData session, WowGuid128 guid)
+    {
+        session.ObjectCacheLegacy.Remove(guid);
+        return session.OnFishingBobberTeardownAnchor(guid);
+    }
+
+    /// <summary>The captured wedge's setup: first cast's bobber still alive when the
+    /// recast's CHANNEL_START arrives (the new bobber's create trails it).</summary>
+    private static GameSessionData RecastWithOldBobberAlive(uint spellId = FishingArtisan)
+    {
+        var session = NewSession();
+        session.OnLocalChannelStart(FishingArtisan, ChannelDurationMs);
+        BobberCreated(session, OldBobber);
+        Assert.Equal(Disposition.Forward, session.ClassifyLocalChannelZeroUpdate(ZeroUpdate())); // natural end
+        session.OnLocalChannelStart(spellId, ChannelDurationMs);
+        BobberCreated(session, NewBobber);
+        return session;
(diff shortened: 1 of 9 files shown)


jameopotato:
Review, 2026-09-05. Build clean, suite 977/977 on 3c5849ce. The player outcome is right and the capture is a good one, but the mechanism story is wrong in a way that changes the design, so this is a request for changes rather than a hold.

## The packet is not a stale tail of the old cast

On mangos-family cores the old bobber's timeout runs `caster->FinishSpell(CURRENT_CHANNELED_SPELL)` (vmangos `GameObject::Update`, `GAMEOBJECT_TYPE_FISHINGNODE`, "can't fish now"), and that helper sends `SendChannelUpdate(0)` for whatever channel the player currently has and then finishes it (cmangos-classic `Unit::FinishSpell`, Unit.cpp:4331). So when the old bobber times out ~100 ms after the recast, the server genuinely ends the new channel. The new bobber survives because the game object's lifetime is independent of the spell, and the `GO_READY` loot path has no channel requirement, which is why fish still bite it and the catch still works. The symptom is the server's true state rendered faithfully; the root cause is a Kronos-side bug where a bobber timeout finishes a channel it did not spawn.

What the guard actually does, then, is keep the client's channel visual alive after the server has ended the channel. That still fixes the player experience, because the player stops recasting into the teardown window and the bobber remains lootable, so I am fine with the approach. But the description and comments should say that, and the trigger should be tightened to match.

## Changes requested

1. **Anchor on packets, not clocks.** The 5 s arm window and the 1.5 s drop window are tuned to one server's tick and will drift. The capture already gives deterministic anchors: arm when the previous bobber game object is still alive in the object cache at the new `MSG_CHANNEL_START`, and drop only a zero-update that arrives in the same read pass as that bobber's `SMSG_DESTROY_OBJECT` or the `SMSG_FISH_NOT_HOOKED`. That is the shape you captured, and it removes both constants.
2. **Add the cast-spell break hook.** A `CMSG_CAST_SPELL` inside the window legitimately ends the fishing channel and is not covered by `RecordLocalChannelBreakAction` today, so its genuine zero-update would be eaten. One call at the top of `HandleCastSpell`.
3. **Gate the event.** `spell.channel.stale_zero_update_dropped` fires when the fix works, not on an unexpected edge, so it goes behind `Settings.DebugOutput` per the diagnostics rubric.
4. **Correct the mechanism text** in the PR body and the code comments: the server finishes the new channel; the proxy keeps the client's channel open so the bobber can be waited out.

Not needed: a config kill switch. The guard is narrow, one-shot and self-clearing. The CHANGES.md entry can be added as housekeeping at merge time.

After the rework it still gets a maintainer in-game pass before it rides a beta: rapid recasts, a recast followed by immediate movement, a recast followed by another spell, and a first cast interrupted by a mob.

Left open on purpose: whether the 1.14 client announces a movement interrupt of a channel with `CMSG_CANCEL_CHANNELLING`. None of our own field logs contain a channel, so if your capture shows one either way, please note it.

Mongrul:
Rework pushed (two commits on this branch), body rewritten. Build clean, suite 986/986.

## Your open question: does the 1.14 client announce a movement interrupt with `CMSG_CANCEL_CHANNELLING`?

Yes, and it sends it *before* the movement packet. The #510 capture cannot show it (the player never moved during a channel in that session, and it has zero cancel packets), but 25 other local sessions from 2026-08-13 through 2026-09-05 do contain channels:

| | Count |
|---|---|
| `CMSG_CANCEL_CHANNELLING` packets across the 25 sessions | 100 |
| Distinct interrupt episodes (cancels within 300 ms collapsed) | 73 |
| Episodes with a `CMSG_MOVE_START_*` / `CMSG_MOVE_JUMP` in the same or next millisecond | 70 |
| Of those, cancel logged before the movement packet | 70 of 70 |
| Episodes with no movement (Eagle Eye, manual cancel) | 3 |

Spells: First Aid bandages, Mend Pet, Eagle Eye, and Fishing once. The fishing specimen (channel 1398 ms old, player pressed back): `CMSG_CANCEL_CHANNELLING(18248)` at +0 ms, `CMSG_MOVE_START_BACKWARD` at +1 ms, then the server's `MSG_CHANNEL_UPDATE(0)` + bobber despawn + destroy at +163 ms. So the movement case was already covered by the break hook in `HandleCancelChannelling`. One behaviour worth knowing: while the client still believes the channel is up it re-sends the cancel with every further movement packet, ~40 ms apart, until the zero-update arrives (three to four cancels in 80 ms is typical).

## Mechanism: agreed, and verified in cmangos-classic

`Unit::FinishSpell` sends `SendChannelUpdate(0)` on whatever is in `CURRENT_CHANNELED_SPELL`, and the `GAMEOBJECT_TYPE_FISHINGNODE` timeout in `GameObject::Update` calls it without checking the bobber belongs to that spell. The body and the comments now say so.

## Changes 1 to 4, as requested

1. **Anchored on packets.** Armed iff the previous bobber (own `FISHINGNODE` create block, tracked in `LocalFishingBobberGuid`) is still in the object cache at the new `MSG_CHANNEL_START`. The zero-update is held, dropped only when that bobber's `SMSG_DESTROY_OBJECT` or `SMSG_FISH_NOT_HOOKED` lands in the same read pass, and released at legacy-socket drain otherwise, reusing the #450 drain rule (`FlushHeldChannelZeroUpdateAtDrain`, `LegacySocketDrained` shared with the preempt attack stop). Both constants are gone. One data point that shaped this: in all three captured teardowns the batch order is `UPDATE_OBJECT(bobber)` → `MSG_CHANNEL_UPDATE(0)` → `SMSG_FISH_NOT_HOOKED` → `SMSG_DESTROY_OBJECT`, so the zero-update precedes its anchors and has to be held rather than matched retroactively. Arming on cache presence matches all 19 casts in the capture.
2. **`CMSG_CAST_SPELL` break hook** at the top of `HandleCastSpell`.
3. **Drop event** behind `Settings.DebugOutput`; the release-at-drain edge logs unconditionally.
4. **Text corrected.**

## The in-game pass found a second wire

Recast-then-move and spam recasts never trip the guard (76 channels, zero guard events). The wedge shape (bite in the last second, delayed click, recast inside the bobber's remaining life) reproduced twice. With the packet anchoring alone the drop landed on `SMSG_FISH_NOT_HOOKED` exactly as designed, and the player symptom was unchanged: no fishing pose, bobber unclickable, and no recast possible until the bobber expired.

The reason is that the same server tick that sends the zero-update also clears the player's `UNIT_CHANNEL_SPELL` and `UNIT_FIELD_CHANNEL_OBJECT`, and that values block is forwarded ahead of the held update. This also means my 09-01 "confirmed fixed" on v1 was wrong: its one live drop is followed in the log by a bite on the new bobber that never got a click. So after a drop the proxy now also re-asserts the two fields onto the new bobber with a values update (`OnStaleChannelZeroUpdateDropped`), and, since the server never ends a channel it no longer has, ends the client's channel itself when that bobber's `SMSG_DESTROY_OBJECT` arrives (`EndOrphanedClientChannelIfBobber`, gated event `spell.channel.orphan_end_synthesized`). A recast before that supersedes it.

With that in place the repro passes: pose held, the new bobber was clicked and looted 29 s later, and an immediate recast on top of it worked.

Your maintainer pass list still stands; the first-cast-interrupted-by-a-mob case is the one I have not seen live.

PR #515, opened 2026-09-07
Lock talent spells at the respec confirm so a lagged press can't reach the server after the wipe

Mongrul:
## Problem

A player respecced during a lag spike, pressed Shadowform while the spellbook was still draining, and got the Kronos "Spell not in player book" kick + ban (since lifted).

The cast-block-unknown-spells guard (#185) is intact and does cover respecs, but it is reactive: it only drops a spell from `CurrentPlayerKnownSpells` when the server's `SMSG_REMOVED_SPELL` reaches the proxy. The server wipes every talent spell the instant it processes `MSG_TALENT_WIPE_CONFIRM`. Between the client's confirm leaving and the removal burst arriving there is a full round trip — seconds under a lag spike — in which a lingering action-bar press is forwarded for a spell the server no longer has. A native 1.12 client has the same race, so this is a hole the reactive design cannot reach rather than a parity divergence.

## Fix

Arm a speculative lock at `CMSG_CONFIRM_RESPEC_WIPE`, before the confirm is forwarded:

- **What gets locked:** every known spell that is, or descends by rank chain from, a talent spell of the player's class (`GameSessionData.CollectRespecLockSpells`). That catches the talent itself (Shadowform) and the trainer-bought higher ranks the server unlearns with it (Mortal Strike R3, whose R1 talent was superseded out of the known set long ago). Class filtering uses a new `GameData.TalentSpellClassMask` built from the ClassMask column already in `TalentSpellRanks.csv`; class 0 (not yet seen) locks every talent-rooted spell.
- **Enforcement:** `HandleCastSpell` rejects locked casts locally with the same `NotKnown` CastFailed the existing guard uses. `ForwardHeldGcdCast` re-runs both the known-set check and the lock at release, since a press parked for the GCD or a cast time can outlive a respec.
- **Release:** each real removal (`SMSG_UNLEARNED_SPELLS` / `SMSG_SEND_UNLEARN_SPELLS`) releases its spell. Right behind the confirm the proxy queues `MSG_QUERY_NEXT_MAIL_TIME`; the server handles both in order, so its reply fences the wipe — success or silent rejection — and stands down whatever the server kept. The reply is matched by ordinal so a reply to the client's own mail-time query from before the confirm can't stand the lock down early. Explicit rejections (empty-guid `MSG_TALENT_WIPE_CONFIRM` = "no talents", `SMSG_BUY_FAILED` = insufficient funds) and a fresh `SMSG_SEND_KNOWN_SPELLS` clear it too. A 120 s lazy timeout is the lost-fence backstop.
- The known-set mirror is never speculatively mutated; the lock is a veto layered on top of the guard.

## Wire confirmation on Kronos

Real respec on a paladin, log `jimsproxy-20260907-141942`: `spell.respec_lock.armed locked_count=28` → exactly 28 `SMSG_UNLEARNED_SPELLS` → `spell.respec_lock.drained` → `SMSG_SPELL_GO` 14867 (untalent visual) → `MSG_QUERY_NEXT_MAIL_TIME` reply, all within one 172 ms round trip. Locked set == server wipe set exactly (trainer ranks like Holy Shock R3 included), and Kronos answers the fence after the burst, i.e. in order.

## Edge cases covered by tests (`RespecCastLockTests`, 17 tests, suite 978/978)

- confirm → press before removal → blocked (the ban case)
- removal releases only its spell; last removal drains
- fence releases what the server kept; stale client-origin mail-time reply does not
- re-arm unions, restarts the timeout, moves the fence forward
- explicit clear forgets the fence, so a late reply can't stand down a new lock
- timeout backstop releases and reports
- other-class talents skipped; class unknown locks all
- never-armed state is inert on the uninitialized mock

## Residual gaps (unchanged from #185 v1)

Pet casts are still ungated. Talent-taught child spells that are not rank successors are not locked (none known in 1.12). The new lock state is synchronized; the pre-existing `CurrentPlayerKnownSpells` set is untouched.

## Log events

`spell.respec_lock.{armed,drained,cleared,expired}`, `spell.cast.blocked_respec_pending`, `spell.cast.blocked_at_held_release`.

https://claude.ai/code/session_01TsYcTs9JWTydFtP2tEbe5z

### DIFF
--- a/HermesProxy.Tests/World/RespecCastLockTests.cs
+++ b/HermesProxy.Tests/World/RespecCastLockTests.cs
@@ -1,0 +1,233 @@
+using System.Collections.Generic;
+using HermesProxy.World;
+using Xunit;
+
+namespace HermesProxy.Tests.World;
+
+// JimsProxy (respec cast lock): truth table for the speculative talent-spell lock armed at
+// CMSG_CONFIRM_RESPEC_WIPE. The cast-block-unknown-spells guard only learns a spell is gone when
+// SMSG_REMOVED_SPELL reaches the proxy; the server wiped it the moment it processed the confirm.
+// A press in that window is the Kronos "Spell not in player book" autoban (Shadowform, 2026-09-06),
+// so the orderings below are ban-critical:
+//
+//   confirm → press Shadowform            → LOCKED: rejected locally (THE ban case).
+//   confirm → REMOVED(Shadowform) → press → released from the lock; the known-set check now
+//                                           blocks it (removal took it out of the mirror).
+//   confirm → burst → fence reply         → whatever the server kept is released — safe to cast.
+//   confirm → silent rejection → fence    → same: nothing was removed, everything releases.
+//   confirm → "no talents" / BUY_FAILED   → explicit rejection releases before the fence.
+//   confirm → nothing, ever               → timeout backstop releases (fence lost).
+public class RespecCastLockTests
+{
+    private const uint Shadowform = 15473;        // priest talent, single rank
+    private const uint MindBlast = 8092;          // priest trainer spell
+    private const uint PermafrostR3 = 12571;      // mage passive talent, rank 3
+    private const uint MortalStrikeR1 = 12294;    // warrior talent
+    private const uint MortalStrikeR3 = 21553;    // trainer-bought rank of that talent
+    private const uint HolyShockR2 = 20929;       // paladin trainer rank of talent 20473
+    private const byte Warrior = 1, Paladin = 2, Priest = 5, Mage = 8;
+
+    static RespecCastLockTests()
+    {
+        GameData.LoadTalentSpellRanks();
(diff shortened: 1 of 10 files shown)


jameopotato:
Review, 2026-09-10. Verified against the code: the reactive guard cannot close the confirm-to-removal window, the fence is sound (both packets are processed in order on the map thread, and your Kronos log shows the reply landing after the 28 removals), the class mask really is column 1 of the talent CSV, and every held-press release path goes through the forwarder you re-check at release. Approved on your live respec log. One maintainer edit pushed to this branch (5640c2a): the armed, drained and cleared events are now DebugOutput-gated, following the fork's rule for fix-working breadcrumbs; blocked_respec_pending, blocked_at_held_release and expired stay unconditional. No behaviour change, suite 978/978. Goes into the v5.2.1-beta.3 batch.

PR #516, opened 2026-09-08
Ranged auto-repeat: give each tick its own CastID so the client keeps one cast object per shot

Mongrul:
The 1.14 client will not play a shot's animation on a cast object whose projectile is still in flight. `HandleSpellGo` re-stamped every Shoot / Auto Shot tick's GO with the press's prepared ServerGUID, so a whole auto-repeat series lived on one cast object, and any GO that arrived before the previous bolt landed lost its animation. Two field reproducers: a wand at max range (30 yd at 20 yd/s is the whole 1.5 s tick) and a hunter under Quick Shots (the tick shrinks under the ~1 s arrow flight). Every other shot animated. A native server sends a fresh CastID per tick and never shows this.

Only the press's first GO now takes the prepared id (`ClientCastRequest.FirstGoDelivered`); later ticks keep their per-tick CastID, which the synthesized per-tick SPELL_START already carried, so START and GO pair per shot. The synth START is a field copy with CastTime 0 and empty target lists (it aliased the GO, whose CastTime is the proxy-uptime GCD anchor). Each tick's damage log is stamped with that tick's CastID like a native server's. DebugOutput-gated per-tick diagnostics added (`spell.go.autorepeat_tick`, `spell.go.autorepeat_sent`, `spell.damage_log.autorepeat_castid_paired`, `spell.damage_log.autorepeat_hit`).

Verified on Kronos: wand at 30 yd and hunter under Quick Shots animate every shot at stock projectile speed. Side effect: toggling auto-repeat off at range now takes effect immediately instead of one tick late.

Ruled out on the way (all tested in game, none changed the skip): synth START cast time, damage-log CastID pairing alone, missile TravelTime stamping, closing the previous tick's cast, dropping the per-tick START, inverting the ranged haste modifier, native cast flags. Raising the client projectile speed via SpellMisc hotfix did cure it, which is what pinned the mechanism to the in-flight projectile before the CastID re-stamp was found.

https://claude.ai/code/session_016zHtnvVHRgu6fN3JLeQT8d


jameopotato:
Review, 2026-09-10. Verdict: request changes, then a re-test before it goes into the v5.2.1-beta.3 batch.

## What checks out

- The defect is on our wire independently of the animation claim. In an 86-minute wand session captured under the in-process harness (phase 1b), 217 synthesized wand STARTs carried per-packet ids while all 309 GOs were re-stamped to the prepared id, so 161 of 162 wand START objects never closed on the client. The code path matches: the synthesized START is built from the GO's cast data before the auto-repeat re-stamp, so it carries the minted id and the GO leaves with the prepared one. With this PR, later GOs keep the minted id and each pair closes per tick. The first GO after a press still pairs with the toggle's natural START, since the prepared id equals the deterministic id a local START gets.
- `ShallowCopy` is safe as written: `WorldSocket.SendPacket` serializes synchronously, so the synthesized START is on the wire before anything below touches the GO's cast data.
- Tracking observed casters' ticks in `AutoRepeatTickCastIds` is fine and consistent with the observed GO ids.
- The four new events are DebugOutput-gated. No interaction with #488 (both refire branches exclude auto-repeat spells) or with #517 (different paths, clean merge simulation).

## Two changes before merge

1. **Retarget STARTs.** Kronos also sends a natural SPELL_START when the auto-repeat target changes. That START keeps the deterministic id (the auto-repeat request is a special slot, nothing re-stamps it), it records `LastNaturalAutoShotSpellStartMs`, which suppresses the synthesized START for the next tick, and that tick's GO now keeps its minted id because `FirstGoDelivered` is already set. So every retarget produces a START with no GO and a GO with no START. Before this PR those two paired on the deterministic id. The client RE (round 18) says a GO on a miss mints a fresh object and finalizes it, while a START-promoted object is never visited by the sweeper, so each retarget would leave one object stranded for the session. The phase 1b session had 94 natural wand STARTs against 217 synthesized ones, and tab-targeting while wanding is ordinary play.

   Suggested shape: in the ranged branch of `HandleSpellStart`, when the slot's `FirstGoDelivered` is already set, mint the natural START a fresh sequence id and remember it on the slot. In the auto-repeat GO branch, if a remembered id exists within the same one-second window the synth uses, stamp the GO with it and clear it; otherwise keep the per-tick id and synthesize the START as now. A cheap check afterwards: run `tools/pkt_cast_ledger.py` from the evidence archive on a wand session with target switches; wand STARTs never closed should read zero.

2. **Damage-log pairing.** Each GO clears the queue and enqueues its own id, on the assumption that the hit lands before the next tick fires. Your own reproducer (30 yards at 20 yards per second, the whole 1.5 s tick in flight) is the case where the hit lands after the next GO, which then pairs the impact with the next shot's id. A two-deep FIFO dequeued in order, capped so misses without a damage log cannot let it grow, pairs both.

## Re-test after the changes (yours or the owner's)

1. Wand a target at 30 yards for a minute: every shot animates.
2. Hunter with Quick Shots up, Auto Shot at max range: every shot animates.
3. While auto-shooting or wanding, tab to a new target every few shots for a minute, then the ledger: zero wand or Auto Shot STARTs left unclosed, and no stuck bow-draw or wand pose afterwards.
4. Toggle auto-repeat off at max range: it stops immediately, not one tick late.
5. At max range, impact numbers and visuals on the target: none missing or doubled.
6. Next to another hunter who is auto-shooting: their shots unaffected.

Thanks for the in-game elimination work in the description; it is what pinned the mechanism.

jameopotato:
Withdrawn on 2026-09-11 after client RE round 20: the round-19 test was the wrong one. The wand's held sound node is gated on the impact state, which is stamped by the per-frame impact drain on the cast object resolved by the damage log's CastID, with an exact 16-byte probe and no fallback. Today every wand damage log is stamped with the natural-START object's id, so a synthesized tick's own object never receives its impact and its node stays parked. With this PR each tick's damage log carries that tick's id, the impact lands on that tick's object, and its node fires. So the per-tick objects cure the wand sound leak on the impact path, and the two-deep damage-log FIFO is what makes that pairing hold when a hit lands after the next tick's GO. The original addendum text is kept below for the record.

---

One addendum from the client RE (round 19, same day): the per-tick objects fix the animation and the START/GO pairing, but they do not cure the wand sound leak; the GO path never stamps the state the held sound node fires on, so that leak moves to the per-tick object rather than closing. Nothing to change here, the description does not claim the sound fix, just so the scope is clear.

Mongrul:
Both changes are in as 562cf34f, plus one finding from the field logs.

1. **Retarget STARTs**: implemented as suggested, with one widening. A natural START for the slot is remembered with the id it went out under (the press START keeps the prepared id, a START after the first GO is minted a fresh sequence id), the next GO carries that id and consumes its FIFO copy, and no START is synthesized while one is pending regardless of the one-second window. The reason for dropping the window: in two hunter sessions (189 presses, 1.14.2 client on Kronos) the client retargeted by CMSG_CANCEL_AUTO_REPEAT_SPELL then a fresh press every time, so no natural START arrived with the first GO already delivered; but 20 of 189 first GOs came more than a second after the press START (swing timer not ready), and each got a synthesized START under a fresh id ahead of a GO stamped with the prepared id. Same stranded-object shape, different trigger. Your retarget shape remains reachable through the OOR retry replay on emulators that send CAST_FAILED before CANCEL_AUTO_REPEAT, and the same code handles it.

   Side effect to check against the RE: the forwarded-START FIFO no longer holds stale auto-repeat ids, so a CANCEL_AUTO_REPEAT followed by CAST_FAILED after a closed pair now sends no CastFailed rather than one on the stale prepared id.

2. **Damage-log pairing**: two-deep FIFO, oldest first, in `RecordAutoRepeatTickCastId` / `TryPairAutoRepeatDamageLog`. Entries older than `AutoRepeatTickCastIdMaxAgeMs` (3 s) are skipped so a tick whose hit never arrives (target dies mid-flight) cannot shift every later pairing by one shot. `AutoRepeatTickCastIdPairingTests` pins the three rules.

**Re-test (Kronos, this build):**

- Wand at range with several target changes, some while a bolt was in flight: 4 presses, 35 ticks. Every press START closed on its first GO (one was a miss), the 31 later ticks each got one synthesized START under their own id, 31/31 damage logs paired, and 25 of those hits landed after the next tick's GO, so the two-deep queue is exercised on Kronos. Raw GO to damage-log gap was 1.0 to 1.7 s at that range: Kronos delays the wand hit by flight time.
- Auto Shot under Quick Shots: 1 press, 45 ticks in one series with three 30% haste windows (16 ticks under 2.5 s), 45 distinct cast ids, 44 synthesized STARTs, 45/45 damage logs paired. Every shot animated.
- Not covered yet: a nearby auto-shooting hunter. The observed-caster path is untouched by this commit.

Noted on the round-19 addendum: the description does not claim the sound fix.

Mongrul:
A finding that cuts against the pairing model in this PR, found while chasing a user report that `/cast !Auto Shot` cancels auto-shot through the proxy. Verified in game today on the 562cf34f build; the fix is not written yet because it changes what we agreed the press START should do.

## What the client does with `!`

Two different states answer "is Auto Shot on" on 1.14.2 (42597), read from unprotected code with the harness:

- **The toggle test** in the cast entry (`0x10c9430`, continuation at `0x10c94a9`): `spellId == [0x2f35544]`, the client's *current auto-repeat spell* global. That global is set by the client's own cast-send path (`0x10c8ce4` in `Spell_C` `0x10c7140`) the moment the button is pressed, so it reads 75 for the whole series through the proxy as well. Match means cancel: `0x10c5340` walks the current-cast list, ends every object whose spell id equals the global, sends CMSG_CANCEL_AUTO_REPEAT_SPELL, clears the global.
- **The `!` test** reads the current-cast list, the same walk `IsCurrentSpell` exposes (`0x10d9440`: list membership by `node+0x40`, then Attack / tradeskill / mount / shapeshift / summoned pet, no aura branch). `IsAutoRepeatSpell` and `IsAutoRepeatAction` read the global instead (`0x10ca220` / `0x10ca200`).

In game through the proxy, while auto-shooting, after the first arrow: `IsCurrentSpell("Auto Shot")` is false and `IsAutoRepeatSpell("Auto Shot")` is true. That combination is exactly the failure: `!` sees no live cast object, falls through to the toggle branch, the global matches, and the client cancels. So `!Auto Shot` turns auto-shot off after the first shot.

## Why the object is gone

The press START goes out on the prepared id and, per this PR, the first tick's GO is stamped with the same id (`HandleSpellGo` auto-repeat branch, `PendingNaturalStartCastId` / `FirstGoDelivered`), which completes and unlinks the press object. On a native server that GO never happens: on TrinityCore-lineage cores the press Spell object stays in the auto-repeat slot and never runs `cast()`; each tick is a fresh triggered Spell from `Unit::_UpdateAutoRepeatSpell` with its own cast id. The press object lives until the series stops. Worth checking against your Classic captures: I expect the press START's cast id never to appear in a SPELL_GO.

How the press object closes natively is the part that matters for the ledger: the SMSG_CANCEL_AUTO_REPEAT handler (`0x12be050`, Unit_C.cpp, local player only) runs the same `0x10c5340` cancel walk as the toggle press, which ends every cast object carrying the active spell id. So the press START is closed client-side by CANCEL_AUTO_REPEAT, not by a wire GO or FAILED. A wire-only ledger counts that as "START never closed", which is what the 161/162 wand figure included.

For completeness, the other server messages that clear the client's auto-repeat global (the CAST_FAILED handler `0x10dd489`, reason jump tables decoded): a CAST_FAILED for the active auto-repeat spell whose target is the auto-repeat target with reason BadImplicitTargets, BadTargets, CasterDead, EquippedItem*, NeedAmmo, OutOfRange, TooClose or UnitNotInfront. DontReport, NotReady, SpellInProgress and LineOfSight leave it alone.

## Second proxy-only killer

A re-press while `CurrentClientAutoRepeatCast` is set is rejected locally in `HandleCastSpell` ("already have one of this type in progress", `SendCastRequestFailed`), which closes the fresh press object too. The client only re-presses when its global has been cleared, so on Kronos this is rare (CANCEL precedes CAST_FAILED and the slot is already null), but on an emulator that sends a table-A CAST_FAILED without a CANCEL, the client re-arms itself when back in range and the proxy kills the new object.

## Fix shape I would propose

1. Never stamp a GO with the press START's id. Every tick, the first included, gets the minted per-tick id with its synthesized START, so each tick still closes as a pair. The press START keeps the prepared id and simply stays open for the series.
2. Drop the press START's forwarded-START FIFO copy when the slot ends (CANCEL_AUTO_REPEAT, CMSG_CANCEL_AUTO_REPEAT_SPELL, non-retryable CAST_FAILED), so a trailing CAST_FAILED can't pair to an object the client has already ended.
3. Treat a re-press while the slot is live as a new series on the client side (PREPARE plus START on the new press id) rather than a rejection; whether to forward it is a separate call, vanilla cores replace the current auto-repeat spell on re-cast.
4. Ledger rule: a press START (prepared id) counts as closed by the CANCEL_AUTO_REPEAT or SPELL_FAILURE that ends the series, not by a GO.

The retarget change from your review is unaffected in substance: a natural START after a retarget is the press START of the new series on Kronos (the client cancels and re-presses), and it should stay open the same way.

One related FYI from the same report: `/cast !Aspect of the Hawk` recasting the aspect is not ours. The active test has no aura branch at all, so the client cannot know an aspect is up; native behaves the same.

jameopotato:
Agreed, and confirmed on the current beta: every tick GO is stamped with the prepared id there too, and the special-slot PREPARE re-keys the press object at forward time, so the first GO has been closing the press object on beta and stable for as long as the re-stamp has existed. Your `!` test explains a behaviour players already have. Your fix shape is the right one: every tick including the first on its own minted id with its synthesized START, the press START left open for the series, its FIFO copy dropped when the slot ends, and a re-press treated as a new series client-side. The retarget pairing and the two-deep damage-log queue from the last round stay as they are. The approval from the re-review is withdrawn until that lands, and I'll re-review the revision in full. The ledger rule you propose (a press START closes by the CANCEL or SPELL_FAILURE that ends the series) will be applied to the earlier wand counts as well, since they included press STARTs.

Mongrul:
The revision is in as 0323456a + 7f5532e8, rebased onto beta 8dffc174. No GO carries the prepared id any more: every tick including the first keeps its parse-minted id and gets a synthesized START on the same id, and the one-second window is gone with it. The press START keeps the prepared id and is never paired; its FIFO copy leaves with the slot in a new `EndAutoRepeatSlot` (client cancel, server cancel, non-retryable failure). The retarget START pairing and the two-deep damage-log queue are unchanged.

Two findings from the in-game rounds. First, `/cast !Auto Shot` re-sends the press on every key repeat for as long as the series runs: the client's `!` test walks its predicted-cast ring for an object with the spell id, and through the proxy that object is not there. Forwarding a re-send made Kronos interrupt the running series (`SMSG_CANCEL_AUTO_REPEAT` + `SPELL_FAILURE` + a fresh START per re-send, and the cancels emptied the slot so later GOs fell back to the press id), so re-sends stay local. Second, answering each one with a client-id `CAST_FAILED` fires `UNIT_SPELLCAST_FAILED_QUIET`, which WeaponSwingTimer reads as the native 0.5 s re-arm delay and bumps its bar. A native server ignores the duplicate outright (TrinityCore's "client is resending autoshot cast opcode"), so the second commit does the same. The client's own cancel walk ends every ring object of an auto-repeat spell when the series stops, so nothing is left behind; that walk, the `!` test and the active-spell global were read in the 42597 image this session (`0x10d9440` walks the state-2 ring head at `0x2f405b0`, link at object+8, compares object+0x48; `0x10c5770` ends every ring-2 object whose spell carries the auto-repeat attribute; `0x10dcb00` sets the active global from the send path).

Series end on Kronos reads the same on the wire: the cancel ends the slot first, and the trailing `SPELL_FAILURE`/`CAST_FAILED` fall to the deterministic seed, which is the press id by construction. Four tests pin the slot-end rule; suite 1083/1083. In-game on Kronos: a full spam series holds, nine ticks paired on their own ids, 52 duplicates dropped, swing timer clean.

https://claude.ai/code/session_0118QnB5nvfhvqNH2AknbhF7

jameopotato:
# PR #516 re-review of 7f5532e8 (2026-09-13): request changes

The code matches the model agreed on 09-13 point for point: no GO carries the prepared id, every tick including the first is its own synthesized START plus GO pair, the press START keeps the prepared id and is never paired, its FIFO copy leaves with the slot through `EndAutoRepeatSlot` on all three series ends, the retarget pairing and the two-deep damage-log queue are unchanged, and a duplicate press for a running series is dropped before the old reject. The slot-end FIFO release also closes a real gap on beta, where the press copy lingered for the next series' first failure to pop. Local suite on the head: 1083/1083 (fork PR, no CI). Merge simulation clean on beta and on beta plus #525. Two findings from the client side change the outcome.

## 1. The silent drop leaves an unbounded per-series accumulation on the client

Each ignored re-send leaves a client-minted state-2 object under its client id. The proxy never answers it, and nothing on the client recomputes a ring-2 object on its own (round 19). They are freed only at series end, when the client's cancel walk (0x10c5770) ends every ring-2 object of an auto-repeat spell; that is why "nothing is left behind" holds between series but not inside one. The 09-14 spam series produced 52 in one series. The ring holding them is capped at 100 entries (round 21); at the cap new predicted casts are diverted or not minted, and the wand cast-start path skips the mint entirely. A series twice as long, or one held under key repeat by a player who parks a finger on the key, crosses the cap mid-series, and from then on every instant pressed during the series gets no client-side object until the series ends. That is a new failure the silent drop introduces in exchange for keeping the swing timer quiet, so the revision cannot merge with the accumulation unbounded.

Requested change, usable now: keep ignoring re-sends, but count them on the slot, and once the count in a series passes a threshold, answer the next one on its client id with the DontReport failure and reset the count. Twenty is the suggested threshold; the trade is one quiet failure per threshold-many re-sends for a player holding the key (the swing-timer bump, once per batch instead of once per repeat), against ring headroom. Per rounds 18 and 19 that failure resolves the object, recomputes it to state 0, and the next frame frees it, so the ring stays far below the cap while a tap never sees the bump. The #517 helper sends a client-id CAST_FAILED without a PREPARE. A test on the session state should pin the counter and threshold rule the way `AutoRepeatSlotEndTests` pins the slot end.

## 2. The press START stays open, yet the `!` test still does not find the press object

The 09-14 comment gives the walk: 0x10d9440 iterates the state-2 ring (head 0x142f405b0, link at object+8) comparing the spell id at object+0x48. Round 19 has the client minting its own predicted press into exactly that ring. So natively and through the proxy the press object starts where the walk looks, and with 0323456a no GO carries its id any more, yet the miss persists. Something after the mint takes it out of ring 2, and nothing in rounds 15 to 21 names it. The 09-12 explanation (closed by the first GO) is superseded. So the first commit delivers the native object shape and the pairing, and the duplicate drop is what stops the cancel, by tolerating re-sends rather than preventing them; a native server sees zero re-sends because the walk finds the object.

Candidates, none proven: the forwarded press START on the prepared id, if its flags or cast-time field drive the terminal recompute (round 19: that recompute never emits state 2, so one pass evicts the object from the ring); the per-tick GOs' inline sibling walk (0x10f8540, rounds 16 and 17), which touches same-caster objects and relinks rings; or the PREPARE rebind itself, if the ring links are among what it rewrites. The discriminator needs no RE round: a harness sweep across one series printing the press object's state and ring at press, after the PREPARE, after the START, and after the first tick GO. Whichever step moves it out of ring 2 is the answer; if it is the START, a field diff of our forwarded press START against a native capture follows. If the walk finds the object, the client stops re-sending as it does natively and item 1's accumulation never starts, so this is the real fix and its answer may change item 1's shape. Not a merge blocker on its own once item 1 bounds the accumulation.

## Also owed

A CHANGES.md entry for the PR, with the next push or as housekeeping at merge.

## Re-test on the next head

1. Wand a target at 30 yards for a minute: every shot animates, first shot included.
2. Hunter under Quick Shots at max range: every shot animates.
3. Tab to a new target every few shots for a minute, then the ledger: no wand or Auto Shot STARTs left unclosed except the press START, and no stuck pose afterwards.
4. Toggle auto-repeat off at max range: immediate.
5. Impact numbers at max range: none missing or doubled.
6. Next to another auto-shooting hunter: their shots unaffected.
7. Kite out of range mid-series and come back: the series resumes on its own, `/cast !Auto Shot` stays quiet, and the button does not stick. Note for this one: `HandleSpellFailure` has no special-slot branch, so a mid-series SPELL_FAILURE for the spell is forwarded on the FIFO head, which during a series is the press id. Same as before the PR, so not a regression, but under the new model it would close the press object; the scenario tells us whether Kronos sends that packet on a bounce.
8. Hold `/cast !Auto Shot` on key repeat through a whole fight, then cast Arcane Shot and Multi-Shot: both fire normally.
9. Priest: wand, then cast Smite, then wand again: the wand restarts cleanly.
10. A deliberately long `/cast !Auto Shot` key-repeat series, two to three minutes, with ring-2 occupancy counted before, during and after, and an instant pressed near the end: it must get its client-side object, and the count must stay well under 100 throughout.

Mongrul:
Item 1 is in as 9d68696d, with the CHANGES.md entry as 4a278407, fast-forward on 7f5532e8. Suite 1087/1087. The shape differs from the one requested, because that one was built first and failed in the field.

## The threshold shape does not bound the ring

Built as asked (count on the slot, the duplicate after twenty drops answered on its client id with DontReport, count reset) and run on Kronos (`jimsproxy-20260914-055933.jsonl`): `/cast !Auto Shot` held on key repeat sent the press 104 times at about 5.5 a second, four were answered (the 21st, 42nd, 63rd, 84th), and at 60 s the client hit the cap: it stopped sending the press altogether (the auto-repeat cast-start path skips the mint at the cap) and every instant pressed after that arrived on a diverted client id (`Low = 0x4000001` upward, a different high qword), the failure from your item 1 exactly. The server still ran some of them, but the client had no object for them, and the player saw dead buttons.

The arithmetic: a client-id CAST_FAILED resolves the one object it names (round 19), so a threshold of N leaves N of every N+1 re-sends in the ring, and a held key reaches 100 entries after about 100 re-sends, around 20 s, for any N. Every duplicate has to be answered; the only freedom is when.

## What landed instead

The swing-timer objection is narrower than the commit message said. WeaponSwingTimer's `OnUnitSpellCastFailedQuiet` bumps the bar only when the failure lands inside the aim window at the end of the swing (`(now - last_shot_time) > (range_speed - auto_cast_time)`, the last 0.5 s). So the duplicates are held on the slot (`HoldAutoRepeatDuplicatePress`) and answered together on their client ids, DontReport, no PREPARE, right after each tick GO (`TakeHeldAutoRepeatDuplicatePresses` in `HandleSpellGo`), when the swing has just reset. The hold is capped at forty; past that a duplicate answers on arrival, which only a series that never fires can reach (a key held through a line-of-sight bounce, where there is no swing to protect). The slot end drops the hold, since the client's cancel walk ends those objects itself. `AutoRepeatDuplicatePressTests` pins the hold, the cap, the slot-end drop and the no-series case.

Field run (`jimsproxy-20260914-063428.jsonl`, Kronos, same macro held for a minute): 159 presses, 158 held and answered across 22 tick GOs, at most 15 per tick, none on arrival, no diverted client ids, Aimed Shot, Multi-Shot and Arcane Shot forwarded on normal ids throughout, swing timer clean.

Item 2 (what takes the press object out of ring 2 after the mint) is untouched; your harness sweep is the right discriminator and I have not run it. If the walk ends up finding the press object, the hold simply never fills.

Re-test 8 and 10 from your list are covered by the run above; 1 to 7 and 9 are on the earlier heads and unchanged by this commit apart from the duplicate path.

https://claude.ai/code/session_01KdHfVN6LWnww4pc7NBSMZW

Mongrul:
Pushed 51a752c0 on top of 4a278407 (fast-forward): a wand-only exception to the "press START stays open" rule, plus three findings from today's wand sessions that bear on your item 2.

**The regression.** With the press START left open for the series, the wand's aim sound ran from the press until the series ended (jimsproxy-20260914-191217.jsonl: press START on the prepared id, never closed, every tick its own START + GO pair, series end = client CANCEL then the server's CANCEL_AUTO_REPEAT + SPELL_FAILURE). A bow's open press object only holds the draw pose, which is why the hunter re-test was clean.

**Finding 1: the wand aim sound lives on the client's own press object, not on the server START.** I first tried withholding the wand press START altogether (PREPARE only, first tick's own START + GO for the aim). The hum was unchanged (jimsproxy-20260914-193932.jsonl, `spell.start.autorepeat_press_withheld` on every press). So the sound starts at the press, on the client-minted object, and the only thing that ends it is closing that object. The shipped fix therefore does what the old build did for wands: the first tick GO carries the prepared id and closes the press object, with no separate synthesized START for that first tick (`HandleSpellGo`, `wandPressPair`, keyed on `GameData.WandShootSpellId`); later wand ticks pair as before. This is also Blizzard's own shape, where every aim START, the press included, is closed by its GO half a second later. User-verified: sound per shot again, first shot animates, 30 yd clean.

**Finding 2: an open, STARTED cast object of the auto-repeat spell is what shields the client from the `!` toggle-off path.** Across today's builds the client's response to a `/cast !Shoot` re-press mid-series was: press object open (STARTed, no GO) → a CMSG_CAST_SPELL re-send (held and answered after the tick GO, the hunter case); press object closed by its GO, or PREPAREd but never STARTed → CMSG_CANCEL_CAST with spell 5019 and an EMPTY cast guid, then CMSG_CANCEL_AUTO_REPEAT_SPELL, and the server ends the series (jimsproxy-20260914-194619.jsonl, four series). So the `!` walk over ring 2 fails in every shape (your item 2), and what decides between re-send and toggle-off is whether a started object of the spell is still open.

**Finding 3: a second, server-minted open Shoot START shields silently but the client then rejects re-presses locally.** As an experiment (never on the branch) I opened one extra Shoot START on a fresh id right after the first tick GO and left it for the series-end SPELL_FAILURE to close. It did not hum and `!Shoot` spam no longer stopped the wand, but the client answered each re-press itself with "Ability is not ready yet" and only one CMSG_CAST_SPELL reached the proxy in the whole session (jimsproxy-20260914-195341.jsonl), and the spam clipped the shot sound. The user preferred the shipped shape without it.

**Net:** wand sound is back to per-shot; wand `/cast !Shoot` toggles off after the first shot exactly as it did on beta, and every proxy-side lever for that is now tested. The real fix is your item 2 (why the press object is not in ring 2 for the `!` walk); if the sweep finds the eviction step, the wand exception here can go and the press START can stay open for wands too. Hunter `!Auto Shot` and the hold-and-flush are unchanged. CHANGES.md entry amended with the wand exception. 1087/1087.

PR #522, opened 2026-09-11
Forward SMSG_FEIGN_DEATH_RESISTED so a resisted Feign Death is reported instead of silently greying the button

Mongrul:
**Issue:** a creature in combat with the player can resist Feign Death. The server then keeps the aura, so the button greys for the cooldown, but never sets the dead flags, and the only word the client gets is `SMSG_FEIGN_DEATH_RESISTED`. The proxy had no handler for that opcode (it logged as `packet.untranslated`), so on a resist the player saw a greyed button and nothing else: no pose, no message, no way to tell a resist from a failed cast.

**Evidence:** Kronos, hunter at a target dummy, `jimsproxy-20260911-102024.jsonl`, with a temporary diagnostic on the local player's `UNIT_DYNAMIC_FLAGS` updates. Three presses. The first two feigned normally: the server set the dead dynamic flag and `UpdateHandler` forwarded it with `UnitFlags2.FeignDeath`. The third, after idling in combat, arrived with `SMSG_FEIGN_DEATH_RESISTED` and no flag change, the unit flags carrying the in-combat bit. The aura applied in all three cases, which is what greyed the button.

**Change:** `Client/PacketHandlers/SpellHandler.cs` gains `HandleFeignDeathResisted`, a passthrough that sends the new `FeignDeathResisted` server packet (`Server/Packets/SpellPackets.cs`); both the legacy and the modern packet are empty. One `spell.feign_death_resisted` event marks it in the log.

**Verification:** field-tested on Kronos: the resist now shows the client's "Feign Death resisted" text, and normal feigns are unchanged. 1075/1075 tests on this branch.

https://claude.ai/code/session_01UfMt474vfrxheQw8Yhm3Y5

### DIFF
--- a/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
+++ b/HermesProxy/World/Client/PacketHandlers/SpellHandler.cs
@@ -3336,6 +3336,16 @@ public partial class WorldClient
         return dbdata;
     }
 
+    // JimsProxy: a creature in combat with the player can resist Feign Death. The server then keeps the aura (the
+    // button greys for the cooldown) but never sets the dead flags, and this packet is the only word the client gets.
+    // Unforwarded, the player sees a greyed button and nothing else (Kronos, 2026-09-11, idle in combat at a dummy).
+    [PacketHandler(Opcode.SMSG_FEIGN_DEATH_RESISTED)]
+    void HandleFeignDeathResisted(WorldPacket packet)
+    {
+        Log.Event("spell.feign_death_resisted", new { });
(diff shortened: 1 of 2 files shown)


jameopotato:
Review, 2026-09-11. Verified: the opcode is in every legacy table (0x2B4) and in the modern table the proxy uses for 42597 (0x273C), the packet is empty on both sides and the class matches the existing empty packets, it touches no cast state, and the merge onto beta is clean. Approved on your Kronos field test; merged for v5.2.1-beta.3, with the fork changelog entry added at review (#524).

